PR #51, opened 2026-05-01
Fix stop-button cancellation on Windows (use taskkill /F /T)

bdraco:
## Summary

The POSIX stop-button path uses `os.getpgid` / `os.killpg` / `signal.SIGKILL` to take down the whole `esphome → platformio → gcc` subtree. None of those exist on Windows, so hitting Stop during a compile on Windows would crash the firmware controller with `AttributeError` instead of cancelling the build.

`_terminate_current_process` now branches on `sys.platform`:

- **POSIX**: unchanged — SIGTERM the process group, wait the grace window, escalate to SIGKILL on the group.
- **Windows**: shell out to `taskkill /F /T /PID <pid>`. `/T` walks the kernel's parent-PID tree, `/F` is the forceful equivalent of SIGKILL. There's no useful polite stage on Windows — non-GUI processes ignore `WM_CLOSE` / `CTRL_BREAK_EVENT` — so we go straight to the forceful kill. If `taskkill` is missing or hangs, fall back to `proc.kill()` so the runner loop can still finalise the job.

Follow-up to #48, which enabled the Windows CI matrix and skipped the POSIX-only test module on Windows. This PR fixes the actual production bug that the test was catching.

## Test plan

- [x] Existing POSIX integration test (`test_firmware_stop.py`) still passes on macOS / Linux.
- [x] New `tests/test_firmware_stop_windows.py` covers the Windows path: an integration test (spawn `python -c "time.sleep(60)"`, call `_terminate_current_process()`, assert it exits within 5s) and a unit test for the missing-taskkill fallback.
- [ ] Windows CI matrix entry passes the new tests (the Windows runner has `taskkill` on PATH by default).

### DIFF
--- a/esphome_device_builder/controllers/firmware.py
+++ b/esphome_device_builder/controllers/firmware.py
@@ -173,6 +173,9 @@ def _signal_process_group(pid: int, sig: int) -> bool:
     same group — ``killpg(getpgid(spawned_pid), sig)`` therefore
     targets the build subtree without touching us.
 
+    POSIX-only — ``os.getpgid`` / ``os.killpg`` don't exist on Windows.
+    The Windows path goes through ``_terminate_subtree_windows`` instead.
+
     Falls back gracefully:
     * ``ProcessLookupError`` — the process already exited; nothing to do.
     * ``PermissionError`` — we lost the right to signal it; treat as a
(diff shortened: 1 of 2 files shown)


bdraco:
Addressed all three Copilot review comments in 080f74e:

1. **Direct `asyncio.create_subprocess_exec`** — switched the `taskkill` spawn to the package's `helpers.subprocess.create_subprocess_exec` wrapper, so `close_fds=False` is applied. The guard test in `test_subprocess_helper.py` now passes against the new code.
2. **Return value didn't reflect taskkill exit code** — now check `killer.returncode` after `wait()` and return `False` on non-zero, so `_terminate_current_process` falls back to `proc.kill()` on access-denied / invalid-pid / partial-kill cases instead of silently leaving the subtree alive.
3. **Test monkeypatched the wrong symbol** — now patches `firmware_module.create_subprocess_exec` (the symbol the production code actually calls), and added a sibling test covering the new non-zero-exit fallback.

PR #57, opened 2026-05-02
Add components/get_integration_docs WS command

bdraco:
## Summary

New WS command that returns `{integration_name: docs_url}` for every loaded-integration identifier we can resolve to a real esphome.io page. Frontend (separate PR) uses this to turn the device drawer's loaded-integration tags into clickable links.

The data comes from the existing component catalog (`components.json`), which `script/sync_components.py` already builds from the schema bundle's "See also" links plus the docs MDX repo. This command just re-shapes the catalog into a name → URL projection so the frontend doesn't have to ship the full catalog when it only needs the lookup.

### Three lookup layers, priority-ordered

The most-specific page wins:

1. **Top-level component** — catalog id without `.` (`api` → `/components/api`, `wifi` → `/components/wifi`).
2. **Category landing** — `sensor` → `/components/sensor`, synthesised from any `<cat>.<stem>` doc URL only when the URL is genuinely under `/components/<cat>/`. Skips multi-platform components like `switch.at581x` (docs at `/components/at581x`) that would otherwise produce a wrong category URL.
3. **Stem alias** — picks up specific drivers like `ltr390` whose catalog id is `sensor.ltr390`.

Names with no catalog hit are simply omitted; the frontend renders those as plain text. The catalog's `docs_url` is sourced from the live esphome.io docs index, so a present URL is also a guarantee that the page exists.

## Test plan

- [x] New `tests/test_components_integration_docs.py` covers all three layers against the shipped catalog (top-level / category landing / stem alias) plus a guard for unknown names being absent.
- [x] Full backend test suite passes: 273 passed, 3 skipped.
- [x] Lint clean.
- [ ] Frontend PR (separate) consumes the endpoint.

### DIFF
--- a/esphome_device_builder/controllers/components.py
+++ b/esphome_device_builder/controllers/components.py
@@ -49,7 +49,12 @@ class ComponentCatalog:
             )
             return
 
-        data = json.loads(_COMPONENTS_JSON.read_text())
+        # ``encoding="utf-8"`` is explicit because the catalog carries
+        # non-ASCII characters (em-dashes, mu, etc.) and Path.read_text
+        # defaults to the platform's locale encoding — Windows' cp1252
+        # then dies on the first multi-byte sequence (UnicodeDecodeError
+        # at the catalog load).
+        data = json.loads(_COMPONENTS_JSON.read_text(encoding="utf-8"))
         self._components = [_load_component(c) for c in data.get("components", [])]
         self._by_id = {c.id: c for c in self._components}
         _LOGGER.info("Component catalog loaded: %d components", len(self._components))
--- a/esphome_device_builder/controllers/components.py
+++ b/esphome_device_builder/controllers/components.py
@@ -78,6 +78,73 @@ class ComponentCatalog:
         """Get all component categories with counts."""
         return self.categories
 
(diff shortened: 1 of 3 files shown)


bdraco:
Addressed all four Copilot points in 6e70387:

1. **Stem ambiguity** — track `stem → set[str]` of candidate URLs and only emit a stem alias when every category that owns the stem agrees on the same docs URL. Multi-platform components (`at581x`, `rotary_encoder`) still resolve because they share one docs page across categories, but genuinely distinct variants like `binary_sensor.gpio` vs `switch.gpio` are dropped — the frontend then renders them as plain text. New `test_ambiguous_stems_omitted` guards the `gpio` case.

2. **Permissive stem-match test** — pinned to `endswith("/components/sensor/ltr390")` so a regression that picks a wrong category for the stem fails instead of trivially passing on substring.

3. **Conditional `api` assertion** — dropped the `if "api" in docs:` guard so the test fails outright if `api` ever drops out.

4. **Confusing update-order comment** — reworded to say what `dict.update()` actually does (later writes win), so layering stems first / category second / top-level last lets the most-specific page win for any colliding key.

PR #58, opened 2026-05-02
Route every JSON callsite through helpers/json (orjson under the hood)

bdraco:
## Summary

The package was a mix of stdlib `json` (slow, plus the platform-locale-encoding trap on `read_text` that bit Windows recently) and a sprinkling of direct `orjson` imports. Centralise on `helpers/json` with three thin wrappers (`loads`, `dumps`, `dumps_indent`) and a re-exported `JSONDecodeError`, then route every callsite through them. `orjson` is now pinned in exactly one file; swapping the underlying serialiser later (or downgrading on a platform that can't build the wheel) is a one-line change.

### How upstream does it

For reference: ESPHome's own dashboard (`esphome/dashboard/core.py`, `dashboard/web_server.py`, `dashboard/status/mqtt.py`) uses stdlib `json` directly with no abstraction. We're deliberately a step above parity here — the dashboard isn't a hot path in their architecture, but it is in ours (WS dispatch, ~896-component catalog load, MQTT discovery payloads, sessions persistence).

### Touched callsites

| File | What |
| --- | --- |
| `api/ws.py` | WS dispatch (already orjson, now via the helper). |
| `api/legacy.py` | Legacy spawn-protocol body parsing. |
| `controllers/components.py` | 896-component catalog load — read_bytes + orjson skips the read_text → encode round-trip. |
| `controllers/config.py` | Metadata sidecar atomic RMW (`dumps_indent` keeps the on-disk file readable / diffable). |
| `controllers/devices.py` | `ignored-devices.json`. |
| `controllers/editor.py` | Line-delimited JSON to the esphome vscode validator subprocess. |
| `controllers/_device_mqtt_monitor.py` | `esphome/discover/<name>` payloads. |
| `helpers/auth.py` | Sessions file (`.device-builder-sessions.json`). |

### Wire shape

`dumps()` returns `bytes` (orjson semantics), so file writes use `write_bytes` / binary-mode `os.fdopen` instead of text mode. The on-disk content is identical UTF-8.

## Test plan

- [x] `ruff check` clean (auto-fixed import order on three files).
- [x] Full backend suite: 274 passed, 3 skipped.
- [x] Sessions / metadata / ignored-devices files written before this PR can still be read after it (forward-compat — both shapes are valid JSON, the serialiser doesn't change the wire bytes).

### DIFF
--- a/esphome_device_builder/api/legacy.py
+++ b/esphome_device_builder/api/legacy.py
@@ -14,7 +14,6 @@ HA uses:
 from __future__ import annotations
 
 import asyncio
-import json
 import logging
 import sys
 from typing import Any
@@ -23,6 +22,7 @@ import aiohttp
 from aiohttp import web
 from esphome import yaml_util
 
+from ..helpers.json import loads
 from ..helpers.subprocess import create_subprocess_exec
 
 _LOGGER = logging.getLogger(__name__)
@@ -41,7 +41,7 @@ async def _handle_legacy_ws_command(
         if msg.type != aiohttp.WSMsgType.TEXT:
             break
 
-        data = json.loads(msg.data)
+        data = loads(msg.data)
         if data.get("type") != "spawn":
             continue
 
--- a/esphome_device_builder/api/legacy.py
+++ b/esphome_device_builder/api/legacy.py
@@ -22,7 +22,7 @@ import aiohttp
 from aiohttp import web
 from esphome import yaml_util
 
-from ..helpers.json import loads
+from ..helpers.json import JSONDecodeError, dumps_str, json_response, loads
 from ..helpers.subprocess import create_subprocess_exec
 
 _LOGGER = logging.getLogger(__name__)
@@ -41,8 +41,16 @@ async def _handle_legacy_ws_command(
         if msg.type != aiohttp.WSMsgType.TEXT:
             break
 
-        data = loads(msg.data)
-        if data.get("type") != "spawn":
+        try:
+            data = loads(msg.data)
+        except JSONDecodeError:
+            # Legacy clients shouldn't send non-JSON, but if one does
+            # we'd rather skip the frame than tear down the whole
+            # spawn handler with the next iteration losing its
+            # subprocess output.
+            _LOGGER.debug("Ignoring non-JSON frame on %s", request.path)
+            continue
+        if not isinstance(data, dict) or data.get("type") != "spawn":
             continue
 
         configuration = data.get("configuration", "")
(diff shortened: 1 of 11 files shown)


bdraco:
Quick note on dependencies: `orjson>=3.9.0` is already a top-level runtime dependency in `pyproject.toml` (line 18, in `[project].dependencies`, not an extra). The package was already shipping orjson — `api/ws.py` and `helpers/json.py` both used it before this PR. No new dependency is being added; this change just routes the rest of the JSON callsites (which had drifted to stdlib `json`) through the dep we already require.

bdraco:
Addressed all three Copilot points in 684c679. Also picked up your follow-up about aiohttp's `dumps=` / `loads=` slots:

1. **Unguarded `loads(msg.data)` in legacy WS** — wrapped in `try / except JSONDecodeError`; bad frames log a debug line and are skipped instead of taking down the spawn loop. Tightened `isinstance(data, dict)` before `.get()` on the same path.

2. **`_load_ignored_devices` not crash-resistant** — catches `JSONDecodeError`, validates the dict shape and the inner list type, and filters names to strings. A corrupt file logs a warning and starts with an empty set instead of failing controller bootstrap.

3. **Claim of full centralisation vs reality** — switched the remaining `web.json_response` callsites to the `helpers/json` wrapper, so HTTP responses serialise via orjson too. The `ws.send_json` callsites now use the new `dumps_str` helper via aiohttp's `dumps=` slot — `send_json(data, dumps=dumps_str)` reads cleaner than `send_str(dumps(data).decode())` and keeps the call site shaped like idiomatic aiohttp. `dumps_str` is just `orjson.dumps(obj).decode()` and lives in `helpers/json` for both WS and HTTP-response use.

Test_auth's WS stub now mocks `send_json(data, *, dumps=...)` to match the new send shape; 274 passed.

bdraco:
Addressed both round-2 Copilot points in affe3b7:

1. **WS stub didn't exercise serialisation** — `test_auth`'s `_send_json` mock now round-trips the payload through the supplied `dumps` callable (`dumps_str`), then `json.loads` the result. Mirrors aiohttp's real behaviour: only the serialised string hits the wire, so a payload that isn't JSON-serialisable (dict with a non-string key, etc.) now fails the test rather than slipping through.

2. **Silent return on non-list `ignored_devices`** — now logs a warning and explicitly resets `self.ignored_devices = set()` before returning, matching the dict-shape and corrupt-JSON paths.

Plus you asked for tests — added `tests/test_ignored_devices_load.py` with six cases covering every defensive path: missing file (silent), corrupt JSON (warn), non-object top-level (warn), non-list field (warn), mixed entry types (filter to strings), happy path. Each runs the sync loader through `asyncio.to_thread` to mirror `DevicesController.start`'s production schedule. 280 passed, 3 skipped.

PR #59, opened 2026-05-02
Bump default executor to 64 threads so DNS resolves don't stall the editor

bdraco:
## Summary

Yes — your hunch about executor thread count was right. The default `ThreadPoolExecutor` sizes itself to `min(32, os.cpu_count() + 4)` — about a dozen threads on a typical 8-core box. With 50+ devices, every ping sweep fans out a batch of `icmplib.async_resolve` calls (and asyncio's `getaddrinfo` underneath), all of which run in the default executor and can block on multi-second timeouts when devices are offline / mDNS-only.

When the pool saturates, **unrelated** executor work queues up behind the resolves: scanner file stats, YAML parses backing `devices/list`, secrets reads, etc. Opening the editor triggers several of those calls on page load, so the page hangs for several seconds while it waits for free workers. The log you posted shows the editor spawn at `20:59:48` and a continuous trickle of "unknown → offline (via ping)" before and after — those resolves were holding the pool.

## Fix

Set the default executor to a 64-thread `ThreadPoolExecutor` in `DeviceBuilder.start()`, right after capturing the running loop. The work the dashboard's executor sees is overwhelmingly blocked-on-I/O (DNS, file stat, subprocess wait, MQTT tcp connect), so a wider pool simply lets those run in parallel without stepping on the foreground request path. Thread name prefix `dashboard` so saturation is easy to spot in `py-spy` / thread dumps.

## Test plan

- [x] Lint clean.
- [x] Full backend suite: 274 passed, 3 skipped.
- [ ] With ~50 devices and a sweep mid-flight, opening the editor responds quickly (devices/list, components, etc. don't queue behind DNS resolves).
- [ ] Idle memory footprint is unchanged in practice — `ThreadPoolExecutor` only spawns threads on demand.

## Follow-ups (not in this PR)

If the editor stall persists after this lands, the next steps are:

- A dedicated executor for the network-bound work (DNS / pings) so it can't even compete with file I/O — keeps total threads bounded but isolates the noisy-neighbour pattern.
- Cache failed DNS resolves more aggressively to skip the timeout entirely on the next sweep.

### DIFF
--- a/esphome_device_builder/device_builder.py
+++ b/esphome_device_builder/device_builder.py
@@ -97,6 +97,18 @@ class DeviceBuilder:
 
         self._ingress_runner: web.AppRunner | None = None
 
+    def _install_default_executor(self) -> None:
+        """Register the dashboard's executor as the loop's default.
+
+        Extracted so the unit test can drive the same registration
+        path the production ``start()`` flow uses, instead of
+        re-implementing ``loop.set_default_executor(self._executor)``
+        and trivially passing even when ``start()`` stopped doing it.
+        """
+        assert self.loop is not None  # type narrowing — set in start()
+        assert self._executor is not None  # type narrowing — set in __init__
+        self.loop.set_default_executor(self._executor)
+
     async def start(self) -> None:
         """Start the application — load catalogs, initialize controllers."""
         from .controllers.auth import AuthController
@@ -109,13 +121,11 @@ class DeviceBuilder:
         from .controllers.firmware import FirmwareController
 
         self.loop = asyncio.get_running_loop()
-        # The executor itself was constructed in ``__init__`` (so
-        # callers that probe ``self._executor`` before ``start()`` see
-        # the right value); here we just register it as the loop's
-        # default. See ``_EXECUTOR_MAX_WORKERS`` for the why behind
-        # the pool size.
-        assert self._executor is not None  # type narrowing
-        self.loop.set_default_executor(self._executor)
+        # Pool itself was constructed in ``__init__`` (so callers
+        # probing ``self._executor`` pre-start see the right value);
+        # here we just register it as the loop's default. See
+        # ``_EXECUTOR_MAX_WORKERS`` for the why behind the pool size.
+        self._install_default_executor()
 
         # Initialize controllers
         self.auth = AuthController(self)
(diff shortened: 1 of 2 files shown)


bdraco:
Addressed both Copilot points + your follow-up about the magic number, all in b41affb:

1. **Executor never shut down** — pool is now held on `self._executor` (constructed in `__init__` so callers probing it pre-start see the right one), and `stop()` calls `loop.shutdown_default_executor()` to drain in-flight tasks and join the worker threads. Without this a test or external loop-owner could leave threads up while long-lived monitors / DNS probes blocked loop close.

2. **`64` should be a constant** — extracted to `_EXECUTOR_MAX_WORKERS` at module level with a doc-block on the why (asyncio's `min(32, cpu+4)` is too tight for the I/O-bound workload, 64 leaves headroom on a saturated sweep without ballooning the thread table).

3. **No coverage** — new `tests/test_executor_pool.py` covers three regressions:
   - `__init__` creates a `ThreadPoolExecutor` sized to `_EXECUTOR_MAX_WORKERS` (silent removal of the constant fails the test).
   - `run_in_executor` under `set_default_executor` lands on a `dashboard*`-prefixed thread (the editor-stall regression is locked down — moving / removing the registration breaks this).
   - `stop()` drains and clears `_executor` (a second stop is a no-op, GC can collect).

277 passed, 3 skipped.

bdraco:
Addressed both round-2 Copilot points (and your nudge that `loop.shutdown_default_executor()` is the right call) in be4a76b:

1. **`stop()` could shut down the wrong pool** — defensively re-pin `self._executor` as the loop's default *before* `shutdown_default_executor()`, so even if a third party swapped the loop default between `start()` and `stop()` we still close the pool we own.

   I briefly tried `asyncio.to_thread(executor.shutdown, wait=True)` to target our pool explicitly, but as you pointed out — `asyncio.to_thread` *uses* the loop's default executor, so it'd try to schedule `shutdown(wait=True)` on the same pool we're closing and deadlock. `loop.shutdown_default_executor()` is the asyncio-engineered idiom: it doesn't route through the executor, waits for in-flight work, and joins the worker threads.

2. **Test didn't actually exercise `start()`'s registration** — extracted `_install_default_executor()` so the unit test drives the same helper production `start()` calls. Previously the test manually re-implemented `loop.set_default_executor(...)`, so a regression where `start()` stops registering the pool would still pass. Now the helper has to disappear from `start()` for the binding to be skipped — and `test_stop_drains_executor` also asserts the pool itself is shut down (submitting work after stop raises), not just that `self._executor` was nilled out.

Lint clean, 277 tests pass, 3 skipped.

bdraco:
Three new Copilot points, all addressed in 3cce5ec:

1. **`assert` in `_install_default_executor`** — asserts get stripped under `python -O`. Replaced with explicit `raise RuntimeError` for both `self.loop is None` ("call start() first") and `self._executor is None` ("already shut down") so a real bug surfaces with a clear message instead of a downstream `AttributeError` in the loop's guts.

2. **`._max_workers` is private API** — dropped the private-attribute access. The pool is sized from `_EXECUTOR_MAX_WORKERS` at construction time; pinning the literal in the test was over-specification *and* reached into a CPython internal that could be renamed. The test now sanity-checks the constant is a positive int and trusts the constructor passed it through.

3. **`stop()` leaked the pool when `loop is None`** — real bug: the pool is created eagerly in `__init__`, so an instance built and disposed without ever calling `start()` was leaving a live `ThreadPoolExecutor` (and possibly threads) behind. New branch in `stop()` calls `executor.shutdown(wait=False)` directly when `self.loop is None` — no tasks are in flight, so the to_thread/deadlock concern doesn't apply. Added `test_stop_without_start_drains_executor` to lock the leak fix down: build, stop, assert `pool.submit(...)` raises.

Lint clean, 284 tests pass, 3 skipped.

PR #111, opened 2026-05-02
Compress text/JS/JSON/SVG responses with isal-backed gzip

bdraco:
## Summary

The frontend wheel ships a ~3.5 MiB JS bundle that the backend served uncompressed. On a typical broadband link that's ~3 s of first-load wait that gzip would knock down to ~600-800 KiB (~85% reduction on minified JS). On a local LAN the wins are smaller but still meaningful for the SPA shell + JSON API.

Wires three pieces:

### 1. `aiohttp-fast-zlib` + `isal` deps

`aiohttp_fast_zlib.enable()` is called once in `DeviceBuilder.run()`. On aiohttp 3.13+ it swaps `aiohttp.compression_utils.ZLibBackend._zlib_backend` to `isal.isal_zlib` (Intel Storage Acceleration Library — hardware-accelerated zlib). This is the same dependency Home Assistant ships in core for the same reason.

Local benchmark on a 2.77 MiB JS-ish payload at level 1:
- stdlib zlib: 2.61 ms/iter
- isal_zlib:   1.08 ms/iter  (~2.4× faster)

### 2. `compression_middleware`

Calls `response.enable_compression()` for compressible responses. Skips for:
- Clients that don't advertise gzip/deflate
- Already-encoded responses (`.gz` / `.br` sidecars served by `FileResponse`, which sets `Content-Encoding` itself)
- Range requests (partial-content + on-the-fly compression aren't compatible in aiohttp)
- Non-compressible content types (firmware bins, opaque images)
- Tiny responses (< 1 KiB floor where framing dominates the savings)

### 3. Public-vs-trusted middleware split

| Site | Middlewares |
|---|---|
| Public | `cors → compression → auth` |
| Trusted Ingress | `cors` only |

The HA supervisor proxy compresses upstream of the trusted Ingress site, so re-encoding here would burn CPU twice for the same wire bytes (and the supervisor uses its own settings, so we'd be fighting it).

## Test plan

- [x] 11 new tests in `test_compression_middleware.py` covering each compress / skip branch + the public-vs-ingress wiring
- [x] 531 backend tests total — all green
- [x] `pre-commit run` — clean
- [x] Verified `ZLibBackend._zlib_backend` flips from stdlib `zlib` to `isal.isal_zlib` after `enable()`

### DIFF
--- a/esphome_device_builder/device_builder.py
+++ b/esphome_device_builder/device_builder.py
@@ -14,13 +14,14 @@ import re
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
+import aiohttp_fast_zlib
 from aiohttp import web
 
 from .controllers.config import DashboardSettings
 from .helpers.api import CommandHandler, collect_api_commands
 from .helpers.auth import auth_middleware
 from .helpers.event_bus import EventBus
-from .helpers.json import cors_middleware
+from .helpers.json import compression_middleware, cors_middleware
 
 if TYPE_CHECKING:
     from .controllers.auth import AuthController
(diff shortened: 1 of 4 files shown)


bdraco:
Closing — wrong call on my part. Both real-world deployments don't benefit:

- **HA add-on (Ingress)**: supervisor proxy already compresses upstream; this PR explicitly disables compression on the trusted site for that reason.
- **ESPHome Desktop / local**: bundle is served over loopback / LAN where the wire-size win is dwarfed by the CPU cost of compressing on every request.

Not worth the dep weight (`isal` adds a native wheel) for the niche case where someone exposes the dashboard over WAN without a reverse proxy in front.

PR #165, opened 2026-05-03
Cover error_response helper

bdraco:
## Summary
`error_response` was the only function in `helpers/json.py` without direct test coverage. Two cases for the dashboard's universal error envelope:

- `error_response(message)` → HTTP 400 + `{"error": message}` body with `application/json` content type.
- `error_response(message, status=...)` honours the override for endpoints that want 403 / 404 / 409 with the same shape.

Pins the contract so a refactor that flipped the default status or the body shape surfaces here rather than at the first endpoint that misbehaves.

## Test plan
- [x] `uv run pytest tests/test_helpers_json.py` — 5 passed
- [x] `uv run pre-commit run` clean

### DIFF
--- a/tests/test_helpers_json.py
+++ b/tests/test_helpers_json.py
@@ -14,6 +14,7 @@ import pytest
 from esphome_device_builder.helpers.json import (
     dumps,
     dumps_str_non_str_keys,
+    error_response,
     loads,
 )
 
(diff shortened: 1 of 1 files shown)


bdraco:
Closing — survey shows no callers of `error_response` in the codebase. Replacing with a removal PR.

PR #186, opened 2026-05-03
Emit auto-name on featured-component YAML for HA entity platforms

marcelveldt:
testing the sonoff basic featured "Front-Panel Button" produced a YAML block with neither `name:` nor `id:` — so the entity showed up unnamed in home assistant. same pattern affects every featured `binary_sensor.gpio` (and a chunk of `sensor.*` whose schema sync is missing the entity-base `name` field).

root cause is two layers:

- `_featured_field_presets` in `controllers/components.py` already auto-derives `id` and `name` defaults from the manifest's local id and display name. the materialised view (`get_component`) used these.
- `_apply_featured_presets` in `controllers/devices/helpers.py` — the path that actually builds the YAML — was reading `record.featured.fields` directly instead of the auto-augmented preset map, so the auto defaults never reached the merge step.

## changes

- route both call sites through `_featured_field_presets(record)` so manifest-explicit fields and auto-derived `id` / `name` defaults flow into the YAML the same way they flow into the catalog entry
- gate the auto-name preset on a new `_supports_name` check (HA entity domains + components that already declare `name:`) so non-entity platforms (`output:`, `i2c:`, `rtttl:`, ...) stay clean — they don't accept a top-level `name:` field
- small bonus: dedup `_generate_id` so a featured display name that leads with the chip name produces `id: hlw8012_power_monitor` rather than `id: hlw8012_hlw8012_power_monitor`
- two regression tests cover the `binary_sensor.gpio` name injection and the non-entity `output.gpio` exclusion

audited all 15 featured components on the 3 hand-curated boards — entities (`sensor`, `binary_sensor`, `switch`, `light`) now get both `id` and `name`; non-entities (`output`, `bus`, `misc`) correctly skip `name`.

note: the underlying sync issue — `binary_sensor.gpio` and a chunk of `sensor.*` are missing the entity-base `name` config_entry — is a separate concern in `script/sync_components.py`. addressed at runtime here so user YAML is correct today; a sync fix would additionally restore the `name` input field in the dashboard UI for those platforms.

### DIFF
--- a/esphome_device_builder/controllers/components.py
+++ b/esphome_device_builder/controllers/components.py
@@ -481,7 +481,7 @@ def _materialise_featured(
     """
     underlying = record.underlying
     fc = record.featured
-    presets = _featured_field_presets(fc)
+    presets = _featured_field_presets(record)
     return ComponentCatalogEntry(
         id=record.full_id,
         name=fc.name or underlying.name,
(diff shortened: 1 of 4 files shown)


marcelveldt:
closing in favor of the manifest-explicit approach — adding `fields.id` / `fields.name` to the definitions directly (and updating the devices.esphome.io sync to emit them) is cleaner than the runtime auto-derivation here. new pr coming.

PR #421, opened 2026-05-08
Add per-sub-field platform gating to ConfigEntry catalog

### LINKED ISSUE #417: [Bug] Component form doesn't gate sub-fields by device platform (e.g. ESP8266 + sensor.debug.psram)
## What

Catalog entries that bundle multiple optional sub-sensors / sub-fields don't gate those sub-fields by what the device's platform can actually populate. The "Add Component" form shows every sub-field in a single component's schema regardless of whether the user's MCU supports it.

Concrete example — `sensor.debug` carries optional sub-sensors:

| Sub-field | Platforms it works on |
|---|---|
| `free` (free heap) | all |
| `block` (max heap block) | esp32, esp8266 |
| `loop_time` | all |
| `cpu_frequency` | esp32, esp8266 |
| `psram` | **esp32 only** (and only when the variant has PSRAM) |
| `fragmentation` | esp8266 only |

If the user picks an ESP8266 board, the form happily lets them fill in `psram:`. Compile fails with a "doesn't apply on this platform" error well after the user's left the dialog.

Conversely, if the user picks an ESP32, the form shows `fragmentation:` even though it's an ESP8266-only field.

This was flagged as a sub-observation inside #383 ("filtered by cpu support ideally"). Splitting it out so the dep-detour fix on that issue (PR esphome/device-builder-frontend#224) can land on its own merits.

## Why this isn't a catalog-shape bug

The catalog mirrors ESPHome's upstream schema — `sensor.debug` is one component with N optional sub-fields, not N separate components. The granularity is correct; what's missing is per-sub-field platform gating *inside the form's renderer*, not in the catalog entries themselves.

The catalog entry's existing `supported_platforms` field gates the *whole component*; we'd need a parallel per-`config_entry` gate for sub-fields. Two possible shapes:

1. **Add `supported_platforms` to nested `ConfigEntry` records** in the catalog and have the form's renderer hide entries whose platform set excludes the device's MCU. Requires the sync script (`script/sync_components.py`) to extract the per-field platform info from the schema bundle.
2. **Form-side hardcoded rules** for the small set of components that have this shape (`debug`, possibly `wifi.power_save_mode`, …). Cheaper but doesn't scale.

(1) is the principled fix. The sync script already plumbs `supported_platforms` at the component level — extending to per-entry is bounded, schema-driven work.

## Reproduction

1. Create or open a device with an `esp8266:` platform block (e.g. a Sonoff S20 board).
2. "Add Component" → search "debug" → pick `sensor.debug`.
3. Form shows `psram:` as one of the optional sub-sensor fields.
4. Fill it in, save, install.
5. Compile fails: `psram` requires ESP32.

Expected: `psram:` doesn't render for an esp8266 device. Same for the inverse on ESP32 + `fragmentation:`.

## Affected components (non-exhaustive)

Components in the catalog whose sub-fields aren't all valid on every platform their parent supports — `sensor.debug`, `text_sensor.debug`, possibly any component with conditional schema branches keyed on platform. A `script/check_catalog.py`-style audit could enumerate the full set.

## Out of scope

- Catalog granularity (one debug entry vs many) — that's #383's other observation, resolved as expected behaviour.
- The dep-detour bug — fixed by esphome/device-builder-frontend#224.

bdraco:

Backend half of esphome/device-builder#417 — per-sub-field platform gating for the visual editor.

The schema bundle doesn't reliably surface per-field platform constraints. The upstream `debug` sensor's `psram` and `fragmentation` sub-sensors live behind `CORE.is_esp32` / `CORE.is_esp8266` branches at codegen time that don't show up in the static schema dump. Today the visual editor happily lets a user fill in `sensor.debug.psram` on an ESP8266 board, then they watch the compile fail three minutes later with a "doesn't apply on this platform" error well after they've left the dialog.

The fix: add a `supported_platforms` field to `ConfigEntry` (sibling-but-different from the existing `ComponentCatalogEntry.supported_platforms` — that one gates the *whole component*; this one gates a single config entry within an otherwise platform-portable component), plus a small explicit override map in `script/sync_components.py` keyed on `(component_id, *path)`. The walker supports nested paths so future entries can target deeper fields without restructuring the map.

**Starter overrides:**
- `("sensor.debug", "psram") → ["esp32"]` — PSRAM stats are ESP32-only with PSRAM-enabled variants
- `("sensor.debug", "fragmentation") → ["esp8266"]` — heap-fragmentation reporting is ESP8266-specific

The wire contract is "empty list = all platforms" (the common case), "non-empty = restrict to listed chips." The model default is `field(default_factory=list)` so existing serialized payloads stay compatible without a migration.

The frontend half (companion PR — drafted separately) reads `supported_platforms` in the form renderer and hides entries whose target chip isn't listed. Until that lands the field is dormant — no behaviour change for existing users.

**Why a hand-curated map and not a schema-side signal.** The platform branches in upstream debug aren't a schema construct; they're imperative codegen guards. Surfacing them through the schema bundle would require an upstream design change (PR/issue not yet filed). The override list is small and targeted — every entry is a workaround for a foot-gun a user actually hit. We can drop the override list once upstream surfaces the gates uniformly.

**Tests.** 6 new tests in `tests/test_sync_components_field_platform_overrides.py`:
- Pin both starter override values (regression-proof against silent map edits)
- Walker stamps the override on a top-level field and leaves siblings untouched
- Walker reaches nested paths via a synthetic test-only entry
- No-op on components without overrides
- Stamping doesn't disturb other entry fields (`type`, `label`, `config_entries`)

`components.json` is **not** regenerated in this PR — the sync script supports the new override but the next nightly catalog sync will pick it up. Doing the regeneration in this PR would mix the schema sync with the override addition and make the diff much harder to review.

- fixes esphome/device-builder#417 (backend half — frontend companion PR pending)

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-dashboard-frontend#<TBD — to be opened next>

The new `ConfigEntry.supported_platforms` field is the wire contract; the frontend reads it in the form renderer to hide entries whose target chip isn't in the list. Until the FE PR ships the field is dormant.

## Checklist

### DIFF
--- a/esphome_device_builder/models/common.py
+++ b/esphome_device_builder/models/common.py
@@ -294,6 +294,24 @@ class ConfigEntry(DataClassORJSONMixin):
     # commonly used there).
     platform_defaults: dict[str, ConfigPrimitive] | None = None
 
+    # Target chips this field is valid on. Empty list means "all
+    # platforms" (the common case). Non-empty restricts the field
+    # to the named chips — e.g. ``sensor.debug.psram`` is
+    # ``["esp32"]`` because PSRAM stats only exist on ESP32 with
+    # PSRAM-enabled variants, and ``sensor.debug.fragmentation``
+    # is ``["esp8266"]`` because heap-fragmentation reporting is
+    # ESP8266-specific. The form renderer hides entries whose
+    # ``supported_platforms`` is non-empty and doesn't include
+    # the device's target chip; a preemptive filter beats the
+    # current behaviour where the user fills the field, compiles,
+    # and then sees a "doesn't apply on this platform" error well
+    # after they've left the dialog. Sibling-but-different from
+    # the catalog-entry-level ``supported_platforms`` on
+    # :class:`ComponentCatalogEntry` (that one gates the *whole
+    # component*; this one gates a single config entry within an
+    # otherwise platform-portable component).
+    supported_platforms: list[str] = field(default_factory=list)
+
     # === value constraints ===
 
     # Constrains the value to a fixed set of choices. When populated the
--- a/script/sync_components.py
+++ b/script/sync_components.py
@@ -1598,6 +1598,7 @@ def build_component_entry(
     _apply_platform_defaults(config_entries, introspection.get("platform_defaults") or {})
     _apply_refined_types(config_entries, introspection.get("refined_types") or {})
     _apply_unit_of_measurement_options(config_entries)
+    _apply_field_platform_overrides(component_id, config_entries)
 
     return {
         "id": component_id,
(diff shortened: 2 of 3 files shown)


bdraco:
Closing — the hand-curated `_FIELD_PLATFORM_OVERRIDES` map is unmaintainable. The right solution is to derive per-field platform gating from the schema (live introspection at sync time), not a manual override list. Will reopen with a schema-derived implementation.

Issue #106, opened 2026-05-02
Remote build offload: discover and delegate compiles to another dashboard on the LAN


bdraco:
# Remote build offload — feature complete

**Status:** Shipped. Closing as completed; remaining gaps tracked as separate issues.

The transparent install flow is user-visible end-to-end on `main`: clicking Install on a paired offloader auto-routes to an idle paired receiver, queues on a busy one before falling back to LOCAL, and the operator can override mid-flight via the install dialog's "Build locally instead" link. Bytes always flow back through the user's offloader for the actual flash (the receiver only ever compiles).

## What shipped

- **Pairing & transport** — Noise XX over plain-TCP WebSocket, two-way OOB pin verification, pairing window gate. Identity is a 32-byte X25519 keypair persisted at `.device-builder-peer-link-key.bin`; cert + key files no longer used and removed from disk (#598).
- **Long-lived peer-link sessions** — one Noise WS per paired receiver, encrypted heartbeat, auto-reconnect with exponential backoff. Application messages: `queue_status` push, `submit_job` + bundle chunk stream + ack, `cancel_job`, `download_artifacts` artifact-tarball pull, `job_state_changed` + `job_output` per-job fan-out.
- **Build scheduler** — `helpers/build_scheduler.pick_build_path` picks an APPROVED + connected + idle paired receiver, deterministic sort with pin-sort tiebreaker, fail-closed on unknown peer states.
- **Source-routed firmware runner** — `FirmwareJob.source: JobSource` discriminator + `source_pin_sha256` + `source_label`. `_execute_job` branches; the REMOTE branch dispatches `submit_job{target: "compile"}`, awaits the receiver's terminal frame, pulls artifacts, and spawns a local `esphome upload --file <staged>` to flash.
- **Install handler integration** — `firmware/install` routes through `pick_build_path`; the install dialog reads `job.source_label` and shows "Building on {receiver_label}" when present. Cancel translates `remote_build/cancel_job` to the receiver; mid-build session loss surfaces a synthetic "receiver lost connection" line in the log dialog.
- **Endpoint rebind** — automatic via mDNS (#539) and user-driven via `remote_build/edit_pairing_endpoint` (#548). Both share probe + commit primitives. Pin-mismatch refuses the rebind; the re-auth wizard exists to gate that case.
[...]

PR #467, opened 2026-05-09
Add backend pairing storage + WS commands (#106 phase 4a)

bdraco:

Offloader-side persistence for paired-remote records and the four WS commands the frontend Pair flow consumes. This is the backend half of phase 4 from issue #106; the frontend Pair dialog (4b) lands separately and consumes this surface.

## WS surface

- **`remote_build/preview_pair({hostname, port})`** — opens a TLS handshake to the candidate receiver, returns `{pin_sha256}`. No bearer presented, no persistence; the frontend renders the fingerprint for the user to OOB-confirm against what the receiver's Build server settings page shows. Errors: `UNAVAILABLE` (transport failure), `INVALID_ARGS`.

- **`remote_build/confirm_pair({hostname, port, pin_sha256, token, label, server_version, esphome_version})`** — re-handshakes (defends against a TOCTOU between preview and confirm), asserts the new handshake's pin matches the user-confirmed value, presents the user-pasted bearer against `GET /remote-build/v1/health` with this dashboard's stable `X-Dashboard-ID`, and persists the `StoredPairing` only when both succeed. Typed errors:
  - `PRECONDITION_FAILED` — re-handshake observed a different pin (cert rotation between preview and confirm, or active MITM)
  - `UNAUTHORIZED` — receiver rejected the bearer (wrong token, revoked, or 403 because token is bound to a different dashboard already)
  - `UNAVAILABLE` — transport failure
  - `ALREADY_EXISTS` — `(hostname, port)` already paired
  - `INVALID_ARGS`

- **`remote_build/unpair({hostname, port})`** — removes a paired record. `NOT_FOUND` if no match.

- **`remote_build/list_pool()`** — returns `PairingSummary` rows projected from persisted `StoredPairing`s. Drops `token_cleartext` from the wire view, same `StoredToken` → `TokenSummary` projection pattern.

## Storage

Extends `RemoteBuildSettings` with `paired_remotes: list[StoredPairing]`. Same `remote_build_settings_transaction` lock as the existing manual-hosts / tokens paths; same `_to_view` projection that already drops on-disk-only fields when settings cross the wire.

`StoredPairing` carries `{hostname, port, label, pin_sha256, token_cleartext, dashboard_id, server_version, esphome_version, paired_at}`. `PairingSummary` (wire) drops `token_cleartext`.

## TLS plumbing

New module `helpers/remote_build_pairing.py`:
- `observe_remote_pin(host, port)` — opens a TLS connection, returns SPKI-SHA-256 fingerprint as lowercase hex (matches `helpers.dashboard_identity._spki_fingerprint`'s output byte-for-byte; pinned by a unit test).
- `verify_bearer(host, port, expected_pin, token_cleartext, dashboard_id)` — re-handshake with pin assertion + authenticated `GET /health`. `PinMismatchError` on divergence; `PairingHealthResult{ok, http_status}` on the request side.

Uses an SSLContext with chain / hostname checking disabled. The pair flow's premise is that the receiver ships a self-signed cert and we pin its fingerprint OOB; OS trust store would reject the cert outright, hostname check would reject any IP-only or `foo.local` URL.

## ErrorCode additions

`UNAVAILABLE`, `UNAUTHORIZED`, `PRECONDITION_FAILED` — distinct from the existing codes:
- `NOT_FOUND` ≠ `UNAVAILABLE` (we couldn't even reach the remote)
- `NOT_AUTHENTICATED` ≠ `UNAUTHORIZED` (the *remote* rejected our bearer, not the dashboard's WS auth gate)
- `INVALID_ARGS` ≠ `PRECONDITION_FAILED` (the pin was valid at preview, just changed under us)

## Bearer-on-wire / bearer-at-rest decision (captured to skip re-litigating)

This was the longest design discussion of the phase. Short version: the cleartext bearer crosses the offloader's WS in `confirm_pair` and lands on disk in cleartext. Considered RSA-OAEP via receiver pubkey, PAKE setup code, auto-HTTPS using the 3a cert — all add real surface and don't move the needle for the actual deployment shapes.

- **ESPHome Desktop** runs locally; user's browser hits `http://localhost:6052`, traffic doesn't traverse the network at all.
- **HA addon** uses HA's ingress (wss:// + HA's auth posture); the bearer leak surface is identical to anything else the user puts in HA.
- **Standalone Docker / pip install accessed remotely over plain HTTP** is niche and already accepts plaintext for device API keys, Wi-Fi YAML, secrets.yaml.

Full discussion captured in issue #106 phase 4a section. The retrospective on 3b3 (which client-side-minted bearers for arguably the same threat we're not defending against here) is in the same issue under the 3b3 design note.

**Related issue:**
- esphome/device-builder#106

## Types of changes

- [ ] Breaking change — `breaking-change`
- [ ] Refactor — `refactor`

## Frontend coordination

- [x] Companion frontend PR: phase 4b (Pair dialog + pool rendering) lands separately once this merges.

## Checklist

- [x] Tests have been added: 31 new tests in `tests/test_remote_build_controller.py` (validators + 4 WS commands incl. error-path coverage) + 5 in new `tests/test_remote_build_pairing.py` (TLS helper + SPKI-fingerprint cross-check against `dashboard_identity`).

## Notes for reviewers

- End-to-end TLS test against a real receiver instance is queued for the cross-cutting E2E harness (issue #106's "End-to-end test infrastructure" section); the controller-level tests use monkeypatched `observe_remote_pin` / `verify_bearer` because spinning up a paired dashboard pair in a unit-test fixture would land that harness twice.
- The duplicate-pair check happens both before the TLS round-trip (fast-path readonly) and inside the persistence transaction (race-safe). The fast-path saves the slow round-trip on the common "user clicked Pair on an already-paired entry by mistake" case.


bdraco:
Abandoning this branch — phase 4a is being redesigned away from the bearer-token auth model toward mTLS with mutual cert pinning. The bearer infrastructure (3b1 / 3b2 / 3b3 + the Fernet sealing this PR added in 4a) ends up being replaced rather than extended. Full design in #106's new "Auth model pivot" section. See you in the new 4a.

PR #532, opened 2026-05-10
Phase 4a-o part 6: re-key offloader-side state on pin_sha256

bdraco:

Phase 4a-o part 6 from
[#106](https://github.com/esphome/device-builder/issues/106) —
the refactor that re-keys offloader-side state from
`(receiver_hostname, receiver_port)` to `pin_sha256`. The pin
is the stable cryptographic identity (hash of the receiver's
static X25519 pubkey); network coords are routing hints stored
as fields on the value rather than as the primary key.

## Why

A receiver hostname rename used to require a multi-dict atomic
remap (pop+insert across four state structures, cancel +
respawn tasks, error-prone). Under pin-keying it becomes a
one-line mutation of the value's `receiver_hostname` field.
Symmetric to the receiver-side `_approved_peers` keyed on
`dashboard_id` — both ends index by stable cryptographic
identity with network coords as fields.

This PR is the prep for 4a-o part 7 (auto-rebind on mDNS
pin-match): with pin-keyed registries the rebind is a one-line
value mutation of `_pairings[pin].receiver_hostname` instead of
a multi-dict atomic remap.

## Scope

**State structures (RAM-only, all controller-side):**

- `_pairings: dict[str, StoredPairing]`
- `_pair_status_listeners: dict[str, asyncio.Task]`
- `_peer_link_clients: dict[str, asyncio.Task]`
- `_offloader_alerts: dict[str, OffloaderAlertSnapshotEntry]`
- `_peer_queue_status: dict[str, PeerQueueStatusSnapshotEntry]`

All five pivot to `pin_sha256` (lowercase-hex SHA-256 of the
receiver's X25519 pubkey) as the primary key.

**Wire shape changes (event payloads):**

`OffloaderPairStatusChangedData` /
`OffloaderPeerLinkOpenedData` / `OffloaderPeerLinkClosedData`
/ `OffloaderPairAlertDismissedData` /
`OffloaderPairPinMismatchData` /
`OffloaderPairPeerRevokedData` /
`OffloaderQueueStatusChangedData` all gain `pin_sha256: str`
so listeners (the controller's own and the frontend's once it
migrates) can look up the row by pin without a follow-up scan.
Receiver coords stay on the payload as display fields the
frontend can render without a second round-trip.

**Wire shape changes (snapshots):**

`OffloaderPinMismatchAlert` / `OffloaderPeerRevokedAlert` /
`PeerQueueStatusSnapshotEntry` rows in
`subscribe_events.initial_state.*` similarly gain `pin_sha256`.

**WS command shape change:**

`unpair` takes `pin_sha256: str` instead of
`hostname: str, port: int`. Frontend already carries
`pin_sha256` on every `PairingSummary`, so the user-clicks-Unpair
path threads that value directly. **Breaking** for the WS shape
— fine pre-release per the no-back-compat memory.

**Internal contract additions:**

- `_sweep_stale_pairings_at_endpoint(host, port,
  keep_pin_sha256=...)` — new helper called by `request_pair`
  to clean up rows under stale pins at the same `(host, port)`
  (receiver rotated identity, or a different receiver moved to
  that hostname). Walks both `_pairings` and
  `_offloader_alerts` so a pre-existing alert under the old
  pin gets dismissed when the user re-pairs under a new pin
  at the same coords.
- `_dismiss_offloader_alert` / `_cancel_pair_status_listener`
  / `_cancel_peer_link_client` now take `pin_sha256` instead
  of `(hostname, port)`.
- `_fire_offloader_pair_status_changed` /
  `_fire_offloader_pair_pin_mismatch` /
  `_fire_offloader_pair_peer_revoked` thread `pin_sha256`
  through to the event payload.
- `PeerLinkClient` constructor takes `pin_sha256` and fires it
  on every event (`OFFLOADER_PEER_LINK_OPENED` / `_CLOSED` /
  `OFFLOADER_PAIR_PIN_MISMATCH` /
  `OFFLOADER_QUEUE_STATUS_CHANGED`).

## On-disk format unchanged

The persisted shape was already a list serialised from the
dict (`OffloaderRemoteBuildSettings.pairings: list[StoredPairing]`),
so re-keying the in-RAM collection doesn't touch disk. Existing
on-disk pairings load cleanly under the new keying.

## Tests

Every existing test that seeded `_pairings[(host, port)]`
migrates to `_pairings[pin_sha256]`; every event-payload
assertion gains `pin_sha256`; every `unpair(hostname, port)`
call becomes `unpair(pin_sha256)`. 2484 tests pass; the touched
suites (controller / peer-link / event-payload contracts /
e2e) are all green.

## Frontend coordination

Frontend follow-up will:

- Pivot `_buildOffloadPairings` keying from
  `${hostname}:${port}` to `pin_sha256`.
- Update `unpairRemoteBuild` API client to take pin.
- Consume the new `pin_sha256` field on each event payload's
  TypedDict.

Default-`""` defensive deserialisation could keep the frontend
working against a backend running this PR's wire shape, but
the cleaner path is a lockstep frontend PR. Drafted PR
[#529](https://github.com/esphome/device-builder/pull/529)
(offloader-connected indicator) is gated on this PR landing
+ rebasing on top.

- part of #106 (4a-o part 6).

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`

(Tagged `breaking-change` because the `unpair` WS arg shape
and several event-payload shapes change. Pre-release, no
installed base; the breaking-change label is for the
release-notes section.)

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-dashboard-frontend#<TBD>

Without the frontend mirror, the existing `${hostname}:${port}`
lookups against the new event payloads still work (host/port
stay on the payload), and `unpair` calls fail until the
frontend ships the new arg shape.

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/remote_build.py
+++ b/esphome_device_builder/controllers/remote_build.py
@@ -753,9 +753,12 @@ class RemoteBuildController:
         # ``subscribe_events`` stream — no separate
         # ``subscribe_pairings`` channel needed.
         #
-        # Keyed on ``(receiver_hostname, receiver_port)`` to match
-        # the unified ``_pairings`` dict.
-        self._pair_status_listeners: dict[tuple[str, int], asyncio.Task[None]] = {}
+        # Keyed on ``pin_sha256`` to match the unified
+        # ``_pairings`` dict (4a-o part 6 — re-keyed offloader
+        # state from ``(host, port)`` to pin so a receiver
+        # rename is a one-line value mutation rather than a
+        # multi-dict atomic remap).
+        self._pair_status_listeners: dict[str, asyncio.Task[None]] = {}
         # PENDING StoredPeer rows live here, keyed on dashboard_id.
         # Never persisted — the per-file ``_peers_store``
         # (``.receiver_peers.json``) only stores APPROVED rows.
@@ -800,8 +803,8 @@ class RemoteBuildController:
         # doubling. Drained in :meth:`stop`.
         self._peer_link_sessions: dict[str, PeerLinkSession] = {}
         # Offloader-side long-lived peer-link client tasks, one
-        # per APPROVED ``StoredPairing``, keyed on the receiver's
-        # ``(hostname, port)``. Spawned by
+        # per APPROVED ``StoredPairing``, keyed on the
+        # receiver's ``pin_sha256``. Spawned by
         # :meth:`_spawn_peer_link_client` from :meth:`start`'s
         # cold-start path and from
         # :meth:`_apply_pair_status_result` flipping a row to
@@ -809,7 +812,7 @@ class RemoteBuildController:
         # on ``unpair``; drained in :meth:`stop`. Each task runs
         # the connect-handshake-park-reconnect loop in
         # :meth:`PeerLinkClient.run`.
-        self._peer_link_clients: dict[tuple[str, int], asyncio.Task[None]] = {}
+        self._peer_link_clients: dict[str, asyncio.Task[None]] = {}
         # Identities cached once at :meth:`start` so each
         # peer-link client can pick them up without an executor
         # hop on every spawn. ``_offloader_dashboard_id`` is the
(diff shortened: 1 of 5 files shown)


bdraco:
Copilot's suppressed-confidence comment on `test_unpair_drops_pending_dict_entry_and_cancels_listener` was correct — the post-`unpair` listener lookup used the old `(hostname, port)` tuple key, so under the pin-keyed registry the `get(...)` always returned `None` and the test no longer verified the listener was actually cancelled.

Fixed in 297366b5: capture the task reference before `unpair()`, then drain via `asyncio.gather(task, return_exceptions=True)` and assert `task.cancelled()`. 862b044e applies the same gather-and-drain shape to two other touched tests (`test_request_pair_repair_then_unpair_clean_state`, `test_request_pair_repair_against_pending_cancels_old_listener`) for consistency with production's `RemoteBuildController.stop()`. Untouched cancel-drain sites in the same file are left alone per the project's no-churn-on-unrelated-tests rule.

PR #586, opened 2026-05-11
Pre-clean macOS .DS_Store / AppleDouble sidecars before bundle extract

bdraco:

User-reported: after cancelling a remote install, the next
`submit_job` for the same device on the receiver fails with

```
[Errno 66] Directory not empty: PosixPath(
  '/Users/.../wD8XFQEYxwjRMhg.../apollo-r-pro-1-eth-5938e0/build'
)
```

The failure persisted across retries even after a 15s
wait, ruling out a subprocess-reap race.

## Root cause

macOS Finder / QuickLook re-creates `.DS_Store` inside
any directory it has viewed. The receiver-side build tree
under
`.esphome/.remote_builds/<dashboard>/<device>/build/`
ends up with a `.DS_Store` sidecar after the user (or any
indexing daemon) views it. Upstream
`esphome.bundle.prepare_bundle_for_compile` cleans the
non-preserved entries via `shutil.rmtree` before
extracting the new bundle:

```
scandir(build/) → [.DS_Store, <env>/]
unlink .DS_Store
rmtree <env>/ (5000+ files, multi-second)
rmdir build/   ← Finder re-created .DS_Store; rmdir → ENOTEMPTY
```

The Finder write happens during the multi-second
`shutil.rmtree(<env>/)` recursion, and lands before the
final `rmdir(build)`.

Confirmed locally — `ls -la build/` after the failure
showed exactly one entry: `.DS_Store` at the top of
`build/`, alongside the still-present `<env>/` subdir.
Once Finder stopped re-creating it, a manual
`shutil.rmtree(build/)` succeeded.

## Fix

Pre-clean `.DS_Store` and AppleDouble (`._<name>`)
sidecars under `target_dir` before handing off to
`prepare_bundle_for_compile`. Shrinks the race window —
Finder can still re-create the sidecar mid-walk, but with
the sidecar gone at the start, the upstream walk has the
full multi-second window to finish without one present,
and the few-ms `scandir → rmdir` window at the end is
unlikely to race a fresh Finder write.

```python
_strip_macos_metadata(target_dir)
extracted = prepare_bundle_for_compile(bundle_path, target_dir)
```

## Performance

- Uses `os.walk` (delegates to `os.scandir` for cached
  `d_type`), mirroring the pattern in
  `helpers/build_size.py`. ~tenths of a second on a
  typical 5000-entry ESP-IDF build vs. ~seconds for
  `Path.rglob` with its per-entry stat syscalls.
- Best-effort: per-file unlink failures are swallowed +
  debug-logged so a permission error doesn't block the
  install — upstream's cleanup walk is the real line of
  defence and would catch the same files anyway.
- Single `info` log at the end summarising sidecar count
  + elapsed time, so the operator sees the pre-clean
  effectiveness without a per-file log line drowning the
  signal on a 200+ folder build tree.

## Tests

Three new in `tests/test_remote_build_submit_job.py`:

- Removes `.DS_Store` / AppleDouble sidecars at every
  level under `target_dir`, leaving real files +
  directories alone.
- No-op when `target_dir` doesn't exist (first submit
  case).
- Swallows unlink errors (best-effort contract).
- 44 submit-job tests still pass.

- Bug surfaced in the local dev loop on macOS while
  testing the transparent install flow
  (esphome/device-builder#106).

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/remote_build/submit_job.py
+++ b/esphome_device_builder/controllers/remote_build/submit_job.py
@@ -58,6 +58,8 @@ from __future__ import annotations
 import asyncio
 import binascii
 import logging
+import os
+import time
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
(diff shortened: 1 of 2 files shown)


bdraco:
not needed. better fix coming

PR #595, opened 2026-05-11
DNM: Log upload-subprocess output + JOB_PROGRESS fires for progress-bar debugging

bdraco:

**DO NOT MERGE — diagnostic logging only.**

Adds temporary `[PROGRESS-DEBUG]` `_LOGGER.info` lines to investigate why the firmware-tasks progress bar doesn't animate during the local upload phase of a REMOTE install, despite the seam reset that landed in #580.

The hypothesis being tested: does the local `esphome upload --file <staged>` subprocess actually emit `espota2.ProgressBar`'s `\rUploading: [...] NN%` frames on its stdout pipe? Code inspection says yes — `perform_ota` unconditionally calls `ProgressBar.update` on every 8 KB chunk — but the observed log on the offloader side shows only `INFO Uploading <path>` and no progress frames at all.

Log lines added:

- `helpers._fire_job_progress` — logs every `JOB_PROGRESS` fire (`job_id`, `progress`)
- `helpers._ingest_output_line` — logs every parsed percent: whether it advanced the gauge or got clamped, with the source line (truncated to 200 chars via `%r` so `\r`/`\n` are visible)
- `remote_runner._fetch_and_run_local_upload` — logs the seam reset entry
- `remote_runner._run_upload_subprocess` — logs the subprocess spawn (cmd), every line yielded from the upload subprocess (with `%r` so frame boundaries are unambiguous), plus EOF + exit code + total line count

How to use:

1. `gh pr checkout` this branch on the offloader build server.
2. Restart the dashboard.
3. Trigger a REMOTE install whose progress bar would otherwise look frozen.
4. Grep the dashboard logs for `[PROGRESS-DEBUG]` and share the lines from `upload subprocess spawning` to `upload subprocess EOF`.

The result tells us:

- If we see `upload-line` entries with `\r`-prefixed `Uploading: NN%` content but no matching `parsed progress=` / `fire JOB_PROGRESS` → the parse path is dropping them (parser bug).
- If we see no `upload-line` entries between `spawning` and `EOF` → the subprocess's stdout pipe truly doesn't carry the ProgressBar frames (upstream `esphome` issue, not ours).
- If we see `fire JOB_PROGRESS` fires advancing 0→100 → the backend is correct and the bug is in the frontend's rendering of the upload phase.

- diagnostic for #580 followup; do not merge

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/firmware/helpers.py
+++ b/esphome_device_builder/controllers/firmware/helpers.py
@@ -338,6 +338,7 @@ def _fire_job_progress(job: FirmwareJob, bus: EventBus, progress: int) -> None:
     where it's readable.
     """
     job.progress = progress
+    _LOGGER.info("[PROGRESS-DEBUG] fire JOB_PROGRESS job_id=%s progress=%s", job.job_id, progress)
     payload: JobProgressData = {"job_id": job.job_id, "progress": progress}
     bus.fire(EventType.JOB_PROGRESS, payload)
 
@@ -379,6 +380,20 @@ def _ingest_output_line(job: FirmwareJob, bus: EventBus, line: str) -> None:
     out_payload: JobOutputData = {"job_id": job.job_id, "line": line}
     bus.fire(EventType.JOB_OUTPUT, out_payload)
     progress = _parse_progress(line)
-    if progress is None or progress <= (job.progress or 0):
+    if progress is None:
         return
+    if progress <= (job.progress or 0):
+        _LOGGER.info(
+            "[PROGRESS-DEBUG] clamped parsed=%s current=%s line=%r",
+            progress,
+            job.progress,
+            line[:200],
+        )
+        return
+    _LOGGER.info(
+        "[PROGRESS-DEBUG] parsed progress=%s (was %s) line=%r",
+        progress,
+        job.progress,
+        line[:200],
+    )
     _fire_job_progress(job, bus, progress)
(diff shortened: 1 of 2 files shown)


bdraco:
Looks like a regression in ESPHome dev .. not here

bdraco:
Closing — the diagnostic logging served its purpose. Root cause turned out to be upstream in esphome, not in device-builder:

`esphome/helpers.py:ProgressBar` (added in esphome/esphome#14678) gates the entire output on `self.stream.isatty()`. When any dashboard (device-builder, legacy `esphome.dashboard`, future tooling) spawns `esphome upload` with `stdout=PIPE, stderr=STDOUT`, the child's stderr isn't a TTY, `enabled` is `False`, every `update()` early-returns, and zero `\rUploading: NN%` frames reach our `iter_lines_with_progress` reader. The dashboard-side seam reset from #580 still does its job — there's just nothing on the wire afterward to advance the gauge.

Empirical confirmation captured inside the HA addon: same `cat -A` capture shows every `_LOGGER.info` line present (`INFO Connecting...`, `INFO Uploading <path>`, `INFO Upload took ...`) and zero `^MUploading: NN%` chunks between `Connected` and `Upload took`. A standalone probe that calls `ProgressBar` directly on a piped stderr in the same container reproduces the empty output without any networking involved.

Fix needs to land upstream — write-up at `fix_progress_bar.md` in the working tree (one-line behavioural change to `ProgressBar.__init__`, three pytest cases for the regression matrix, no device-builder changes once it ships). Once that upstream PR merges and reaches an `esphome` release the addon picks up, the firmware-tasks progress bar will animate without further work on this side.

PR #599, opened 2026-05-11
Exit cleanly when launching supervisor disappears (parent watchdog)

bdraco:

Adds an opt-in **parent-process watchdog**: when the dashboard's launching supervisor disappears (clean exit, force-quit, crash, `kill -9`, etc.), the dashboard signals SIGTERM to itself, runs aiohttp's usual `on_cleanup` chain, and releases the listening port — instead of being reparented to launchd/init and silently holding port 6052 forever.

**Why this matters in practice:** the ESPHome Builder desktop app (Tauri) leaves orphan dashboard processes even on a **clean exit from the menu bar / tray's Quit menu** — verified by the reporter. The desktop app has a `block_on(daemon::stop())` chain that's meant to SIGTERM the dashboard's process group and SIGKILL after 5 s, but the orphan survives anyway. Whatever's going wrong on the desktop side (`block_on` racing `app.exit(0)`, `killpg` finding an unexpected pgid, SIGKILL queued but not delivered) is a separate problem worth chasing on its own — but the dashboard doesn't need to wait for that fix. With this watchdog, the dashboard takes ownership of its own lifecycle: the moment Tauri goes away (graceful, forced, or crashed), the dashboard's `os.getppid()` poll sees the reparent and self-terminates cleanly.

**Engagement is conservative.** `helpers.parent_watchdog.should_engage`:

- Honours `--exit-on-parent-changed` / `--no-exit-on-parent-changed` when explicitly passed.
- Otherwise auto-engages only when `sys.executable` contains the `io.esphome.builder` bundle marker (the desktop app's bundled Python on macOS / Linux / Windows).
- Other launchers — HA addon's s6 supervisor, systemd units, ad-hoc `python -m esphome_device_builder` from a terminal — stay un-watched by default, since they manage the dashboard's lifecycle themselves.

**Cross-platform implementation:**

- **Unix** (macOS / Linux): polls `os.getppid()` every 2 s. When the parent dies the kernel reparents the dashboard under PID 1 (or a configured subreaper); the ppid value changes and the next poll triggers self-shutdown.
- **Windows**: PPID does **not** change on parent death there (no `init` adopts orphans), and `os.kill(pid, 0)` would actually `TerminateProcess` the parent via CPython's Windows shim. We open a `PROCESS_SYNCHRONIZE` handle on the parent and block on it with `WaitForSingleObject` from a worker thread (via `asyncio.to_thread`); when the handle becomes signalled, the parent has exited.

Self-shutdown signals SIGTERM on Unix and `CTRL_BREAK_EVENT` on Windows — both routes hit aiohttp's existing signal handler so the normal `on_cleanup` chain runs (firmware queue drains, state persists, sites close).

The watchdog task is scheduled via the existing `create_background_task` helper, so it's auto-cancelled and gathered on normal `DeviceBuilder.stop()` paths and adds no extra teardown bookkeeping.

- fixes <link to issue>

## Types of changes

## Frontend coordination

The change is purely backend (CLI flag + a background task). The frontend never sees the new flag.

## Checklist

### DIFF
--- a/esphome_device_builder/__main__.py
+++ b/esphome_device_builder/__main__.py
@@ -201,6 +201,22 @@ def main() -> None:
         ),
     )
     parser.add_argument(
+        "--exit-on-parent-changed",
+        action=argparse.BooleanOptionalAction,
+        default=None,
+        help=(
+            "Exit cleanly when the original parent process disappears "
+            "(reparented to init / launchd on Unix; parent handle "
+            "signalled on Windows). Auto-enabled when the dashboard is "
+            "spawned by the ESPHome Builder desktop app — detected via "
+            "the ``io.esphome.builder`` marker in ``sys.executable``. "
+            "Pass ``--no-exit-on-parent-changed`` to disable the "
+            "auto-engage, or ``--exit-on-parent-changed`` to force-enable "
+            "(useful when launching under a custom supervisor that wants "
+            "the same behaviour)."
+        ),
+    )
+    parser.add_argument(
         "--trusted-domains",
         default=None,
         help=(
--- a/esphome_device_builder/controllers/config.py
+++ b/esphome_device_builder/controllers/config.py
@@ -119,10 +119,23 @@ class DashboardSettings:
     # operators who want to acknowledge the knob without
     # restricting hosts.
     trusted_domains: list[str] = field(default_factory=list)
+    # ``--exit-on-parent-changed`` tri-state.
+    #
+    # ``None`` = auto-engage on launcher signature (the ESPHome
+    # Builder desktop app's bundled Python — see
+    # :func:`helpers.parent_watchdog.should_engage`).
+    # ``True`` = force-on regardless of launcher.
+    # ``False`` = force-off regardless.
+    #
+    # Lives on the settings dataclass rather than as a flag the
+    # watchdog reads at import time so the test fixtures can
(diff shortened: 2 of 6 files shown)


bdraco:
Closing — the right fix landed on the desktop side at [esphome/esphome-desktop#77](https://github.com/esphome/esphome-desktop/pull/77).

That PR addresses the actual root causes:

1. **Tauri 2 doesn't fire `RunEvent::ExitRequested` for macOS Dock-Quit on tray-only apps** — that was the silent-orphan path. The desktop-side fix unifies all three quit paths (tray Quit / SIGINT-SIGTERM / Dock-Quit) through a single `app.exit(0)` + `RunEvent::Exit` handler, so the dashboard cleanup actually runs.
2. **The 5 s `SIGKILL` escalation was corrupting dashboard state mid-flush** — real shutdown can take up to ~30 s, and the timeout was force-killing during in-flight persists. The desktop-side fix removes the timeout escalation entirely (also disables `kill_on_drop(true)`, which was a hidden second SIGKILL source).

A watchdog on the dashboard side would have masked the orphan symptom but couldn't distinguish "graceful parent exit" from "force-quit / crash," and — more importantly — wouldn't have fixed the state corruption caused by the aggressive `SIGKILL` from the desktop side. The desktop fix removes the orphan and the corruption at the source.

Local branch + plan files are being cleaned up.

PR #616, opened 2026-05-11
Fix mDNS SRV target publishing the system FQDN instead of a .local hostname

bdraco:

The dashboard's mDNS SRV record target was sometimes published as the system's full FQDN (e.g. `mac.koston.org`) instead of the canonical mDNS form (`<label>.local`). The same machine could even show up under two different names on the same network depending on which code path queried it: `MacBook-Pro.local` from `hostname`, `mac.koston.org` from the advertiser's TXT record. The mDNS spec (RFC 6762) requires SRV targets on the `.local` domain to end in `.local.`; publishing a `.org` FQDN isn't a valid mDNS hostname and won't resolve for any peer browsing the `.local` domain.

## Root cause

`_default_hostname()` had a "trust any dotted form as already FQDN-ish" branch:

```python
if "." in raw:
    return raw          # ← publishes "mac.koston.org" verbatim
return f"{raw}.local"
```

`socket.gethostname()` returns whatever the kernel was given at boot. That's often a bare label (`mac`, `MacBook-Pro`) but on macOS or Linux with a configured search domain, or under some DHCP setups, it can come back as the full FQDN. The previous branch trusted it and published the FQDN into the SRV target.

## What this PR does

Two changes in `_default_hostname()`:

1. **Strip to the leftmost label and always append `.local`.** Same machine no longer reports two flavours of itself; both `mac.koston.org` and `MacBook-Pro.local` now normalise to a correct mDNS hostname regardless of OS, DHCP, or search-domain config.

2. **Lowercase the resulting label.** mDNS hostnames are case-insensitive per RFC 6762 §16, but downstream consumers (URLs, log strings, dedupe keys keyed on hostname) may compare case-sensitively. Bonjour, Avahi, and ESPHome itself all advertise hostnames lowercase; normalising at the publish site stops the same machine showing up as `Mac.local` in some flows and `mac.local` in others.

## What this PR explicitly does NOT change

The **service-instance name** (`_default_friendly_name`) stays case-preserving. That's user-visible display text in mDNS browsers / dashboard UIs; operators name their machines with intentional case ("MacBook-Pro", "Office Server"), and lowercasing the friendly name would degrade UX. Only the SRV hostname target normalises, because that's a wire-protocol identifier where case-sensitive consumers can drift. A new test (`test_default_friendly_name_preserves_case`) pins both contracts together so a future refactor that collapses the two helpers gets the right shape.

## Tests

| Test | Pins |
|---|---|
| `test_default_hostname_strips_fqdn_and_appends_local` | `mac.koston.org` → `mac.local` (the user's reported shape); `MacBook-Pro.local` → `macbook-pro.local`; `desktop.lan` → `desktop.local`. |
| `test_default_hostname_lowercases_label` | `MacBook-Pro` → `macbook-pro.local`; `Mac` → `mac.local`. |
| `test_default_hostname_handles_leading_dot` | Pathological `.koston.org` (no leftmost label) yields `""` rather than `.local`. |
| `test_default_friendly_name_preserves_case` | `MacBook-Pro.local` → `MacBook-Pro` (NOT lowercased); `Office-Server` → `Office-Server`. |

Existing `test_default_hostname_keeps_dotted` asserted the bug as a feature; replaced with the three above. Full suite: 3101 passed (+2 net new), 4 skipped, lint clean.

- User report: a single machine showing up as both `mac.koston.org` and `MacBook-Pro.local` on the dashboard, depending on which code path queried it.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue), `bugfix`
- [ ] New feature (non-breaking change which adds functionality), `new-feature`
- [ ] Enhancement to an existing feature, `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected), `breaking-change`
- [ ] Refactor (no behaviour change), `refactor`
- [ ] Documentation only, `docs`
- [ ] Maintenance / chore, `maintenance`
- [ ] CI / workflow change, `ci`
- [ ] Dependencies bump, `dependencies`

## Frontend coordination

- [x] No frontend change needed (the frontend reads whatever the advertise publishes; the fix is purely backend-side).

## Checklist

### DIFF
--- a/esphome_device_builder/helpers/dashboard_advertise.py
+++ b/esphome_device_builder/helpers/dashboard_advertise.py
@@ -199,18 +199,45 @@ def _default_hostname() -> str:
     """
     System mDNS hostname for the ``ServiceInfo.server`` SRV target.
 
-    Returns ``socket.gethostname()`` with ``.local`` appended when
-    the result has no dot. Doesn't use ``socket.getfqdn()``: on
-    macOS that resolver can return the reverse-DNS arpa form (e.g.
-    ``...ip6.arpa``) when reverse lookup fails, which is worse
-    than no hostname at all.
+    Always returns the leftmost label of ``socket.gethostname()``
+    with ``.local`` appended. The mDNS spec (RFC 6762) requires
+    SRV targets on the ``.local`` domain to end in ``.local.``;
+    a FQDN like ``mac.koston.org`` is not a valid mDNS hostname
+    and won't resolve on the ``.local`` domain peers are browsing.
+
+    ``socket.gethostname()`` returns whatever the kernel was
+    given at boot; that's often a bare label (``mac``,
+    ``MacBook-Pro``) but on macOS or Linux with a configured
+    domain, or under some DHCP setups, it can come back as the
+    full FQDN (``mac.koston.org``). The previous "if there's a
+    dot, trust it as-is" branch published that FQDN verbatim
+    into the SRV target, which broke discovery for any peer
+    expecting a ``.local`` hostname.
+
+    Doesn't use ``socket.getfqdn()``: on macOS that resolver can
+    return the reverse-DNS arpa form (e.g. ``...ip6.arpa``) when
+    reverse lookup fails, which is worse than no hostname at all.
     """
     raw = (socket.gethostname() or "").strip()
     if not raw:
         return ""
-    if "." in raw:
-        return raw
-    return f"{raw}.local"
+    # Strip any domain suffix the OS may have appended; the SRV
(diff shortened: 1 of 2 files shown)


bdraco:
Superseded by #617 (just opened). The narrow strip-and-lowercase approach left LAN-level hostname collisions unfixed; #617 mirrors HA's zeroconf advertise pattern with a per-install identifier in the SRV target ({short_hostname}-{short_dashboard_id}.local), which structurally eliminates the FQDN leak, the case-sensitivity drift, AND the collision class in one pass. See #617's PR body for the full rationale.

PR #729, opened 2026-05-13
Prioritize stale devices in bulk firmware jobs

Komzpa:

Queues bulk firmware jobs in stale-first order in the Device Builder backend:

- devices with an ESPHome version update available first, oldest deployed version first
- devices with pending YAML changes next
- already-current devices and unknown configuration names keep their relative order

This ports the legacy `update-all` ordering intent from esphome/esphome#15991 to the new Device Builder bulk firmware API.

- supersedes esphome/esphome#15991

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/firmware/bulk.py
+++ b/esphome_device_builder/controllers/firmware/bulk.py
@@ -3,7 +3,8 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
+import re
+from typing import TYPE_CHECKING, NamedTuple
 
 from ...helpers.api import CommandError
 from ...models import FirmwareJob, JobType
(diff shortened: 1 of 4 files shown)


esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -149,6 +150,20 @@ def _resolve_download_component(target_platform: str | None) -> str:
     return platform
 
 
+def _esphome_version_sort_key(version: str | None) -> tuple[int, int, int, int, str]:
+    """Return a sortable key for ESPHome versions."""
+    if not version:
+        return (-1, -1, -1, 0, "")
+
+    match = re.match(r"^(\d+)\.(\d+)\.(\d+)(?:[-.]?(.+))?$", version)
```
bdraco: time complexity of recompiling the regex is a bit of an issue here

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -149,6 +150,20 @@ def _resolve_download_component(target_platform: str | None) -> str:
...
+    if match is None:
+        return (-1, -1, -1, 0, version)
+
+    prerelease = match[4] or ""
+    release_rank = 0 if prerelease else 1
+    return (int(match[1]), int(match[2]), int(match[3]), release_rank, prerelease)
```
bdraco: 5 item return is really hard to follow code

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -659,6 +674,31 @@ async def install_bulk(
             jobs.append(job)
         return jobs
 
+    def _bulk_configuration_order(self, configurations: list[str]) -> list[str]:
+        """Return bulk firmware configs with stale devices first."""
+        devices_controller = getattr(self._db, "devices", None)
```
bdraco: should be a none check not a getattr

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -659,6 +674,31 @@ async def install_bulk(
...
+        if devices_controller is None:
+            return configurations
+
+        devices = {device.configuration: device for device in devices_controller.get_devices()}
+
```
bdraco: note that update available is defined as `bool(device.current_version and version != device.current_version)` -- which is from the legacy dashboard.. maybe we want to change that first as the sort likely doesn't do what you expect given that

esphome_device_builder/controllers/firmware/controller.py
bdraco: maybe 
```
 class _VersionKey(NamedTuple):
      major: int
      minor: int
      patch: int
      release_rank: int  # 0 = prerelease, 1 = stable; prerelease sorts first
      prerelease: str
```

But you could use the fast creation path like aiohttp uses for bulk keys

bdraco: TUPLE_NEW = tuple.__new__
avoids kwarg binding

```python
return TUPLE_NEW(
          _VersionKey,
          (int(match[1]), int(match[2]), int(match[3]), 0 if prerelease else 1, prerelease),
      )
```

bdraco (changes requested):
Thanks for the PR, please see review comments above

bdraco:
Please cover the missing lines https://github.com/esphome/device-builder/pull/729#issuecomment-4439565305

bdraco:
The code path you are targeting should be less of a moving target in the next hour or so. Still doing some cleanups

bdraco:
LGTM if copilot doesn't find any new issues

bdraco (approved):

bdraco:
Thanks @Komzpa

PR #886, opened 2026-05-15
Adding Default Components

### LINKED ISSUE #885: [feature-request] Allow adding default components like web_server
### What happened?

The starter kit doesn't have a way to default add a web server and the accessory power. We believe this should come in the default yaml as it is a bit confusing on when users need to add it and why. This will help with making it easier to start with the kit.

Discussion in starter kit slack

### Dashboard version

v0.0.0

### How did you install it?

Wheel from a GitHub release

### Where is it running?

Local — macOS

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

TrevorSchirmer:

** Claude generated code. Reviewed and tested boards with no default components and default components of web_server and the starter kit accessory power. **

Allows specifying default components for a device in the manifest. This helps a lot with the onboarding process for the starter kit

- fixes #885 

## Types of changes

## Frontend coordination

## Checklist


marcelveldt:
I'm fine with this change, looks good. @bdraco wdyt ?

bdraco:
I have 11 security PRs to review and then I'll take a look

esphome_device_builder/controllers/components.py
```diff
@@ -396,6 +397,47 @@ def get_featured_record(self, component_id: str) -> _FeaturedRecord | None:
...
+        catalog ``component_id`` lookup. The entry's own ``fields``
+        dict layers on top of any featured presets, with inline
+        values winning. Unknown references are skipped with a
+        warning — the manifest validator is the contract that
+        keeps these from reaching runtime.
+        """
+        # Lazy import sidesteps the components ↔ devices.helpers
+        # cycle (devices.helpers already imports _FeaturedRecord from
+        # this module under TYPE_CHECKING).
+        from .devices.helpers import _apply_featured_presets  # noqa: PLC0415
```
bdraco: Probably fine to import at top level

bdraco:
@bluetoothbot review

esphome_device_builder/controllers/components.py
bluetoothbot: Agreed — the lazy import isn't load-bearing here. `controllers/devices/helpers.py:34` references `_FeaturedRecord` from `..components` only under `TYPE_CHECKING`, and the chain `controllers/devices/__init__.py` → `controller.py` doesn't import `..components` at module level either, so a top-level `from .devices.helpers import _apply_featured_presets` in `components.py` won't trip a circular import. Safe to hoist; the `# noqa: PLC0415` and the cycle-explanation comment can both go away at the same time.

bdraco:
Bot found the same things I did so probably worth cleanup pass

esphome_device_builder/controllers/components.py
TrevorSchirmer: Moved to top level

TrevorSchirmer:
Thanks for the review and help Nick!

bdraco (approved):

jesserockz:
@marcelveldt @bdraco 

I feel like I should push back on this PR.

The whole point of the new UI (and the starter kit specifically) is to teach users how to use esphome. If we automatically provide things, then the user won't learn why they are there.

> Use this for board-specific config the device can't compile or work without

Webserver is none of those especially.

Just my 2c.

bdraco:
That's fair push back on this.  

I think we need to do two things

- write a project statement on who the target user is and review all PRs / changes through that lens
- testing with real users to see how high the bar is to show this is truly needed

TrevorSchirmer:
I understand the pushback. We thought the barrier of understanding the webserver or the accessory power was more than we wanted for the first step to use the kit. The user had no way to interact with the device without adding the webserver.

From the value prop the target user is someone who wants to learn electronics/programming or someone who wants to make their own smart home device. 

The entry point was set at no knowledge of ESPHome, home assistant, microcontrollers, or programming ( including yaml ). Since the webserver is needed to use the kit on step one we thought it was a good area to smooth the process. 

Users will still learn to use ESPHome through adding all the other components, once we can cover automations, and also with a "challenges" type section ( just need automations first ).

I do think it'd be good to have a "Explaining the base config" section on our wiki, we can get that added. We are also missing a section about "Explaining the ESPHome UI" as well I think. Thoughts on those two?

We'll also be getting feedback soon from the initial group any day.

jesserockz:
> The user had no way to interact with the device without adding the webserver

Then that should be a step in the getting started, and making sure the user understands it and why. I don't think we should automatically add it.

carlhye:
Hi @TrevorSchirmer,

I tend to agree with @jesserockz on this one, but I also see why the component is important.
We need to give the users a chance to learn why it's added, and not just add the webserver blindly.

What we'll suggest to do instead of adding it to the default components, is to add the webserver to the recommended components for the starter-kit, and also included it in the component bundles for the starter-kit.

That way, the user can either add it individually, or via a bundle, but in both cases they will clearly be able to see that a new component was added to the project, while also being able to read about the webserver component, what it does, and why it's important.

TrevorSchirmer:
Can we find a way to do some UX testing? Maybe utilizing the beta kits we've sent out. I think the experience/onboarding drops off quite a bit with the added complexity of adding these. We should have a wiki article done tomorrow for "Explaining the base config" which walks users through what is added automatically and should cover the teaching part.

marcelveldt:
I propose that we keep the default components feature as that may make sense at some point. Especially if a device requires some very specific default config (like platform config).

I think the accessory power might be a candidate for this, to add to the default components.
Its pretty much specific to the starterkit board. Still it makes sense that a user learns why we add it but that can also be explained in the tutorial/docs.

For the webserver component I'm more on the fence and I would propose to add that to the featured components.
So you can educate people to add it to their config if they want to use the device directly with a webinterface or skip it if they tend to only use it with HA/homey for example.

But, I'm also open to do a bit of A/B testing here.
Send one half of the test users the current flow where its automatically added and the other half an updated version where webserver is a featured component instead of default added.

PR #891, opened 2026-05-16
Scope automations/get_available to configured components

marcelveldt:

`automations/get_available` previously returned every trigger / action / condition in the catalog regardless of what was configured on the device, so the frontend's automation editor lists were flooded with entries that didn't apply (e.g. `switch.turn_on` showing on a device with no switches; `template.switch.publish` showing on a gpio-only switch). Filter the response down to the entries that actually match the device's configured components, with core control-flow (`if` / `delay` / `lambda` / combinators) and device-level triggers always included.

The matching uses the codebase-canonical `<domain>.<platform>` id form — the same shape the components catalog and `AvailableComponentInstance` already use — so the filter is one set-membership check against the qualified ids derived from the YAML (`switch`, `switch.template`, `datetime.date`, …).

- follow-up to #760 (automation editor backend)

## Types of changes

## Frontend coordination

Entry ids in the wire format are unchanged (e.g. `template.switch.publish` stays `template.switch.publish`), so existing YAML round-trips keep working. The `domain` / `applies_to` metadata fields now use `<domain>.<platform>` instead of `<platform>.<domain>`; any frontend code that groups or labels by those fields should adopt the new form to stay consistent with how the components catalog ids are shaped.

## Checklist


bdraco:
@bluetoothbot review

PR #903, opened 2026-05-19
firmware: collapse CR-overwritten progress lines at storage time

### LINKED ISSUE #898: Collapse CR-overwritten progress lines at job.output storage time
### Summary

`job.output` retains every CR-terminated progress chunk (`[1229/1233] Generating binary image...\r`, etc.) even though each one is immediately overwritten by the next. For ninja-driven ESP-IDF builds this is the majority of the line count — ~1200+ "Compiling …\r" / "Linking …\r" / "[N/total] …\r" lines that never appear on screen but are stored in full, replayed in full during historical follow, and discarded by the frontend on render.

Collapsing them at the storage seam (i.e. before they hit `job.output`) is a strict win:

- Less memory per running job.
- Far fewer events on historical replay (fixes the bulk of the perceived slowness in esphome/device-builder-frontend#348 from the producer side).
- The retained 2000-line tail covers more *useful* history — currently most of those 2000 slots are wasted on overwritten progress lines, so a real error 30 seconds before completion can fall outside the window during a chatty build. After collapse, the same 2000 slots hold real per-step output.

### Where it happens today

The streaming loop in `esphome_device_builder/controllers/firmware/controller.py:848` just appends every chunk:

```python
async for line in iter_lines_with_progress(proc.stdout):
    job.output.append(line)
    ...
    self._db.bus.fire(EventType.JOB_OUTPUT, {"job_id": job.job_id, "line": line})
```

`iter_lines_with_progress` is doing the right thing — it splits on `\n` *or* `\r` and forwards each chunk with its terminator so consumers can distinguish "new line" from "overwrite the previous one." But on the storage side we don't act on that signal; we just keep every chunk.

The frontend's `ansi-log.ts::_chunksToVisualLines` *does* honour the signal — it pops the previous visual line when the current chunk ends in `\r` and the next chunk isn't a bare `\n`. So everything is wired correctly for live rendering. It's just that the same collapse never gets folded back into the source-of-truth list.

### Proposed change

Push the same CR-overwrite rule into the storage step. Sketch:

```python
async for line in iter_lines_with_progress(proc.stdout):
    # Collapse CR-overwritten progress lines at storage time.
    # If the last stored chunk ended in '\r' and this one isn't
    # a bare '\n' (the trailing half of '\r\n'), the new chunk
    # is replacing the previous on-screen line — drop the old one.
    if job.output and job.output[-1].endswith("\r") and line != "\n":
        job.output[-1] = line
    else:
        job.output.append(line)

    if len(job.output) > _MAX_OUTPUT_LINES_INFLIGHT:
        _trim_job_output(job, keep=_INFLIGHT_TRIM_KEEP)
    self._db.bus.fire(EventType.JOB_OUTPUT, {"job_id": job.job_id, "line": line})
```

A few things to be careful about:

1. **Keep firing `JOB_OUTPUT` for every chunk, not just the appended ones.** Live followers still need to see each overwrite so their progress bars animate. The collapse is *storage-only*; the wire protocol stays the same. Historical replay (`_send_initial`) sends from `job.output`, which is already collapsed, so it benefits automatically.
2. **Mirror the frontend's bare-`\n` rule** — a `\n`-only chunk arriving after a `\r`-terminated chunk finalises the line in place rather than starting a new one. The frontend handles this; storage needs to handle it the same way to stay in sync.
3. **Don't collapse across the trim notice.** `_trim_job_output` inserts a `... N earlier line(s) elided` marker at index 0; the collapse check should compare against the actual last subprocess line, not the notice. (In practice the notice doesn't end in `\r`, so the existing rule won't falsely collapse it — but worth a comment so a future refactor doesn't regress it.)
4. **No behaviour change for `\n`-terminated lines.** Real log lines are unaffected; only chunks that the producer marked as "ephemeral with `\r`" get folded.

### Test impact

Existing tests in `tests/controllers/firmware/test_progress.py` exercise the live-streaming path with esptool-style `\r` progress lines (`"Writing at 0x10000 [bar]  84.8% NNN/NNN bytes..."`). Adding a unit test that:

- Feeds a sequence of `\r`-terminated chunks plus terminal newlines into the storage loop.
- Asserts that `job.output` ends up with one line per overwrite group (the last state of each).
- Asserts `JOB_OUTPUT` events still fire once per input chunk.

would lock in both behaviours.

### Why now

This is independent of the frontend rAF-batching fix proposed in esphome/device-builder-frontend#348 — both should land. Batching helps with bursty live streams from tools that *don't* use `\r` (pip install, package fetching). Storage-side CR collapse helps with the historical replay of long ninja builds, which is the worst case for issue #348 today.

The two together: historical replay of a 2000-line "stored" buffer becomes a replay of ~300–500 lines (real per-step output, no overwritten progress), and the frontend coalesces those into ~10 render frames instead of 2000. That should take the dialog-open delay from 2–4 s to well under 100 ms.

swoboda1337:

`job.output` retained every CR-terminated progress chunk
(`[1229/1233] Generating binary image...\r`, `Compiling …\r`,
`Linking …\r`) even though each was immediately overwritten on
screen by the next. For ninja-driven ESP-IDF builds that's the
majority of the line count — ~1200+ chunks that never appear on
screen but were stored in full and replayed in full during
historical follow.

Fold the CR-overwrite rule into `_ingest_output_line` (shared by
the local-subprocess streaming loop and the remote-source
listener): when the last stored chunk ends in `\r` and the new
chunk isn't a bare `\n`, replace the last entry instead of
appending. The collapse is storage-only — `JOB_OUTPUT` still
fires per chunk so live followers' progress indicators animate
unchanged; only `job.output` (the historical-replay source) is
collapsed.

Net effect on a long ESP-IDF build: the retained 2000-line tail
now holds real per-step output instead of overwritten progress
lines, so a real error 30 s before completion stays inside the
window. Independent of the frontend rAF-batching fix proposed in
device-builder-frontend#348 — both should land.

- fixes #898

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/firmware/helpers.py
+++ b/esphome_device_builder/controllers/firmware/helpers.py
@@ -316,7 +316,7 @@ def _fire_job_progress(job: FirmwareJob, bus: EventBus, progress: int) -> None:
 
 
 def _ingest_output_line(job: FirmwareJob, bus: EventBus, line: str) -> None:
-    """
+    r"""
     Append *line* to ``job.output`` and fire local follower events.
 
     Shared bookkeeping for "one line of build output arrived" —
@@ -326,7 +326,11 @@ def _ingest_output_line(job: FirmwareJob, bus: EventBus, line: str) -> None:
 
     Steps:
 
-    1. Buffer the line on ``job.output``.
+    1. Buffer the line on ``job.output``. CR-terminated chunks
+       overwrite the previous entry instead of accumulating, so
+       a ninja-driven ESP-IDF build's ~1200 "[N/total] …\r"
+       updates don't fill the retention tail with overwritten
+       progress lines (#898).
     2. Trim down to ``_INFLIGHT_TRIM_KEEP`` if the in-flight
        cap is hit, so a chatty build doesn't grow ``output``
        without bound between terminal-event trims.
@@ -346,7 +350,15 @@ def _ingest_output_line(job: FirmwareJob, bus: EventBus, line: str) -> None:
     ``failed`` status from the receiver instead of having to
     scrape stderr).
     """
-    job.output.append(line)
+    # Collapse CR-overwritten progress lines at storage time: a
+    # CR-terminated last entry followed by a non-bare-``\n`` chunk
+    # gets replaced rather than retained. ``JOB_OUTPUT`` still fires
+    # per chunk (live followers unchanged); only ``job.output``
+    # (historical replay) is collapsed.
+    if job.output and job.output[-1].endswith("\r") and line != "\n":
+        job.output[-1] = line
+    else:
+        job.output.append(line)
     if len(job.output) > _MAX_OUTPUT_LINES_INFLIGHT:
         _trim_job_output(job, keep=_INFLIGHT_TRIM_KEEP)
     out_payload: JobOutputData = {"job_id": job.job_id, "line": line}
--- a/tests/controllers/firmware/test_execute_job_e2e.py
+++ b/tests/controllers/firmware/test_execute_job_e2e.py
@@ -777,3 +777,73 @@ async def test_compile_repeats_error_pattern_short_circuits_check(
 
     assert job.status == JobStatus.FAILED
     assert job.exit_code == 0
+
+
+@pytest.mark.asyncio
+async def test_compile_cr_progress_lines_collapsed_in_storage(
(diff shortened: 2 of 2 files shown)


bdraco:
@Bluetoothbot review

bdraco (approved):

PR #909, opened 2026-05-20
Fix MQTT broker-resolve log spam for packages-based configs

### LINKED ISSUE #893: [Bug] Devices using MQTT spamming device builder log file
### What happened?

My devices that contain MQTT code are spamming the new ESPHOME Device builder log file. 
Only while dashboard is open.
Devices are running perfectly fine other wise and have been running this yaml for many months.
I have only just started trying out the new builder running on Mac and am very impressed at this development effort.
I tested some yaml code changes (eliminated secrets use) to see if it was my amateur code causing this.
The errors are not seen in older classic builder.

### Dashboard version

0.1.0b59

### How did you install it?

Other / not sure

### Where is it running?

Local — macOS

### Relevant logs / errors

```shell
Many many of these messages constantly from all yaml containing MQTT

[33m2026-05-17 09:47:43.795 WARNING (dashboard_1) [esphome_device_builder.controllers._device_mqtt_coordinator] Device pm37.yaml declares mqtt: but broker could not be resolved (missing secret or invalid config)[0m
```

### Anything else?

snippet of validated mqtt code from one of the devices
```
mqtt:
  broker: 192.168.1.203
  username: mquser
  password: <removed>
  discovery: false
  topic_prefix: ''
  id: mqtt_client
  on_connect:
    - then:
        - lambda: !lambda |-
            ESP_LOGI("mqtt", "Connected to MQTT broker");
  on_disconnect:
    - then:
        - lambda: !lambda |-
            ESP_LOGW("mqtt", "Disconnected from MQTT broker");
  on_message:
    - topic: pm37/totalenergy
      then:
        - lambda: !lambda |-
            static bool tot_e_restored = false;   // Declare and initialise only
            if (tot_e_restored)  {                // Total Energy Restored Flag
                return;           // Global last_tot_e has already been restored
            }
            // Only proceed to restore total energy value on 1st msg from MQTT
            auto mqtt_te = esphome::parse_number<float>(x);
            if (mqtt_te.has_value()) {
                ESP_LOGI("energy", "Restored total energy: %.8f", mqtt_te.value());
                id(last_tot_e) = mqtt_te.value(); // Set Global from MQTT Value
                tot_e_restored = true;            // Set Restored Indicator True
            } else {
                ESP_LOGW("energy", "Failed to parse energy value from MQTT");
            }
      qos: 0
    - topic: pm37/totalcost
      then:
        - lambda: !lambda |-
            static bool tot_c_restored = false;   // Declare and initialise only
            if (tot_c_restored)  {                // Total Cost   Restored Flag
                return;           // Global last_tot_c has already been restored
            }
            // Only proceed to restore total cost value on 1st msg from MQTT
            auto mqtt_tc = esphome::parse_number<float>(x);
            if (mqtt_tc.has_value()) {
                ESP_LOGI("cost", "Restored total cost: %.8f", mqtt_tc.value());
                id(last_tot_c) = mqtt_tc.value(); // Set Global from MQTT Value
                tot_c_restored = true;            // Set Restored Indicator True
            } else {
                ESP_LOGW("cost", "Failed to parse cost value from MQTT");
            }
      qos: 0
  enable_on_boot: true
  port: 1883
  clean_session: false
  discovery_retain: true
  discover_ip: true
  discovery_prefix: homeassistant
  discovery_unique_id_generator: legacy
  discovery_object_id_generator: none
  use_abbreviations: true
  keepalive: 15s
  reboot_timeout: 15min
  publish_nan_as_none: false
  wait_for_connection: false
  birth_message: {}
  will_message: {}
  shutdown_message: {}
  log_topic: {}
```

marcelveldt:

Devices that bring their `mqtt:` block in via `packages:` / `!include` —
a common shape for fleets that share a broker — caused
`_device_mqtt_coordinator` to log
`Device <foo>.yaml declares mqtt: but broker could not be resolved`
as a WARNING every 5 s while the dashboard is open.

The flag side (`Device.uses_mqtt`) is computed from the *resolved*
config, so it sees mqtt blocks pulled in by packages. The
broker-extraction side called `parse_mqtt_block` on the *raw* YAML,
which doesn't see packages — so it returned `None` and the warning
fired on every poll.

Changes:

- Add a slow-path fallback in `_resolve_broker` that re-uses
  `load_device_yaml` (the same resolver that drives `uses_mqtt`)
  when the raw-text parse misses. Cached per `(filename, mtime)`
  so steady-state polls don't pay the package-resolve cost.
- De-dupe the unresolved-broker WARNING per device: log it once,
  drop subsequent polls to DEBUG until the broker resolves again
  or the device drops `mqtt:`.

- fixes https://github.com/esphome/device-builder/issues/893

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/_device_mqtt_coordinator.py
+++ b/esphome_device_builder/controllers/_device_mqtt_coordinator.py
@@ -17,6 +17,7 @@ from typing import Any
 
 import yaml
 
+from ..helpers.device_yaml import load_device_yaml
 from ..helpers.yaml import FastestSafeLoader
 from ..models import Device
 from ._device_mqtt_monitor import (
@@ -52,6 +53,18 @@ class DeviceMqttCoordinator:
         self._on_state_change = on_state_change
         self._on_ip_change = on_ip_change
         self._monitors: dict[tuple[str, int], DeviceMqttMonitor] = {}
+        # Cached broker per device, keyed on the YAML's mtime so the
+        # expensive ``load_device_yaml`` fallback (resolves
+        # ``packages:`` / ``!include``) only runs again when the file
+        # actually changes. Unchanged YAMLs return their cached
+        # broker on every poll.
+        self._broker_cache: dict[str, tuple[float, MqttBrokerConfig | None]] = {}
+        # Device filenames we've already logged a broker-unresolvable
+        # WARNING for. Subsequent polls drop to DEBUG so a single
+        # mis-configured device doesn't spam the log every 5 s.
+        # Cleared when the broker resolves or the device stops
+        # declaring ``mqtt:``.
+        self._unresolved_logged: set[str] = set()
 
     @property
     def active_brokers(self) -> int:
@@ -98,23 +111,17 @@ class DeviceMqttCoordinator:
     def _collect_brokers(self) -> list[MqttBrokerConfig]:
         secrets_map = _load_secrets(self._config_dir)
         seen: dict[tuple[str, int], MqttBrokerConfig] = {}
+        seen_devices: set[str] = set()
         for device in self._get_devices():
             if not device.uses_mqtt:
                 continue
+            seen_devices.add(device.configuration)
             yaml_path = self._config_dir / device.configuration
-            try:
-                yaml_content = yaml_path.read_text(encoding="utf-8")
-            except OSError:
-                _LOGGER.debug("Could not read %s for MQTT broker config", device.configuration)
-                continue
-            broker = parse_mqtt_block(yaml_content, secrets_map)
+            broker = self._resolve_broker(yaml_path, secrets_map)
             if broker is None:
-                _LOGGER.warning(
-                    "Device %s declares mqtt: but broker could not be resolved "
-                    "(missing secret or invalid config)",
-                    device.configuration,
-                )
+                self._log_broker_unresolved(device.configuration)
                 continue
+            self._unresolved_logged.discard(device.configuration)
             existing = seen.get(broker.key)
             if existing is None:
                 seen[broker.key] = broker
@@ -126,7 +133,59 @@ class DeviceMqttCoordinator:
                     broker.host,
                     broker.port,
                 )
+        # Forget per-device tracking for devices that no longer
+        # declare ``mqtt:`` so the next time they do, the WARNING
+        # surfaces again instead of being silently suppressed.
+        self._unresolved_logged &= seen_devices
+        self._broker_cache = {k: v for k, v in self._broker_cache.items() if k in seen_devices}
         return list(seen.values())
+
+    def _resolve_broker(
+        self, yaml_path: Path, secrets_map: dict[str, Any]
+    ) -> MqttBrokerConfig | None:
+        """Resolve the broker for *yaml_path*, falling back to a full YAML resolve.
+
+        Fast path: raw-text parse of the device YAML — covers the
+        common case where ``mqtt:`` is at the top level of the file
+        itself. Slow path: ``load_device_yaml`` expands
+        ``packages:`` / ``!include`` so configs that pull the mqtt
+        block in from a shared package still resolve. The slow path
+        is cached per (file, mtime) so a steady-state poll doesn't
+        re-resolve packages every 5 s.
+        """
+        try:
+            yaml_content = yaml_path.read_text(encoding="utf-8")
+        except OSError:
+            _LOGGER.debug("Could not read %s for MQTT broker config", yaml_path.name)
+            return None
+        broker = parse_mqtt_block(yaml_content, secrets_map)
+        if broker is not None:
+            return broker
+        try:
+            mtime = yaml_path.stat().st_mtime
+        except OSError:
+            return None
+        cached = self._broker_cache.get(yaml_path.name)
+        if cached is not None and cached[0] == mtime:
+            return cached[1]
+        resolved = load_device_yaml(yaml_path)
+        broker = _extract_broker_from_config(resolved)
+        self._broker_cache[yaml_path.name] = (mtime, broker)
+        return broker
+
+    def _log_broker_unresolved(self, configuration: str) -> None:
+        if configuration in self._unresolved_logged:
+            _LOGGER.debug(
+                "Device %s declares mqtt: but broker still could not be resolved",
+                configuration,
+            )
+            return
+        _LOGGER.warning(
+            "Device %s declares mqtt: but broker could not be resolved "
+            "(missing secret or invalid config)",
+            configuration,
+        )
+        self._unresolved_logged.add(configuration)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


esphome_device_builder/controllers/_device_mqtt_coordinator.py
```diff
@@ -126,8 +133,60 @@ def _collect_brokers(self) -> list[MqttBrokerConfig]:
...
+            mtime = yaml_path.stat().st_mtime
+        except OSError:
+            return None
+        cached = self._broker_cache.get(yaml_path.name)
+        if cached is not None and cached[0] == mtime:
+            return cached[1]
+        resolved = load_device_yaml(yaml_path)
+        broker = _extract_broker_from_config(resolved)
+        self._broker_cache[yaml_path.name] = (mtime, broker)
+        return broker
```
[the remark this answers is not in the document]
marcelveldt: good catch — dropped negative caching so a fix recovers on the next poll, and folded `secrets.yaml`'s mtime into the cache key. didn't bother tracking per-`!include` mtimes since a packages edit usually warrants a restart anyway, and the no-negative-cache path handles recovery there too. follow-up commit pushed.

esphome_device_builder/controllers/_device_mqtt_coordinator.py
```diff
@@ -98,23 +115,17 @@ async def stop(self) -> None:
...
-            broker = parse_mqtt_block(yaml_content, secrets_map)
+            broker = self._resolve_broker(yaml_path, secrets_map)
             if broker is None:
-                _LOGGER.warning(
-                    "Device %s declares mqtt: but broker could not be resolved "
-                    "(missing secret or invalid config)",
-                    device.configuration,
-                )
+                self._log_broker_unresolved(device.configuration)
                 continue
```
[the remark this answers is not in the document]
marcelveldt: yep, fair — that was an unintended regression from the refactor. moved the read back into `_collect_brokers` so unreadable / deleted YAMLs hit the silent DEBUG path again. test now asserts no WARNING for that case.

esphome_device_builder/controllers/_device_mqtt_coordinator.py
```diff
@@ -52,6 +53,22 @@ def __init__(
...
         self._monitors: dict[tuple[str, int], DeviceMqttMonitor] = {}
+        # Cached broker per device, keyed on the YAML's mtime and
+        # ``secrets.yaml``'s mtime so the expensive
+        # ``load_device_yaml`` fallback (resolves ``packages:`` /
+        # ``!include`` / ``!secret``) only runs again when something
+        # the resolve depends on actually changes. Only successful
+        # resolves are cached — ``None`` results are re-tried every
+        # poll so a user edit to a ``!include``d / package file
+        # (whose mtime we don't track) recovers without a process
+        # restart.
```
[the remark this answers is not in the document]
marcelveldt: agreed — tightened the cache comment to be explicit: a package / !include edit on a previously-cached device won't invalidate until the device YAML or `secrets.yaml` mtime changes. didn't add a TTL since the no-negative-cache path already covers the realistic 'I fixed my config' recovery, and a stale broker on a successful resolve is rare enough that a manual touch / restart is fine.

esphome_device_builder/controllers/_device_mqtt_coordinator.py
```diff
@@ -52,6 +53,26 @@ def __init__(
...
+        # walking every transitively-included file on every poll
+        # isn't worth the I/O for the rare live-package-edit case;
+        # the user can touch the device YAML or restart the
+        # dashboard if they need an immediate refresh.
+        self._broker_cache: dict[str, tuple[tuple[float, float], MqttBrokerConfig]] = {}
+        # Device filenames we've already logged a broker-unresolvable
+        # WARNING for. Subsequent polls drop to DEBUG so a single
+        # mis-configured device doesn't spam the log every 5 s.
+        # Cleared when the broker resolves or the device stops
+        # declaring ``mqtt:``.
```
bdraco: comments are overly long

esphome_device_builder/controllers/_device_mqtt_coordinator.py
```diff
@@ -126,8 +151,62 @@ def _collect_brokers(self) -> list[MqttBrokerConfig]:
...
+        except OSError:
+            return None
+        cache_key = (yaml_mtime, _safe_mtime(self._config_dir / "secrets.yaml"))
+        cached = self._broker_cache.get(yaml_path.name)
+        if cached is not None and cached[0] == cache_key:
+            return cached[1]
+        resolved = load_device_yaml(yaml_path)
+        broker = _extract_broker_from_config(resolved)
+        if broker is not None:
+            self._broker_cache[yaml_path.name] = (cache_key, broker)
```
bdraco: I think there is some existing cache logic that is very similar in the code. Need to find the module. But likely could hoist it and make it more dry

bluetoothbot: The closest match in the tree is `esphome_device_builder/controllers/devices/_yaml_search_cache.py` (`YamlSearchCache`) — same `dict[str, tuple[mtime, value]]` shape, keyed on `configuration`, with a `prune(seen_configurations)` step that does what the inline `& seen_devices` does here. It also already handles the async-lock + stat-then-read race that this coordinator doesn't need (single-threaded poll) but would be free if hoisted.

The one mismatch is the two-mtime key: this cache invalidates on `(yaml_mtime, secrets_mtime)` rather than a single file mtime. Generalising `YamlSearchCache` to take a tuple-producing `key_fn: Callable[[Path], Hashable]` would cover both call sites without coupling the search cache to the secrets layout. Happy to do that as a follow-up refactor PR if it's worth pursuing — the device-yaml-build-info reader in `helpers/build_size.py` also looks like a candidate, which would make the hoist a real DRY win rather than a two-caller one.

esphome_device_builder/controllers/_device_mqtt_coordinator.py
marcelveldt: yeah, agreed — trimmed them across the board in a6ec0950.

esphome_device_builder/controllers/_device_mqtt_coordinator.py
marcelveldt: bot's right that `YamlSearchCache` is the closest shape but the fit isn't clean — single mtime vs my `(yaml_mtime, secrets_mtime)`, `list[str]` value vs `MqttBrokerConfig`, and the search cache does its own read+split while mine just memoises the resolve output. hoisting would need generic key_fn + value_fn params. happy to do that as a follow-up refactor PR if a third caller shows up (`build_size.py` looks like one), but doing it for just two call sites here would add abstraction faster than it removes duplication. keeping this PR scoped to the bugfix.

marcelveldt:
@bluetoothbot good catch on the stat-race — that was a twin of the read-side regression copilot flagged earlier, just on the other I/O call. fixed in c326940e by moving both the read and the stat into `_collect_brokers` so any OSError there hits the silent DEBUG path. test now also asserts no WARNING records. also did suggestion #1 (stat `secrets.yaml` once per poll instead of N times) and #3 (DRY'd the two broker extractors into a shared `_broker_from_block`). leaving suggestion #2 alone — the inline cache comment is now 4 lines, which is in line with the project's docstring/comment defaults.

PR #897, opened 2026-05-18
Sanitize peer-supplied mDNS labels in discover CLI

bluetoothbot:
# What does this implement/fix?

The dashboard's `_esphomebuilder._tcp.local.` browse CLI
(`esphome-device-builder-discover`) was decoding peer-controlled TXT
values (`server_version`, `esphome_version`, `pin_sha256`,
`remote_build_port`) and the mDNS instance name straight to stdout —
anything on the LAN advertising the service type could embed ANSI
escape sequences in those fields and reflow / clear the operator's
terminal on every browse event, or break the column-aligned table with
an oversized string.

**What**: Route every peer-supplied label through a new `_safe_label`
helper that strips non-printable characters (via `str.isprintable`)
and caps length to the matching `_FORMAT` column width. Switch the
bytes path to a single `decode("utf-8", errors="replace")` call so
malformed payloads can't raise out of the zeroconf callback.

**Why**: Same attack surface as the noise/hello/mDNS sanitization
work the aioesphomeapi project did in
[esphome/aioesphomeapi#1669](https://github.com/esphome/aioesphomeapi/pull/1669)
— mDNS is a peer-controlled channel and the discovery CLI prints its
output to a terminal that interprets ANSI escapes. A device on the
LAN broadcasting `\x1b[2J` in a TXT value could wipe a watcher's
terminal on every discovery event.

**How**: New `_MAX_*_LEN` constants per column, applied in `_decode`
(now takes a `limit` arg) and to `short_name` in
`_on_service_state_change`. No public API or wire-format change —
backwards-compatible for every existing caller of `_decode` because
the new `limit` parameter defaults to `_MAX_NAME_LEN`.

**Testing**: New tests demonstrate the attack vector against the
sanitized fields (ESC / CR / LF / NUL / tab stripping, length cap,
non-ASCII printable survival) and assert that hostile service names
and TXT values never reach stdout intact. Full suite green
(3482 passed, 4 skipped).

**Related issue or feature (if applicable):**

n/a — mirrors `esphome/aioesphomeapi#1669` defensively.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`

## Frontend coordination

- [x] No frontend change needed

## Checklist

- [x] The code change is tested and works locally.
- [x] Pre-commit hooks pass (`ruff`, `codespell`, yaml/json/python checks).
- [x] Tests have been added or updated under `tests/` where applicable.
[...]

PR #911, opened 2026-05-20
Sanitize peer-supplied mDNS labels in discover CLI

bdraco:

Ports esphome/aioesphomeapi#1689 to the dashboard's `esphome-device-builder-discover` CLI. Anything broadcasting `_esphomebuilder._tcp.local.` controls both the service instance name and every TXT value; the CLI was passing those bytes straight to stdout, so a hostile peer could embed ANSI escapes in `server_version` / `pin_sha256` / etc. and reflow the operator's terminal, or break the column-aligned table with an oversized value.

Routes every peer-supplied label through `_safe_label` (strip non-printables, length-cap), switches the bytes path to `decode("utf-8", "replace")` so a malformed payload no longer raises out of the zeroconf callback, and renames `_decode` to `_decode_mdns_label_or_unknown`. Per-column caps are derived from `_FORMAT` widths via regex with a module-load assertion pinning `_COLUMN_NAMES` against the derived widths, so a future `_FORMAT` edit cannot silently drift from the caps.

Replaces and closes https://github.com/esphome/device-builder/pull/897.

- mirrors esphome/aioesphomeapi#1689

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/discover.py
+++ b/esphome_device_builder/discover.py
@@ -22,6 +22,7 @@ import argparse
 import asyncio
 import contextlib
 import logging
+import re
 import sys
 
 from zeroconf import IPVersion, ServiceStateChange, Zeroconf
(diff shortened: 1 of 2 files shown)


bdraco:
@bluetoothbot review

PR #925, opened 2026-05-21
docs: README section on setting username and password

bdraco:

Discord question came in about where to set username and password on the new Device Builder Preview, especially the HA add-on case where 6052 stops being exposed and there's no field in the UI. Adding a short README section to answer it and to set expectations for each install path.

For the HA add-on the README now opens user-first: you reach the dashboard from the HA sidebar in your browser and from the Home Assistant Companion App, both come in over Ingress, and that means your Home Assistant login already protects it; there's no separate username / password to configure, and port 6052 stays unbound by default. The rationale for not exposing 6052 directly is the supervisor `/auth` endpoint, which has no rate limiting or lockout, so binding the port would turn the dashboard into an open brute-force target against every account on the instance. Users who actually need LAN reachability on 6052 are pointed at filing a feature request describing the use case, with the standalone PyPI install as the workaround today.

The standalone section leads with `ESPHOME_USERNAME` / `ESPHOME_PASSWORD` env vars and notes that `--password` is observable in process listings (`ps`, Task Manager, `/proc/<pid>/cmdline`). Verified the standalone behaviour locally with curl before writing.

- N/A

## Types of changes

- [x] Documentation only — `docs`

## Frontend coordination

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,78 @@ every few minutes.
 
 </details>
 
+## Username / password authentication
+
+The three install paths each handle network reach and the auth gate
+differently. The Home Assistant add-on runs ingress-only with no
+password under the new preview, so access goes through the HA
+sidebar and port `6052` stays unbound; ESPHome Desktop binds only
+to `127.0.0.1`, so there's nothing on the LAN to authenticate
+against in the first place; the standalone PyPI install binds
+`0.0.0.0:6052`, is LAN-reachable by default, and prints a startup
+banner warning that there is no auth gate until you configure one.
+
+Configuring a username and password is only wired up for the
+standalone install today. The HA add-on doesn't ship a knob for it
+(see below for the rationale, this is by design) and Desktop is
+localhost-only, so the knob doesn't add anything there. The backend
+itself accepts credentials anywhere it runs; it's the packaging
+around the HA add-on and Desktop that doesn't pass any in.
+
+### Home Assistant add-on
+
+When the **Use new Device Builder Preview** toggle is on, the add-on
+is reached through Home Assistant itself: you open it from the HA
+sidebar in your browser, and the Home Assistant Companion App opens
+it the same way. Both paths come in over Home Assistant Ingress, so
+the dashboard is already protected by your Home Assistant login;
+there's no separate username or password to configure for the
+add-on, and port `6052` stays unbound to keep the dashboard off the
+LAN by default.
+
+This is the supported way to use the add-on and we're not planning
+to add a "just expose the port directly" option for it. The classic
+dashboard let you opt in to binding port `6052` on the LAN and
+forwarded your Home Assistant username and password through the
+supervisor `/auth` endpoint to gate it; that endpoint has no rate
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,6 +109,76 @@ every few minutes.
...
+to Home Assistant credentials via the supervisor `/auth` POST flow
+isn't carried forward, and we don't plan to bring it back: forwarding
+HA credentials through an exposed port turns the dashboard into a
+brute-force surface against every account on the Home Assistant
+instance. The decision is intentional rather than missing
+functionality (see
+[device-builder issue #85](https://github.com/esphome/device-builder/issues/85)).
+
+If you have a use case that genuinely needs the dashboard reachable
+over the LAN (not through the HA sidebar), run the standalone PyPI
```
marcelveldt: Or file a feature request that raises our attention that this need exists (I would be curious about the usecase) - so if there is enough interest, we can decide to let you optionally enable port 6052 with mandatory auth. This can work independently of the Ingress path as we host that on a dedicated socket

README.md
```diff
@@ -109,6 +109,76 @@ every few minutes.
...
+against in the first place; the standalone PyPI install binds
+`0.0.0.0:6052`, is LAN-reachable by default, and prints a startup
+banner warning that there is no auth gate until you configure one.
+
+Configuring a username and password is only wired up for the
+standalone install today. The HA add-on doesn't ship a knob for it
+(see below for the rationale, this is by design) and Desktop is
+localhost-only, so the knob doesn't add anything there. The backend
+itself accepts credentials anywhere it runs; it's the packaging
+around the HA add-on and Desktop that doesn't pass any in.
```
marcelveldt, line 128: Can we rephrase this a bit more user-oriented and less technical implementation oriented ?
Also it would be good to note that the HA App uses Ingress and thus is protected by Home Assistant's authentication. Maybe that is not obvious for everyone

README.md
```diff
@@ -109,6 +109,76 @@ every few minutes.
...
+```
+
+Both values must be set together; setting only one fails the
+credential check at startup. The env var names are `ESPHOME_USERNAME`
+and `ESPHOME_PASSWORD`, not the legacy dashboard's bare `USERNAME` /
+`PASSWORD`, because the bare names collide with the OS-supplied
+`$USERNAME` on Windows and most login shells.
+
+The `--username` / `--password` CLI flags are still accepted for
+parity with the legacy dashboard's CLI, but please avoid
```
marcelveldt, line 175: What's holding us back from deprecating these ?
We can decide to throw a warning message for 6 months and then phase it out.

README.md
bdraco: fbdb5095

README.md
bdraco: AFAICT, nothing, can be done in a followup

README.md
bdraco: adjusted

marcelveldt (approved):

bdraco:
thanks

Issue #934, opened 2026-05-21
cut static memory footprint via lazy-loading catalogs + firmware logs


marcelveldt:
follow-up to #922 — an idle dashboard sits at ~700MB before a compile even runs. on the 2GB HA-addon VM in that issue there's no headroom for gcc.

three big eager loads contribute (sizes on disk):

- `definitions/components.json` — 22MB, ~14k mashumaro instances after parse
- `definitions/automations.json` — 15MB
- firmware-job persistence — restores up to 55 historical jobs × 2000 lines of build output (~15MB worst case) at startup

most of this is wasted RAM: list/search endpoints don't need `config_entries`, users mostly open one component card at a time, and old build-logs are rarely re-opened.

**approach** — slim-index + lazy-body, keeping orjson + mashumaro on the fast path:

- catalogs split build-time into an `index.json` (id / name / description / category / etc.) loaded eagerly, plus per-entry body files loaded on demand into a small LRU (~64 entries)
- firmware logs split into job-metadata (in the existing metadata blob) + a per-job sidecar log file under `ext_storage_path("dashboard-jobs")/<job_id>.log` — `FirmwareJob.output` populates lazily when the frontend opens a job detail view

ruled out:
- **sqlite** — read-only catalog regenerated by sync script, no relational queries we can't do in python, would lose the mashumaro fast path
- **`__slots__` + drop `DataClassORJSONMixin`** — would save ~4MB pre-lazy-body but only ~400KB once the LRU is bounded. not worth losing schema consistency, mashumaro's wire-codegen, and gaining a hand-rolled `to_dict()` per leaf model that has to stay in lockstep with field additions.

**estimated saving** — ~50-80MB resident off the 700MB baseline. headroom for #922's 2GB VM to stop sigkilling during compiles.

### checklist

- [x] **PR 1** — measurement scaffolding: tracemalloc-based memory benchmark in `tests/benchmarks/test_catalog_memory.py` covering all three catalog loaders + the firmware job restore path. regression gate, no production change.
- [ ] **PR 2** — `sys.intern()` on closed-vocabulary strings in `_load_component` / `_load_config_entry` (`category` / `platform_type` / `supported_platforms` / `references_component`). free win, no wire change. estimated 3-8MB.
[...]

PR #939, opened 2026-05-22
components: slim index + lazy-loaded bodies

marcelveldt:

PR 3 from #934 — split the component catalog so an idle dashboard stops carrying every component's `config_entries` tree in RAM. The slim index loads eagerly at startup; per-id bodies hydrate lazily on detail-view open through a bounded LRU.

**Related issue:**

- part of #934

## Changes

- `definitions/components.index.json` (slim, ~370KB) loads at startup; bodies live at `definitions/components/<id>.json` and hydrate via `ComponentCatalog.get_body` through a 128-entry LRU.
- `script/sync_components.py` emits the split layout — bodies land in a sibling `components.next/` dir first, then the whole dir + index swap into place so a Ctrl-C mid-write leaves the old catalog intact.
- `components/get_components` (list) returns the slim `ComponentCatalogIndexEntry` shape.
- `_FeaturedRecord` stores `underlying_id: str` instead of the body; the body is fetched at materialise time. `resolve_default_components` collects every underlying id up front so the whole board defaults batch loads in one executor hop.
- Scripts that need the full body (`validate_definitions`, `check_catalog`, `sync_esphome_devices`) join the index with the on-disk body files.

## Batch endpoint + loader rework

- New `components/get_component_bodies` WS command; the navigator now requests every body it needs in one round trip instead of one `get_component` call per id.
- Loader rewritten to the Home Assistant `translation.py` shape: single `asyncio.Lock` plus a batched `_load_bodies_from_disk` that reads every missing body sequentially in one `asyncio.to_thread` call. The interim per-id-`asyncio.Future` shape was replaced because it dispatched N small executor jobs per batch instead of one.
- `_load_bodies` returns the loaded entries directly so a batch larger than `_BODY_CACHE_MAXSIZE` doesn't silently lose its own early entries to mid-batch eviction; the cache is now a hot-read optimization, not the correctness path.
- The legacy `components/get_component` WS endpoint is gone; `get_component` survives as an internal wrapper so `add_component` and existing tests keep working without forking the lookup logic.
- New CodSpeed benchmark hydrates 100 commonly-used component bodies so regressions in orjson decode, dataclass build, or per-id recursion surface per release.

## Types of changes

## Frontend coordination

Companion frontend PR ships together; it adds microtask-batched `fetchComponent` driving the new endpoint and removes every direct `getComponent` caller. No backcompat window.

- [x] Companion frontend PR: esphome/device-builder-frontend#424

## Checklist


marcelveldt:
codeql is flagging the `_TIME_PERIOD_DEFAULT` regex on `script/sync_components.py:583` as a redos risk, but both alerts ([#12](https://github.com/esphome/device-builder/security/code-scanning/12), [#13](https://github.com/esphome/device-builder/security/code-scanning/13)) already exist on `main` — they're being re-attributed to this PR because the 900+ added body files push the diff over the "code changes were too large" threshold codeql warns about in the summary. not introduced here. happy to look at that regex separately if it's worth tightening.

bdraco:
@esphbot review

bdraco (approved):

Issue #941, opened 2026-05-22
[Bug] Light `effects` list rendered as a single text input instead of a list editor


bharvey88:
## Description

When editing a light component that has multiple entries under `effects:`, the device-builder UI collapses the entire list into a single text box. Each effect should appear as its own list item with its own editor row.

## Steps to reproduce

1. Open a device config that includes an `esp32_rmt_led_strip` (or any light platform) with multiple effects, e.g.:
   ```yaml
   light:
     - platform: esp32_rmt_led_strip
       name: RGB LEDs
       id: rgb_leds
       pin: 14
       num_leds: 10
       rgb_order: GRB
       chipset: WS2812
       rmt_symbols: 48
       effects:
         - addressable_rainbow:
         - addressable_color_wipe:
   ```
2. Open the light entity in the visual editor.
3. Scroll to the **Effects** field.

## Expected

The **Effects** field renders as a list with one row per effect (`addressable_rainbow`, `addressable_color_wipe`), each editable individually, matching how other list-of-objects fields are handled.

## Actual

The whole list is parsed into a single text input. The value appears as `- addressable_rainbow:- addressable_color_wipe:` jammed into one box, with no way to add, remove, or edit entries individually.

## Screenshot

<img width="1133" height="466" alt="Image" src="https://github.com/user-attachments/assets/6d3848d4-2ab3-4b7e-88c6-7f63b9f8578f" />

## Environment

- esphome/device-builder (web UI)
- Component: `light` (`esp32_rmt_led_strip`, but the issue is with the generic `effects` list rendering)

PR #996, opened 2026-05-28
Promote to_ntc_resistance / to_ntc_temperature calibration to multi_value

bdraco:

`to_ntc_resistance` and `to_ntc_temperature` accept their calibration data as a YAML list (`- 10.0kOhm -> 25°C`, `- 27.219kOhm -> 0°C`, …), but the upstream schema bundle drops the `is_list` marker on the field because the validator is a custom callable (`ntc_process_calibration`) rather than a structural `cv.*` combinator the bundle dumper can introspect. The catalog emits `calibration` as a plain required string, the visual editor renders a single text input, and saving clobbers the user's list. Add a per-filter override map in `_convert_registry_entry` that promotes the two known stuck fields to `multi_value: True`; the frontend's existing multi-value list editor then renders them as add/remove rows, same shape as `calibrate_polynomial.datapoints`.

- follow-up to #941

## Types of changes

## Frontend coordination

## Checklist


bdraco:
looks like these are the only 2 edge cases. the detection logic for them was 100+ lines so hard code is simpler

PR #1024, opened 2026-05-29
Lazy-load firmware-job output to per-job disk sidecars

bdraco:

Firmware-job build output was held in RAM for every retained job and reloaded eagerly at startup, costing roughly 15MB at idle. This moves terminal-job output to per-job sidecar logs under `CORE.data_dir/dashboard-jobs/<job_id>.log`; the metadata blob keeps only metadata, and the log is read lazily only when a user opens a job via `firmware/follow_job`. Active (queued or running) jobs keep their output in RAM and inline in the blob so a mid-build restart still recovers the pre-crash log; there are no active jobs at idle, so the resting footprint is metadata only.

A one-time migration moves any legacy inline output into sidecars on load, and stale sidecars are reaped on the next persist when their job is pruned or cleared. All disk I/O runs in the executor, so the event loop never blocks.

- part of #934

## Types of changes

## Frontend coordination

## Checklist


bdraco:
Probably spending too much time tuning a migration for a beta with stale data nobody is going to every look at again

PR #1046, opened 2026-05-29
Dispatch WS commands sequentially, bound streaming tasks

esphbot:

The `/ws` message loop spawned an unbounded, untracked task for **every** inbound frame (`create_task(client._handle_command(raw))`). Because each task awaits its handler, mutating commands on the *same* connection interleaved arbitrarily (two `set_preferences`, or `auth` racing a privileged command), and a client firing many frames fanned out unbounded server-side tasks and memory. The inbound command path had none of the backpressure the outbound stream queue (4000-slot) carefully enforces.

**Why:** highest-blast-radius robustness gap in the request path — both a correctness dimension (out-of-order execution of dependent commands) and a resource-exhaustion dimension (no cap / backpressure).

**How:**
- Await ordinary commands **inline** in the message loop → same-connection commands run in submission order and spawn no task.
- Spawn a tracked task **only** for long-running streaming handlers (`subscribe_events`, `firmware/follow_job(s)`, `devices/logs`, `devices/subscribe_reachability`), marked via `@api_command(..., streaming=True)` and collected into `DeviceBuilder.streaming_commands`.
- Cap concurrent streams per connection at `_MAX_CONCURRENT_STREAMS` (32); refuse the overflow with `RATE_LIMITED` instead of unbounded fan-out.

- Advisory GHSA-mg7m-j658-c6r9

## Types of changes

## Frontend coordination

## Testing

- New `tests/test_ws_dispatch_sequential.py`: ordinary commands run in submission order with no interleave (end-to-end through the aiohttp WS loop), streaming commands are spawned (not awaited inline) so a parked stream doesn't block later commands, ordinary commands spawn no task, and the stream cap returns `RATE_LIMITED`.
- Full suite: 3698 passed (1 pre-existing unrelated `real_compile` failure from a sandbox `xdg-user-dir` permission error). `ruff` clean.

## Checklist

---
### Quality Report

**Changes**: 2596 files changed, 2815 insertions(+), 11911 deletions(-)

**Code scan**: clean

**Tests**: failed ([Errno 13] Permission denied: 'pytest')

**Branch hygiene**: clean

*Generated by Kōan post-mission quality pipeline*


bdraco:
Yeah it's a real problem. Minor since auth is usually required but the bot is right that we shouldn't be making tasks for non streaming commands but it's not the best implementation because now everything has to wait on things that use the executor

bdraco:
@esphbot Rr

bdraco (commented):
Home Assistant has a better pattern for this that tracks the task and cancels them when it disconnects

bdraco:
@esphbot Rr

bdraco:
Not sure this is a real problem. Will think about it

Issue #1012, opened 2026-05-28
components: drive pin-field UX with hardware-capability requirements (pin_features) and pin_mode


bdraco:
## Background

While reviewing PR #1011 (mashumaro from_dict refactor), we noticed that two pieces of `ConfigEntry` metadata are wired up end-to-end but never carry any actual data, so the corresponding UX gates are no-ops today across all 317 PIN-type fields in the catalog:

| Field | What it would drive | Actual content today |
|---|---|---|
| `entry.pin_features: list[PinFeature]` | Filter the pin dropdown to board pins whose `features` list includes every required capability (`required.every(f => pin.features.includes(f))` at [`config-entry-pin-renderer.ts:125-128`](https://github.com/esphome/device-builder-frontend/blob/main/src/components/device/config-entry-pin-renderer.ts#L125-L128)) | Empty list for every component |
| `entry.pin_mode: PinMode | None` | Disable input-only pins for fields needing output (line 64-68 of the same file) | `None` for every component |

Background trail: PR #1011 surfaced this when filtering `_resolve_pin_features` in `script/sync_components.py` against the `PinFeature` enum dropped 100% of catalog `pin_features` entries (the schema was emitting GPIO mode strings like `input` / `output` / `pullup` that don't belong on a hardware-capability enum; the runtime was already silently dropping them via `_safe_enum` before they reached the wire). The data shape is now honest, but the metadata pipe runs dry.

The frontend filter wiring, the model shape, and the board manifests are all already in place — board manifests list real hardware features (`adc`, `dac`, `i2c_sda`, `spi_clk`, `uart_tx`, `pwm`, `touch`, `strapping`, `input_only`, …). The missing piece is just the per-field requirement.

## What this issue covers

1. **Hand-curated `_PIN_FEATURE_OVERRIDES` table in `script/sync_components.py`.** Map `(component_id, field_key) → list[PinFeature]` for high-value entries. ~20-30 rows covers most of the user-visible benefit:

   | Component / field | Required feature |
   |---|---|
   | `sensor.adc.pin` | `ADC` |
   | `dac.gpio.pin` | `DAC` |
   | `i2c.sda` / `i2c.scl` | `I2C_SDA` / `I2C_SCL` |
   | `spi.clk_pin` / `mosi_pin` / `miso_pin` / `cs_pin` | `SPI_CLK` / `SPI_MOSI` / `SPI_MISO` / `SPI_CS` |
[...]

PR #1051, opened 2026-05-30
Populate pin-field capability metadata (pin_features/pin_mode)

esphbot:

Populate `pin_mode` / `pin_features` on PIN-type catalog entries by introspecting ESPHome's **live pin validators** at sync time, and regenerate the catalog so the data ships now — restoring the frontend's pin-dropdown capability filter and input/output gate, which have been no-ops.

`ConfigEntry.pin_mode` / `pin_features` are wired end-to-end (model -> wire -> frontend) but have carried no data for any PIN field since #1011 made the schema-derived shape honest (the schema only emits GPIO mode strings like `input` / `output` / `pullup`, not hardware capabilities). This fills the pipe by deriving the metadata from the validators themselves rather than a hand-maintained table, so it tracks ESPHome across releases.

## Implementation

Derivation, not curation:

- `_pin_mode_from_closure()` reads the `default_mode` dict the `gpio_*_pin_schema` closure captures (`inspect.getclosurevars(validator).nonlocals["default_mode"]`) and maps `input`/`output` to a `PinMode`.
- `_pin_constraint_from_validator()` unwraps composed `vol.All` / `vol.Any` validators and returns a `_PinConstraint(mode, features)`. It **probes every node** for the `default_mode` closure (no brittle `__qualname__` match) and **merges modes order-independently** (`_merge_pin_modes`: any input + any output => INPUT_OUTPUT), so a composed validator is stable. Capabilities come from `_PIN_FEATURE_VALIDATORS` (a small `validator.__name__ -> PinFeature` map: `validate_adc_pin -> ADC`, `validate_touch_pad -> TOUCH`, `valid_dac_pin -> DAC`). Bus capabilities (i2c/spi/uart/pwm) are deliberately omitted: on the supported chips those route through the GPIO matrix to nearly any pin, so a capability requirement would wrongly narrow the dropdown. Fixed-function bus pins on esp8266 / libretiny are a separate follow-up (#1164).
- `_collect_pin_constraints()` walks the component's own live schema (`loader.get_platform` / `get_component`) via `_walk_schema_keys`, scoped to the exact manifest so the shared `gpio` stem doesn't stamp one direction onto both `binary_sensor.gpio` (input) and `output.gpio` (output). Logs a debug trace on manifest-load failure so resulting uncovered pins are traceable.
- `_apply_pin_constraints()` stamps the derived `pin_mode` / `pin_features` onto pin entries in `build_component_entry`.
- `_audit_catalog_for_pin_metadata()` runs in `main()` (gated on esphome introspection being available, so the lean/esphome-less install path isn't flooded) and logs a sync-time WARNING for any uncovered pin — telemetry for a new validator shape / renamed field rather than a silent no-op.

## Catalog regeneration

Unlike the usual "nightly owns the catalog regen" split, this PR **commits the regenerated catalog** (schema 2026.5.2) so the metadata ships immediately and the companion frontend affordance can be tested against real data. 148 component bodies gain `pin_mode` / `pin_features`; index / pin-registry / automation artefacts are unchanged. Spot-checks: `sensor.adc.pin` -> `[adc]`, `output.esp32_dac.pin` -> `output` + `[dac]`, `binary_sensor.esp32_touch.pin` -> `[touch]`, `output.gpio.pin` / `switch.gpio.pin` -> `output`, `binary_sensor.gpio.pin` -> `input`; i2c/spi/uart carry no `pin_features`. (The machine-dependent `compile_process_limit` range drift was reverted so the diff is pin-metadata-only.)

- Backend portion of https://github.com/esphome/device-builder/issues/1012
- Companion frontend grouping: esphome/device-builder-frontend#597
- Follow-up (fixed-function bus pins per platform): #1164

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#597 (softens the pin dropdown to group Recommended vs Other pins instead of hiding mismatches; consumes this `pin_features` / `pin_mode` data)

## Checklist

- [x] `components.index.json` / `definitions/components/*.json` were regenerated via `script/sync_components.py`, not hand-edited (the split-catalog roundtrip guard validates the bodies on commit).

## Testing

- `tests/test_sync_components_pin_features.py` — `_resolve_pin_features` (schema `modes` -> capability filter), `_pin_constraint_from_validator` direction derivation off the real `pins.gpio_input/output/pin_schema` validators, a non-pin validator returning empty, and `_collect_pin_constraints` against the installed esphome (gpio input vs output without cross-contamination; ADC / DAC / TOUCH). These walk real ESPHome validators, pinning the contract to ESPHome's pin schema.
- Regenerated catalog spot-checked (above) and validated by the commit-time roundtrip guard; `ruff check` + `ruff format --check` clean.
- Verified end-to-end in the browser via the companion frontend PR against this regenerated catalog.


script/sync_components.py
```diff
@@ -2568,6 +2570,53 @@ def _resolve_pin_features(raw: dict) -> list[str]:
...
+    ("binary_sensor.gpio", "pin"): PinMode.INPUT,
+    ("stepper.a4988", "step_pin"): PinMode.OUTPUT,
+    ("stepper.a4988", "dir_pin"): PinMode.OUTPUT,
+    ("stepper.a4988", "sleep_pin"): PinMode.OUTPUT,
+    ("stepper.uln2003", "pin_a"): PinMode.OUTPUT,
+    ("stepper.uln2003", "pin_b"): PinMode.OUTPUT,
+    ("stepper.uln2003", "pin_c"): PinMode.OUTPUT,
+    ("stepper.uln2003", "pin_d"): PinMode.OUTPUT,
+}
+
```
bdraco: This is unmaintainable

bdraco (changes requested):
current design is impossible to maintain

bdraco:
@esphbot rr

script/sync_components.py
esphbot: Agreed — the static table is the wrong long-term shape, and the failure mode is worse than "incomplete": there's no signal when it goes stale. It covers ~20 of dozens of pin-bearing components, and the only guard (`test_override_keys_exist_as_pin_entries_in_catalog`) checks that the listed keys exist, never that an uncovered pin field is intentional — so a renamed field or a new component just silently produces a no-op filter again, which is exactly the regression this PR is trying to fix.

The maintainable path: this metadata already lives authoritatively in ESPHome's pin validators (`pins.gpio_input_pin_schema` / `gpio_output_pin_schema` for direction, `adc.validate_adc_pin`, the i2c/uart pin validators for capability). `#1011` established the prebuilt schema bundle can't carry it, but `sync_components.py` already does narrow live `esphome` introspection for precisely the things the schema can't express (`multi_conf`, `platform_defaults`, `supported_platforms`). Deriving `pin_mode`/`pin_features` from the component's real pin schema in that same pass keeps the catalog correct across ESPHome releases instead of requiring a human to chase each one.

If the full introspection derivation is too big for one PR, I'd propose at minimum a sync-time coverage report (pin-typed entries with no derived metadata) so an uncovered component becomes a CI signal rather than something that goes unnoticed — and then layer the real derivation on top. Will rework along those lines rather than expanding the hand-curated table.

bdraco:
```
- exit code: 1

error[PLC0415]: `import` should be at the top-level of a file
   --> tests/test_sync_components_pin_features.py:121:5
    |
119 | def test_apply_only_targets_pin_typed_entries() -> None:
120 |     entries = [{"type": "string", "key": "pin"}]
121 |     from script.sync_components import _PinConstraint  # type: ignore[import-not-found]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
122 |
123 |     _apply_pin_constraints(entries, {("pin",): _PinConstraint(PinMode.OUTPUT, ())})
    |

error[PLC0415]: `import` should be at the top-level of a file
   --> tests/test_sync_components_pin_features.py:128:5
    |
127 | def test_apply_merges_without_duplicating_existing_features() -> None:
128 |     from script.sync_components import _PinConstraint  # type: ignore[import-not-found]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
129 |
130 |     entries = [_entry(pin_features=["adc"])]
    |

Found 2 errors.

```

bdraco:
@esphbot rr

script/sync_components.py
esphbot: This is addressed by the rework in `71a683e1`: the static `_PIN_FEATURE_OVERRIDES`/`_PIN_MODE_OVERRIDES` tables are gone entirely. Direction and capability are now derived from the component's live ESPHome schema — `pin_mode` off the `default_mode` dict the `gpio_*_pin_schema` closure captures (covering every gpio-typed field, not the ~20 that were curated), and `pin_features` from the fixed-silicon capability validators by name (`validate_adc_pin`→adc, `validate_touch_pad`→touch, `valid_dac_pin`→dac). Bus pins (i2c/spi/uart/pwm) are intentionally left featureless because they route through the GPIO matrix. The maintainability failure mode you flagged — a renamed field or new component silently producing a no-op filter — is now caught by `_audit_catalog_for_pin_metadata`, which emits a sync-time WARNING for any pin entry the walker derived nothing for. Only the test-file lint failure remains before this is mergeable.

bdraco (approved):

bdraco:
Pushed d42d03f7:

**Regenerated the catalog** (schema 2026.5.2) so the derived metadata ships in the committed bodies — 148 component bodies gain `pin_mode` / `pin_features`; index/registry/automation artefacts unchanged. Spot-checks: `sensor.adc.pin`→`[adc]`, `output.esp32_dac.pin`→`output`+`[dac]`, `binary_sensor.esp32_touch.pin`→`[touch]`, `output.gpio.pin`/`switch.gpio.pin`→`output`, `binary_sensor.gpio.pin`→`input`; i2c/spi/uart carry no `pin_features`. (Reverted the machine-dependent `compile_process_limit` range drift in `esphome.json` so the diff is pin-metadata-only.)

**Copilot review addressed:**
- Derive mode by probing every validator for the gpio-schema closure's `default_mode` (no brittle `__qualname__` match), and merge modes order-independently (input + output ⇒ INPUT_OUTPUT). Output is unchanged vs the previous logic (only `_schema_creator` closures capture `default_mode`), just robust to composed validators and traversal order.
- Gate `_audit_catalog_for_pin_metadata` on esphome introspection being available, so the lean/esphome-less install path isn't flooded with expected warnings.
- Log the manifest-load failure in `_collect_pin_constraints` so resulting uncovered pins are traceable.

The frontend soft-filter (issue #1012 item 3) is a companion device-builder-frontend PR.

Issue #1049, opened 2026-05-30
[Bug] Visual editor does not support key-value list in time ontime automation


masterofnonemx:
### What happened?

- Create/Edit device
- Add/Select time component (sntp, bm8563)
- Try to add an ontime automation on veditor/add ontime key in yeditor with multiple keys for time (hours, minutes, etc)
- Visual editor shows an error "invalid_args"

<img width="1534" height="843" alt="Image" src="https://github.com/user-attachments/assets/0eb619bd-cf17-42c7-8b97-26d8ac3638f7" />

Also, when an automation has an error like the above, all other automation appear as invalid.

### Dashboard version

Latest Beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

PR #1065, opened 2026-05-30
Isolate per-automation parse failures

### LINKED ISSUE #1050: [Bug] When an automation fails, others automations become invalid
### What happened?

Related to https://github.com/esphome/device-builder/issues/1049

The above issue generates an invalid automation, when this happens, other automations are marked as invalid

Image

### Dashboard version

Latest beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

When one automation in the device YAML had an unknown action or condition id, the whole `automations/parse` call failed; every automation editor then refused to load, so a single bad automation made all the others appear invalid. Each automation is now parsed independently; a bad entry comes back with an `error` and an empty tree while its siblings parse normally. A YAML that won't load at all still raises. The line based writer already splices by location, so editing a valid automation never rewrites the broken block; a regression test pins that. A companion frontend PR renders a flagged automation read only so its empty tree can't overwrite the real YAML.

- fixes https://github.com/esphome/device-builder/issues/1050

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#453

## Checklist


bdraco:
Thanks, both addressed.

Suggestion 2 (test robustness): agreed, the `on_time`/`hours` assertion pinned a decompose limitation rather than the isolation itself, and #1049 is a separate follow-up that should make `on_time` parse natively. Reworked the test to use a clearly-unknown action id (`not_a_real_action`) on a sibling handler, so it pins isolation independent of which real triggers decompose supports. The `on_time` shape still lives in the writer no-clobber test, where it survives regardless of decompose.

Suggestion 1 / silent-failure (catch scope): I went with your option (a), keep the narrow catch and pin the contract, rather than broadening to `except Exception`. Two reasons; first, every `raise` reachable from `build()` is `CommandError(INVALID_ARGS)` (the unknown action/condition paths); the only other raise, the `yaml.load` failure, is upstream in `parse_device_yaml` and intentionally still aborts. Second, the repo's ruff config enables `BLE` specifically to flag `except Exception:` so the catch scope stays deliberate, with no waiver for the controllers package; a blanket catch would cut against that policy and demote a genuine bug to a per-automation read-only state. The helper docstring now pins that decompose normalises per-automation faults to `CommandError`, so a future non-`CommandError` fault is a contract break to fix at the source, not something to swallow here.

PR #1071, opened 2026-05-30
Split device_yaml helper into a package

esphbot:

Splits the 1351 line `helpers/device_yaml.py` into a `helpers/device_yaml/` package; it was the largest module in the repo without an open split PR, well past the 800 line soft cap.

Cut along the three concerns the file already grouped with section banners: `_generation.py` synthesises new YAML, `_parsing.py` inspects raw and resolved config, `_loading.py` builds `Device` models from disk. `__init__.py` re-exports the public API plus the private symbols tests import, so every `helpers.device_yaml` import is unchanged. It's a behaviour free refactor; the test edits just repoint monkeypatch seams whose consumer moved to a submodule.

One of those seams (`test_detect_platform_from_yaml_swallows_parser_exceptions`) was patching the re-exported name, so after the split it no longer reached the binding `detect_platform_from_yaml` actually calls; the patch went silent and the `except` branch lost coverage. Repointed it at `device_yaml._parsing`, and added three tests for paths the suite never reached (`load_device_yaml` on a non-mapping top level, `extract_directly_referenced_integrations` skipping a non-string key, `_generation`'s import guard binding the upstream `has_native_wifi` when present). The package is back to 100%.

- N/A

## Types of changes

## Frontend coordination

## Checklist

## Testing

Full suite passes (4055 passed, 4 skipped, ignoring `tests/real_compile` which needs a live toolchain); `device_yaml` package coverage is 100%; ruff lint and format are clean; every new module is under the 800 line cap (max 520).


bdraco (approved):

bdraco:
accepted bad copilot suggestion needs fix up

PR #1072, opened 2026-05-30
Trim event-enum comments to bring models/common.py under the size cap

esphbot:

Trims the verbose comments and docstrings on the `EventType` and `StreamEvent` enums in `models/common.py` so the module drops back under the project's 800 line soft cap (869 to 590 lines).

I first tried extracting the two enums into a dedicated `models/events.py`, but that fragments a small, cohesive vocabulary across two files for little gain and adds an import indirection every reader pays. The comments were the actual bulk; most members carried a multi-line payload-shape note that restates the type annotation or duplicates the protocol docs. Trimming them in place keeps the enums where every caller already imports them from and gets the module comfortably under the cap.

No code or behaviour change: every enum member, its string value, and object identity are preserved (87 enum members before and after, ruff clean). Only comment and docstring lines are removed.

- N/A (opportunistic size split of one of the largest modules, per CLAUDE.md)

## Types of changes

## Frontend coordination

## Checklist


bdraco:
Part of the issue here is the very verbose docstrings and comments would be better to keep it in the same file and trim the docstrings as a first pass

bdraco:
@esphbot Rr

bdraco (changes requested):
PR title and body is stale

bdraco:
@esphbot rr

bdraco:
PR body must use PR template

bdraco (approved):

PR #1074, opened 2026-05-30
Clean the wizard and clone name/friendly_name to ESPHome's schema rules

### LINKED ISSUE #1070: [Bug] Device name not set as friendly_name
### What happened?

Follow-up for issue #926 

When I add a new device and use a descriptive name, any special characters are removed, and the name is used for both the name: and friendly_name: fields.

The fix in #926 seems to be incomplete. If I'm not mistaken, there's still a frontend-PR missing?

### Dashboard version

v0.1.0b74

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The new-device wizard kept the user's raw name as `friendly_name` and derived the hostname with an unbounded slugify, but ESPHome validates `friendly_name` with `string_no_slash` plus a 120 byte cap (and accepts control chars), and caps the hostname (`esphome.name`) at 31 chars; so a slash, an over-cap name, a pasted control character, or a long name produces YAML ESPHome rejects or that is unparsable (a literal NUL). This adds `clean_friendly_name` (swap the reserved `/` for `⁄` exactly as ESPHome's validator does, drop control / EOL chars, clamp to the byte cap on a character boundary) and `slugify_hostname` (clamp to the 31 char name cap, trimming a dash the cut may leave), and routes the wizard name through both before write. `_safe_yaml_scalar` still handles YAML quoting; a parametrized round-trip test pins that no character class lands invalid YAML or a schema-invalid name. A companion frontend PR stops the wizard from slugifying the name, so the backend receives the descriptive value to preserve. Clone's caller-supplied friendly_name is routed through `clean_friendly_name` too, so the clone path can't reintroduce the same symptom.

- fixes https://github.com/esphome/device-builder/issues/1070

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#461

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/devices/helpers.py
+++ b/esphome_device_builder/controllers/devices/helpers.py
@@ -16,6 +16,11 @@ try:
 except ImportError:  # pragma: no cover; covered by the import below
     from esphome.dashboard.util.text import friendly_name_slugify
 
+try:
+    from esphome.core.config import FRIENDLY_NAME_MAX_LEN
+except ImportError:  # pragma: no cover; older esphome without the constant
+    FRIENDLY_NAME_MAX_LEN = 120
+
 from esphome.helpers import rmtree, sort_ip_addresses
 from esphome.storage_json import StorageJSON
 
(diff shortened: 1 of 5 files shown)


bdraco:
Addressed in 1353fa9a. Routed clone's caller-supplied `new_friendly_name` through `clean_friendly_name` (the now-shared helper) before the `rewrite_name_or_substitution` write, so a clone with a slash / control char / over-cap friendly name gets the same cleaning as the create wizard and can't reintroduce #1070. Added a clone test (`Bath / Lamp\twith\x00ctl` → `Bath ⁄ Lamp with ctl`). Left clone's `new_name` (hostname) as-is: it's a distinct, frontend-validated input (the clone dialog collects the hostname directly via `validateDeviceName`), not a display name, so it's not the same generator gap.

PR #1085, opened 2026-05-30
Derive a repeatable flag for automation triggers

### LINKED ISSUE #1080: [Bug] on_time automation missing from add automation wizard when at least one is already added
### What happened?

Follow-up to #1049 

With the previous changes allowing multiple on_time events, when at least one on_time automation is already defined, the add automations wizard do not list Time -> On Time as an available automation.

Image

### Dashboard version

Lates Beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Adds a `repeatable` flag to the automation trigger catalog so the add-automation wizard knows which triggers ESPHome accepts as a list of stacked entries (multiple `time.on_time` schedules, `sensor.on_value_range` bounds, `mqtt.on_message` topics). Without it the wizard hid these after the first handler. This is the backend half; the frontend reads the flag.

The flag is derived, not a hardcoded id list. ESPHome's schema bundle has no list marker on triggers, and the `validate_automation` single flag is its default (102 triggers, far too broad). The signal that tracks intent is whether a trigger carries its own per-entry params, so `repeatable = has per-entry params`. I confirmed it by validating real two-entry configs through ESPHome's `validate_config`: params-carrying triggers validate as a list, paramless ones fire once. That gives 22 triggers and self-maintains as ESPHome adds more.

The flag lands on `AutomationTrigger` and the slim `AutomationTriggerIndex`; the catalog is regenerated via `script/sync_components.py`, every trigger gains the field, true for the 22, no other section changes.

Testing: a sync test pins the rule; a writer round-trip test proves the indexed-append is trigger-agnostic (`sensor.on_value_range` appends and validates), so the 21 newly-enabled triggers share `on_time`'s code path. Full suite green (4071 passed); semantic diff confirms only `repeatable` was added per trigger.

- fixes https://github.com/esphome/device-builder/issues/1080 (paired with the frontend change)
- Companion frontend PR: esphome/device-builder-frontend#468

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#468

## Checklist

- [x] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (the automations catalog is regenerated via `script/sync_components.py`).


bdraco:
merge bug is not a regression on this pr. its an existing bug so merging this and fixing in a new pr

PR #1126, opened 2026-06-01
Drop oversized LVGL update actions from the automation editor

### LINKED ISSUE #1123: [Bug] Using a LVGL action in an automation brings all lvgl properties in the editor
### What happened?

When using an LVGL action in an automation (e.g. lvgl.label.update), all LVGL properties are included in the editor's form. I guess it is related to the fact that LVGL sections are not supported in the editor

Image

### Dashboard version

Latest Beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Selecting an LVGL action like `lvgl.label.update` in the automation editor rendered a form with ~164 fields, because ESPHome's LVGL `*.update` schema flattens the entire widget style schema (`bg_color`, `arc_*`, `border_*`, `shadow_*`, ...). Every other action stays under 30 config entries, so there is a clean gap. This flags any action above a config-entry threshold (80) as `form_editable=False` in the generated slim index; the catalog drops those from the action picker and from the known-id set, so an automation that uses one falls through the existing unknown-action path to read-only raw-YAML editing. Small LVGL actions (`lvgl.pause`, `lvgl.page.show`, the `canvas.draw_*` verbs) stay editable. No parser changes and no hand-maintained domain list; the signal is the schema size itself, so it adapts as ESPHome's schema evolves.

- fixes https://github.com/esphome/device-builder/issues/1123

## Types of changes

## Frontend coordination

## Checklist


bdraco:
Thanks, all three addressed in 44f28db:

1. Index/wire bloat: the slim index now carries `form_editable` only on the 31 non-editable actions; everything else defaults to True on load, so the index grows by 31 keys instead of one per action. Kept the runtime filter design (single central gate, bodies stay coherent on disk rather than orphaned) rather than dropping the entries entirely; the wire still carries a constant `form_editable: true` on returned actions, which is harmless and the frontend ignores it.
2. Observability: added an `_LOGGER.info` that prints the id and config-entry count whenever an action is flagged, so a future action that trips the threshold surfaces in the sync log; the comment also notes the signal is top-level breadth, not total form size.
3. Vacuity: the catalog test now asserts the excluded ids ARE present in the unfiltered slim list and gated out of `_ACTION_IDS`, so it fails loudly if the index isn't regenerated or an id is renamed. That assertion also fixed a blockbuster `os.stat` failure, the previous `action_by_id` call did a blocking body read inside the async test; the in-memory `_ACTION_IDS` check proves the same thing with no I/O.

bdraco:
Addressed in 8920973:

1. Switched the lean-index cleanup to `slim_dict.pop("form_editable", None)` so it can't KeyError if the key is ever absent.
2. Trimmed the `form_editable` field comment, the `_editable_actions()` docstring, and the `_MAX_FORM_CONFIG_ENTRIES` comment to contract-only per CLAUDE.md; the why now lives here in the PR.
3. Cut the new test docstrings to one-liners that name what each pins, no production-story retelling.

No behaviour change, so the index didn't need regenerating; 287 automation tests still green.

PR #1131, opened 2026-06-02
Pipeline the firmware queue: concurrent compile + upload lanes

bdraco:

Pipelines the firmware job queue so a network-bound upload runs concurrently with a CPU-bound compile, per esphome org discussion #3702. Before this, the queue ran one job at a time and `install` was a fused `esphome run`, so a slow upload (for example ~9 minutes over Thread, CPU idle) blocked the next device's compile.

The queue now has two concurrent single-worker lanes: a compile lane (CPU; compile, clean, reset, rename) and an upload lane (network). `firmware/install` is a two-job chain, a COMPILE job plus a dependent local UPLOAD job (`FirmwareJob.depends_on`); the upload is held off its lane until the compile succeeds, then runs on the upload lane while the compile lane is free for the next device. `install_bulk` queues the same chain per device, so a multi-device install pipelines (device B's compile overlaps device A's upload) instead of running serially. A cancelled or failed compile cascades to cancel the held upload, so a cancelled build never flashes; remote installs use the same chain since a remote compile already materialises artifacts locally, and the receiver advertises compile-lane idleness so an uploading receiver still accepts delegated compiles.

Lane state lives on `FirmwareState` as the shared source of truth (`lane_for` / `dependency_satisfied` / `place_on_lane`); the install chain is committed atomically (one rename-lock check, then place + announce both jobs synchronously, one persist) so a concurrent rename can't strand a half-queued pair. `_run_queue` spawns the two lane consumers via the eager-task helper and drains both on any exit. Because the lanes run concurrently, build-tree-mutating ops cancel conflicting work: a per-device clean cancels that device's in-flight chain, and `reset_build_env` (clean-all) cancels every active job on both lanes before wiping. The cancel cascade is persisted so a restart doesn't re-cancel dependents.

Tests cover the chain, the cancel cascade (unit + real-runner e2e), lane concurrency e2e, cross-lane verify-chip, persistence re-queue by lane, supersede/clean/reset cancellation, and the compile-idle scheduler signal; firmware suite green with 100% patch coverage on the changed controllers. CLAUDE.md, docs/ARCHITECTURE.md, and docs/API.md are updated.

- fixes <link to issue>

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#551

Frontend impact: `firmware/install` now produces **two** job rows (a compile and a dependent upload), and two firmware jobs can be RUNNING at once. The events are all standard `JOB_*`; the frontend must render concurrent RUNNING rows and the held/queued upload (new `FirmwareJob.depends_on` field). The compile→upload progress reset at the seam already matches the existing remote-install behavior.

## Checklist


bdraco:
### Scope note: parallel use of multiple remote build servers is a follow-up (not in this PR)

This PR pipelines **compile vs. upload** — each lane is a single worker, run concurrently — so a slow upload no longer blocks the next compile.

It does **not** make multiple remote build servers run at once. The compile lane processes one job at a time, and a REMOTE compile occupies that lane for its whole duration (it awaits the receiver's compile + artifact download). So with N paired servers and N queued installs, the servers are still used **one at a time**, not in parallel — the scheduler picks one receiver per job (oldest-idle, then oldest-otherwise); it doesn't fan out concurrently.

**Follow-up:** allow the compile lane to dispatch multiple **remote** compiles in parallel (bounded by the number of idle receivers), since remote compiles are I/O-bound waits rather than local CPU — e.g. a small remote-dispatch pool on the compile lane, or a dedicated remote lane with N concurrent in-flight dispatches keyed on idle receivers. This composes cleanly on top of the lane model here and is intentionally out of scope for this PR (which keeps each lane single-worker).

bdraco:
Thanks for the review. Addressed in 9374161e and 7f5f6918:

- Repo-root scratch note: removed `compile_upload_queue_split_plan.md`; the design lives in the PR description and the working copy is kept out of the tree, so nothing rots after merge.
- Back-compat shims: removed the `state.queue` / `current_job` / `current_process` proxies and migrated the test call sites to `state.compile_lane.*`. Production was already fully lane-aware, so the shims only survived for tests; with them gone, `state.current_job` can no longer read as a half-truth that hides an in-flight upload.
- Single-element task list: reverted `_runner_tasks` back to the original `_runner_task`, since `_run_queue` is one wrapper task that spawns the two lane consumers.

Folded in a small DRY pass while here: hoisted the duplicated `wire_real_queue` / `run_until_terminal` / `upload_of` e2e helpers into the firmware conftest, and switched the bulk-install rename-lock test to the `with_queue` fixture so its `put_nowait` is sync, clearing an AsyncMock warning. Full firmware suite is green (463 passed, 2 skipped).

bdraco:
Addressed in 5189ba13:

- **Stale comment**: fixed the held-branch comment in `factories.enqueue` to reference the public `lifecycle.release_dependents` (no leading underscore).
- **MEDIUM 1 (`_run_queue` only handling CancelledError)**: switched the `except asyncio.CancelledError` to a `try/finally`, so any lane consumer error (not just shutdown cancellation) now cancels and drains the sibling lane before the original error propagates, rather than orphaning it mid-flight. Added a regression test (`test_run_queue_cancels_sibling_lane_when_one_raises`).

On **MEDIUM 2 (`dependency_satisfied` could strand a dependent)**: this is already covered by the callers, so I left it as-is. `dependency_satisfied` is deliberately narrow, but no caller can strand a dependent:

- At enqueue, the chain creates the COMPILE before the dependent UPLOAD, so the prerequisite is always present.
- At restart, `persistence._restore_to_lane` explicitly disambiguates: a prerequisite that is already COMPLETED enqueues the dependent directly; one still active holds it; one missing or terminal-but-not-completed cancels it with an error. So a pruned or already-finished prerequisite can't leave a dependent QUEUED forever.
- At runtime, `release_dependents` fires on the prerequisite's terminal transition (every terminal path goes through `finalize_terminal`) synchronously, before any `prune_history` can remove the prerequisite, and active jobs are never pruned.

Full firmware suite green (468 passed, 2 skipped).

bdraco:
Addressed in 9685a50b:

- **`supersede_active_jobs` cascade double-cancel (real bug)**: confirmed. Re-installing a config supersedes the prior COMPILE + UPLOAD chain; cancelling the prior compile cascades to cancel its held upload, so when the loop reaches that already-terminal upload, `cancel` raises `CommandError("Cannot cancel a cancelled job")`. `CommandError` is a bare `Exception`, not a `ValueError`/`RuntimeError`, so the existing `suppress` missed it and the second install would have raised. Added `CommandError` to the `suppress` (the job is already in the state we wanted) and pinned it with `test_reinstalling_supersedes_prior_chain_without_raising`.
- **`JobType.INSTALL` comment**: corrected. There is no on-load migration; persisted INSTALL jobs are re-queued and run as a fused `esphome run` via the runner + CLI. The comment now says INSTALL is retained so older persisted jobs still deserialise and run, and that new installs enqueue a COMPILE + dependent UPLOAD instead.
- **`capture_enqueue_order` docstring**: updated to `put_nowait` (the enqueue path is synchronous now).

On the related "collapse async to sync" cleanup: I audited the firmware package for `async def`s that lost their last awaitable to the `put_nowait` switch — there are none. `enqueue` / `enqueue_install_chain` / `supersede_active_jobs` / `load_jobs` / `cancel` all still `await` (persist, supersede, executor, terminate); `release_dependents` and `persistence._restore_to_lane` were already sync. The only awaitless `async def`s are the pre-existing read-only API getters `jobs.get_jobs` / `get_job`, which predate this PR and are unrelated to the queue change.

Full firmware suite green (469 passed, 2 skipped).

bdraco:
Addressed in 4a63c7a9:

- **Orphan on rename-lock (🟢)**: `enqueue_install_chain` now wraps the two enqueues in a `try/except CommandError` that pops both just-created jobs from `state.jobs` and re-raises. So a rename-locked install (or any rejected enqueue) leaves nothing stranded for restart to re-run. Pinned by `test_install_chain_rolls_back_when_rename_locked`.
- **Broadened suppression (🟡)**: narrowed it — supersede now catches `CommandError` and re-raises unless the code is `INVALID_ARGS` or `NOT_FOUND` (the benign already-terminal / already-gone cases the cascade produces). An unexpected typed cancel failure now surfaces. Pinned by `test_supersede_reraises_unexpected_cancel_error`; the benign path stays covered by the reinstall cascade test.

Full firmware suite green (472 passed, 2 skipped).

bdraco:
Addressed in d9c169f0 + b0d30785:

- **Rollback didn't re-persist / left a ghost JOB_QUEUED (🟡)**: rather than re-persist in the `except`, `enqueue_install_chain` now commits the pair **atomically** — one rename-lock check, then place + fire both JOB_QUEUED synchronously, then a single supersede + single persist, with no `await` between the check and the commit. That closes the window entirely (a rename can't slip in mid-chain), so there's no half-persisted upload and no orphan-on-restart, and the rollback only covers the synchronous pre-commit lock check.

While here I also fixed a bug the reviews didn't surface but you'd hit immediately: **`install_bulk` still created a fused `INSTALL` job per device**, which runs on the compile lane — so a multi-device install ran serially and never used the upload lane (no pipelining). It now routes each device through `enqueue_install_chain`, so device B's compile pipelines against device A's upload.

And per a follow-up request, factored the duplication: added `FirmwareState.place_on_lane` (callers no longer reach through `lane_for(job).queue.put_nowait`) and a shared `_place_and_announce` helper used by both `enqueue` and the install chain; `lifecycle.release_dependents` / `persistence._restore_to_lane` route through `place_on_lane` too.

The over-broad `suppress` from the silent-failure note was already narrowed in 4a63c7a9 (re-raise unless `INVALID_ARGS` / `NOT_FOUND`), now with both the swallow and re-raise paths tested. Full firmware suite green (473 passed); `factories.py` / `bulk.py` at 100% patch coverage.

bdraco:
On the two Copilot discussions:

- **r3338567551 (QUEUED-cancel cascade not persisted)** — fixed in a63661bc. The QUEUED-cancel path persisted the cancelled job *before* `release_dependents` cascaded to its held upload, so the upload's CANCELLED status never reached disk and a restart re-cancelled it every boot. `release_dependents` now returns whether it acted, and the cancel path re-persists (+ prune) when it did. Pinned by a test asserting two persists on a chain cancel.

- **r3338567568 (reset can run concurrently with an upload)** — partly addressed in 359cd1ae: `reset_build_env` now calls `cancel_all_active_jobs` to cancel every in-flight job on both lanes before queueing the wipe, and per-device clean supersedes the device's chain. That closes the *existing-job* race. The residual is a *newly-started* upload during a clean/reset (a standalone `firmware/upload` lands on the upload lane and could run while the wipe touches the same artifacts) — a deliberate, user-triggered window rather than an automatic collision. Making clean/reset fully exclusive (hold/block new jobs for the affected scope until they finish) is the robust fix; I've left it as a follow-up rather than add an exclusive-gate mechanism in this PR. Flagging for a call on whether to do it here.

bdraco:
Thanks — status on each:

- **🟡 #1 (QUEUED-cancel cascade not persisted)** — fixed in a63661bc (landed after this review ran). `release_dependents` now returns whether it acted, and the QUEUED-cancel branch re-prunes + re-persists when it did, so the cascade-cancelled upload's CANCELLED status reaches disk. Pinned by a test asserting two persists on a chain cancel.

- **🟡 #2 (clean now supersedes instead of rejecting — undocumented reversal)** — intentional (requested: a clean is the user asking for a fresh build, and with the two lanes a held/running upload could otherwise read artifacts the clean wipes). Now documented in 3f42cc00: docs/API.md's `firmware/clean` row, the "one active job per device" section, and the reset section all call out the cancel semantics; the PR description was also updated. No required frontend change — the cancellations fire standard `JOB_CANCELLED` (handled generically; there was no clean-reject error path to remove). A pre-cancel confirmation is a possible optional UX add.

- **🟡 MEDIUM (`dependency_satisfied` collapsed boolean)** — runtime can't strand a held upload: `release_dependents` fires synchronously on the compile's terminal (placing or cancelling the upload) *before* any `prune_history`, so a succeeded compile is never pruned out from under a still-held upload. On restart, `_restore_to_lane` deterministically cancels a dependent whose prereq is missing/terminal (with an error) rather than holding it — and 3f42cc00 adds a log there so a pruned-but-succeeded prereq is diagnosable, not silent.

- **🟢 #3 (dead aggregate `queue_status_snapshot`)** — retained intentionally as the public aggregate accessor (idle only when both lanes are). The offloader path correctly uses the compile-lane variant (`compile_queue_status`) by design; the aggregate is still exercised by the slot-release-before-fire ordering tests as a sync probe. Happy to drop it if you'd prefer no unused accessor.

Firmware suite green (470 passed, 2 skipped).

bdraco:
Addressed:

- **🟡 #1 (reset/clean wipe races a newly-started upload)** — implemented the exclusive gate in 86e3402e. The upload lane now holds an upload (`runner._await_build_gate`) while a `RESET_BUILD_ENV` (any) or a same-configuration `CLEAN` is active — `FirmwareState.upload_blocked` is the predicate, woken by `build_gate` which is set on every job terminal. Compile-lane jobs serialize behind the clean/reset already, so only uploads need the gate. Unit + run_lane integration tests added.
- **Copilot (`_cancel_active_jobs` swallows RuntimeError)** — fixed in 35a8d3c3. The *global* cancel (`cancel_all_active_jobs`, for reset) now re-raises `RuntimeError`/`ValueError` (a RUNNING job it couldn't terminate would race the wipe), while per-configuration supersede keeps swallowing it. Test added.
- **Copilot (stale "one job runs at a time" doc)** — fixed in 35a8d3c3; the queue note now describes the two concurrent lanes.
- **🟡 MEDIUM (`dependency_satisfied`)** — runtime can't strand (release-before-prune); restart cancels a missing-prereq dependent deterministically, now with a log.

Also hoisted `FirmwareState.active_jobs()` and reused it across `upload_blocked`, the supersede/cancel-all loop, and the rename-lock check. Firmware suite green (473 passed).

PR #1155, opened 2026-06-02
Resolve esphome.area from included/templated configs

### LINKED ISSUE #1153: [Bug] Area not displayed in new ESPHome Device Builder Preview (Templates)
### What happened?

Similar to https://github.com/esphome/device-builder/issues/966

When using devices that reference a template, the area does not populate.

Example

device.yaml
```
substitutions:
  devicename: "powermon-dryer"
  friendly_devicename: "Power Monitor - Dryer"
  device_description: "Sonoff S31"
  device_area: "Laundry Room"
  device_area_id: "laundry_room"

<<: !include templates/template-s31-outlet.yaml
```

template-s31-outlet.yaml
```
esphome:
  name: "${devicename}"
  comment: "${device_description}"
  friendly_name: "${friendly_devicename}"

esp8266:
  board: esp12e #esp01_1m
  early_pin_init: false #This prevents the on-off-on of the relay upon firmware upload
  area:
    id: ${device_area_id}
    name: "${device_area}"
...
```

### Dashboard version

ESPHome Device Builder v0.1.0b85 ESPHome 2026.5.2

### How did you install it?

### Where is it running?

Docker container

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

When a device's `esphome:` block is pulled in from another file — via a
merge-key include (`<<: !include templates/foo.yaml`), a plain
`!include`, or `packages:` — and the area is written with flat
substitutions, the resolved area never reached the device card; the drawer
and area column stayed blank.

The meta reader (`parse_esphome_meta`) only scans the **raw text of the
main file**, so when the `esphome:` block lives in an included file the
main file has no `esphome:` block and the reader returns
`(None, None, None, None)`. Meanwhile `load_device_from_storage` already
loads a fully resolved config (`load_device_yaml`, used for
api/integration detection) that expands merge-keys, includes, and
packages — but we only ever read its `substitutions:`, never the merged
`esphome:` block.

This reads `name` / `friendly_name` / `comment` / `area` off the resolved
config's `esphome:` block as a source slotted **between** the raw-text
parse (immediate-edit / broken-draft resilience) and StorageJSON
(post-compile authority), resolving `$var` / `${var}` against the merged
`substitutions:` map with the existing resolver. Inline configs are
unaffected (raw text and resolved config agree). The dotted
`${device.area}` form (key-value substitutions) stays unresolved and
continues to defer to StorageJSON (esphome persists it there as of
esphome/esphome#16710), so there is no behaviour change for that case.

New `extract_esphome_meta_from_config` handles the string and `{id, name}`
mapping forms of `area`; `_effective_meta` is generalised to `_pick_meta`
to take the resolved-config value as a second preferred source.

Both meta readers now return an `EsphomeMeta` `NamedTuple`
(`name` / `friendly_name` / `comment` / `area`) instead of a bare
4-tuple, so fields read by name at the call sites. It stays positionally
compatible, so existing unpacking and tuple-equality call sites are
unchanged.

- fixes https://github.com/esphome/device-builder/issues/1153

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/helpers/device_yaml/__init__.py
+++ b/esphome_device_builder/helpers/device_yaml/__init__.py
@@ -31,12 +31,14 @@ from ._loading import (
     load_device_yaml,
 )
 from ._parsing import (
+    EsphomeMeta,
     _parse_inline_value,
     config_has_top_level_block,
     configuration_stem,
     detect_platform_from_yaml,
     device_uses_mqtt,
     extract_directly_referenced_integrations,
+    extract_esphome_meta_from_config,
     get_api_encryption_block,
     get_api_encryption_key,
     parse_esphome_meta,
@@ -46,6 +48,7 @@ from ._parsing import (
 )
 
 __all__ = [
+    "EsphomeMeta",
     "StorageJSON",
     "_esphome_has_native_wifi",
     "_fallback_has_native_wifi",
@@ -59,6 +62,7 @@ __all__ = [
     "detect_platform_from_yaml",
     "device_uses_mqtt",
     "extract_directly_referenced_integrations",
+    "extract_esphome_meta_from_config",
     "generate_device_yaml",
     "generate_minimal_stub_yaml",
     "get_api_encryption_block",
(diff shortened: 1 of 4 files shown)


bdraco:
Addressed in 67bf9952. `_pick_meta`'s final fallback now returns only the raw-text token (or `None`), never the config-derived value — so a never-compiled device whose `esphome:` block lives in an `!include` / `packages:` file with a dotted `${device.area}` no longer leaks the literal token into the area column; it stays blank, matching pre-PR behaviour. Added `test_load_device_unresolvable_area_in_include_is_blank_not_token` pinning `area == ""` for that case.

PR #1156, opened 2026-06-02
Fix peer-link pairing failure from clobbered dashboard_id

### LINKED ISSUE #1154: [Bug] Peer-link pairing still fails on 0.1.0b86: offloader identity at peer_link differs from the approved record (no_approved_peer / pin_mismatch). Follow-up to #1045
## Summary
Follow-up to #1045 (same two machines). With both ends upgraded to `0.1.0b86`, **Send builds** pairing still never completes. The b86 release now surfaces rejection reasons in the logs (thanks, this is what #1045 asked for), and they show the actual fault clearly:

**The peer identity observed during the approval handshake does not match the one presented at the final `peer_link` step**, so the approved-peer lookup never matches and the link is rejected (`no_approved_peer` on one side, `pin_mismatch` on the other). The GUI shows matching fingerprints and a successful pairing, but the underlying link silently fails.

Importantly, a clean reset does not fix it: I rotated the peer-link identity, then deleted the key and paired again from scratch, and the exact same failure reproduced. So this is not just leftover stale state, the pairing/link logic itself is broken.

## Environment

| | Windows host | HAOS host |
|---|---|---|
| OS | Windows 11 Pro | HAOS bare-metal, N100 mini PC |
| IP | 192.168.1.132 | 192.168.1.252 |
| device-builder | `0.1.0b86` | `0.1.0b86` |
| ESPHome CLI | `2026.5.2` | (not captured) |
| peer-link key file | `C:\Users\bharv\esphome\.device-builder-peer-link-key.bin` | (esphome config dir) |

Both hosts on the same `/24`. WS connections and handshakes succeed in both directions, so this is not mDNS/firewall.

## Steps to reproduce
1. Autodiscovery did not surface the build server this time, so from the HAOS dashboard I used **Send builds** and entered the Windows PC IP (`192.168.1.132`) manually.
2. Accept the pairing request.
3. The GUI showed the **same fingerprint icons on both sides** and reported pairing **successful**.
4. The link never actually establishes; the backend rejects it.

## Expected
Pairing completes and the build server becomes an eligible remote build target.

## Actual
GUI reports success (matching fingerprints), but the peer-link is rejected on every attempt. The observed offloader identity at the `peer_link` step differs from the approved record.

### HAOS (receiving from Windows 192.168.1.132)
Approval phase observes Windows pin `9913453f` / dashboard_id `6rHrXQu`, but the actual `peer_link` arrives with pin `9c3d7d59` (Windows' pre-rotation identity) and a different dashboard_id `L1Uxzi5`:
```
preview      from 192.168.1.132 -> ok       observed_offloader_pin=9913453f… dashboard_id=
pair_request from 192.168.1.132 -> pending  observed_offloader_pin=9913453f… dashboard_id=6rHrXQu_oUeKuT1TPf1zYLyLKaQ_PSXi
pair_status  from 192.168.1.132 -> approved observed_offloader_pin=9913453f… dashboard_id=6rHrXQu_oUeKuT1TPf1zYLyLKaQ_PSXi
peer_link    from 192.168.1.132 -> REJECTED reason=no_approved_peer observed_offloader_pin=9c3d7d59… dashboard_id=L1Uxzi5QBiGi2HU7aYgv648lwd4stHxK
```

### Windows (receiving from HAOS 192.168.1.252)
Same dashboard_id `kbTV8ZyO` throughout, but the offloader pin changes between approval (`63938e88`, stored) and link (`4f582c18`, observed):
```
preview      from 192.168.1.252 -> ok       observed_offloader_pin=63938e88…
pair_request from 192.168.1.252 -> pending  observed_offloader_pin=63938e88… dashboard_id=kbTV8ZyO…
pair_status  from 192.168.1.252 -> approved observed_offloader_pin=63938e88… dashboard_id=kbTV8ZyO…
peer_link    from 192.168.1.252 -> REJECTED reason=pin_mismatch stored_offloader_pin=63938e88… observed_offloader_pin=4f582c18…
  WARNING: "offloader identity rotated or a stranger is claiming this dashboard_id"
...
peer-link client connected to 192.168.1.252:6055 -> rejected by receiver (reason=no_approved_peer); orphaning until the operator re-pairs or unpairs
```

### Windows peer-link identity rotation (operator-initiated)
```
16:21:49 Loaded peer-link identity … pub=5dd7d219… pin=9c3d7d59…
16:23:44 Rotated peer-link identity at C:\Users\bharv\esphome\.device-builder-peer-link-key.bin
16:23:44 Loaded peer-link identity … pub=5b804c46… pin=9913453f…
```

## Analysis / suspected cause
1. **Approval and link-establishment observe different identities for the same peer.** On both sides the `peer_link` step carries a different offloader pin than the record approved seconds earlier, so the approved-peer lookup always misses. This is the direct cause of `no_approved_peer` / `pin_mismatch`.
2. **After an identity rotation on Windows** (`9c3d7d59 → 9913453f`), HAOS approves the new pin but the inbound `peer_link` still carries the old pin, suggesting a cached/stale identity is used for the outbound link rather than the current one.
3. **A clean reset does not help.** Rotating, then deleting the key file and re-pairing from scratch on b86 reproduces the identical failure. So the cause is in the pairing/link logic, not merely leftover state from a prior version (though pin `4f582c18` does also appear in #1045's older `b65`/`b73` logs).

## What's new vs #1045
- Both ends now on matched `0.1.0b86`; behavior unchanged from #1045 but now diagnosable thanks to the surfaced rejection reasons.
- Confirmed a clean rotate + delete + re-pair still fails, ruling out stale state as the sole cause.
- Autodiscovery did not work this round (had to enter the Windows IP manually under **Send builds**).
- GUI shows matching fingerprints and "success" while the backend rejects the link, so the operator gets a false-positive paired state.

Related: #1045, #607.

bdraco:

Fixes the peer-link pairing failure in #1154: "Send builds" reports success in the GUI, but the long-lived peer-link is rejected on every attempt with `no_approved_peer` (one direction) and `pin_mismatch` (the other), and it survives restarts, key rotation, and re-pairing.

Two distinct root causes.

### Primary (deterministic, survives restarts): `_remote_build` co-tenancy clobbered `dashboard_id`

`dashboard_id` is a permanent correlation token. It was stored at `_remote_build.dashboard_id`, but the receiver-settings writer co-owned that block and replaced it wholesale (`data["_remote_build"] = settings.to_dict()`, which carries only `enabled` + `cleanup_ttl_seconds`). Any settings write (toggling enabled, changing the TTL, or enabling the feature) erased `dashboard_id`; the next read minted a fresh one. The offloader then paired under one `dashboard_id` and opened `peer_link` under another, so the receiver's `approved_peers` lookup missed (`no_approved_peer`). It recurred after every restart because re-enabling the feature re-erased the id, which is why the reporter's resets never helped.

Rather than just guard the write, the fix removes the **two-owners-of-one-block** design: `dashboard_id` now owns a dedicated `_dashboard_identity` metadata block, so each top-level key has a single writer and a settings write structurally can't reach the identity. `_get_or_create_dashboard_id` migrates a legacy `_remote_build.dashboard_id` in place (same value, no re-pair). The settings writers also merge rather than replace, as a safety net during the migration window.

This also fixes a latent bug: minting the identity used to create `_remote_build`, which tripped `has_remote_build_settings_persisted` and made a fresh HA-addon install bind the receiver on the next boot without the operator opting in. Minting the identity no longer trips that gate (the three existing HA-addon gate tests still pass; a regression test pins the new behavior).

### Secondary (single-run): peer-link key snapshot drift

The offloader snapshotted its peer-link private key once at `start()`; the long-lived `PeerLinkClient` kept presenting it after a key rotation while the pairing handshake used the live key (`pin_mismatch`). `_load_offloader_identities_async` now refreshes that snapshot on every load, and the offloader subscribes to `REMOTE_BUILD_IDENTITY_ROTATED` to re-snapshot and respawn its APPROVED peer-link clients so a rotation reaches the wire.

- fixes https://github.com/esphome/device-builder/issues/1154

## Types of changes

## Frontend coordination

## Checklist

- [x] Tests have been added or updated under `tests/`.
- [x] Architecture-level changes are reflected in docs (n/a).

## Tests

- E2E (`test_remote_build_peer_link.py`): a receiver settings write keeps an existing pairing accepted; a key rotation breaks the link until re-pair, then recovers, over the real Noise wire.
- Migration / gate (`test_dashboard_identity.py`): legacy `_remote_build.dashboard_id` migrates to its own block with the same value; a legacy-only block is removed; minting the identity does not trip the HA-addon persisted gate; both real settings writers preserve foreign `_remote_build` keys.
- Full e2e (`tests/e2e/test_identity_rotation.py`): the real two-instance harness pairs, rotates the offloader key + fires the rotation event, watches the session drop, then re-pairs and confirms recovery pinned to the rotated key.
- Unit (`test_remote_build_controller.py`): the identity-load path refreshes the snapshot; a rotation event re-snapshots and respawns only APPROVED peer-link clients.

New regression tests fail on the pre-fix tree.

### DIFF
--- a/esphome_device_builder/controllers/config/remote_build_settings.py
+++ b/esphome_device_builder/controllers/config/remote_build_settings.py
@@ -124,11 +124,9 @@ def _merge_settings_into_block(data: dict[str, Any], settings: RemoteBuildSettin
     """
     Overlay *settings* onto the existing ``_remote_build`` block in place.
 
-    The block is co-owned: ``dashboard_id`` lives here too
-    (:func:`helpers.dashboard_identity._get_or_create_dashboard_id`).
-    Replacing the whole block with ``settings.to_dict()`` would
-    drop ``dashboard_id`` on every write, re-minting the
-    dashboard's stable identity and breaking paired peers.
+    Merges rather than replaces so a co-resident key the settings
+    model doesn't own survives the write — notably a legacy
+    ``dashboard_id`` awaiting migration out of this block (#1154).
     """
     rb = data.get(_REMOTE_BUILD_KEY)
     if not isinstance(rb, dict):
--- a/esphome_device_builder/controllers/config/remote_build_settings.py
+++ b/esphome_device_builder/controllers/config/remote_build_settings.py
@@ -121,13 +121,7 @@ def has_remote_build_settings_persisted(config_dir: Path) -> bool:
 
 
 def _merge_settings_into_block(data: dict[str, Any], settings: RemoteBuildSettings) -> None:
-    """
-    Overlay *settings* onto the existing ``_remote_build`` block in place.
-
-    Merges rather than replaces so a co-resident key the settings
-    model doesn't own survives the write — notably a legacy
-    ``dashboard_id`` awaiting migration out of this block (#1154).
-    """
+    """Merge *settings* into the existing ``_remote_build`` block, preserving foreign keys."""
     rb = data.get(_REMOTE_BUILD_KEY)
     if not isinstance(rb, dict):
         rb = {}
--- a/esphome_device_builder/controllers/config/remote_build_settings.py
+++ b/esphome_device_builder/controllers/config/remote_build_settings.py
@@ -120,10 +120,27 @@ def has_remote_build_settings_persisted(config_dir: Path) -> bool:
     return isinstance(_load_metadata(config_dir).get(_REMOTE_BUILD_KEY), dict)
 
 
+def _merge_settings_into_block(data: dict[str, Any], settings: RemoteBuildSettings) -> None:
+    """
+    Overlay *settings* onto the existing ``_remote_build`` block in place.
+
+    The block is co-owned: ``dashboard_id`` lives here too
+    (:func:`helpers.dashboard_identity._get_or_create_dashboard_id`).
+    Replacing the whole block with ``settings.to_dict()`` would
+    drop ``dashboard_id`` on every write, re-minting the
+    dashboard's stable identity and breaking paired peers.
+    """
+    rb = data.get(_REMOTE_BUILD_KEY)
+    if not isinstance(rb, dict):
+        rb = {}
+        data[_REMOTE_BUILD_KEY] = rb
+    rb.update(settings.to_dict())
+
+
 def save_remote_build_settings(config_dir: Path, settings: RemoteBuildSettings) -> None:
     """Persist the receiver-side remote-build settings."""
     with metadata_transaction(config_dir) as data:
-        data[_REMOTE_BUILD_KEY] = settings.to_dict()
+        _merge_settings_into_block(data, settings)
 
 
 @contextmanager
@@ -150,4 +167,4 @@ def remote_build_settings_transaction(
     with metadata_transaction(config_dir) as data:
         settings = _settings_from_raw(data.get(_REMOTE_BUILD_KEY, {}))
         yield settings
-        data[_REMOTE_BUILD_KEY] = settings.to_dict()
+        _merge_settings_into_block(data, settings)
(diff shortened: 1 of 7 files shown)


bdraco:
I ripped out all the id rotation and race guards. The bot reviews are trying to protect against a scenario that never happens. No sane case rotates the id of the dashboard back to back to back.  Its a once in a lifetime or never event for this production

PR #1163, opened 2026-06-03
Emit per-registry pin mode flags for the visual editor

bdraco:

The long-form pin **Mode** checkboxes in the visual editor (`input` / `output` / `pullup` / `pulldown` / `open_drain`) are offered for every pin, but each external pin provider in ESPHome's `PIN_SCHEMA_REGISTRY` allows only a subset: an I2C expander like `pca9554` permits just `input` / `output`, a shift register `sn74hc595` just `output`. The catalog attaches the generic esp32 flag set to every pin field, so the editor lets users build configs ESPHome rejects (`[pullup] is an invalid option for [mode]`).

The constraint is **provider-keyed, not field-keyed** (the `switch.gpio` `pin` field is generic and can hold any provider's value, selected by the provider key in the value), so it can't be baked per-field. This emits it as a global map the frontend consults at render time.

## Implementation

- Introspect ESPHome's **live** `PIN_SCHEMA_REGISTRY` at sync time (`_build_pin_registry_modes`), reusing `_get_esphome_loader()`. Each provider registers on import, so every component is imported to populate the registry, then `_pin_registry_allowed_modes` walks each registered pin schema to its `mode` mapping (`_pin_schema_mode_mapping` unwraps `vol.All` / `Schema`) and reads the allowed flag keys.
- **Native target platforms are excluded.** They allow every checkbox flag, so scoping a native pin is a no-op; only external providers restrict. Natives register under both the `Platform` enum (esp32) and bare platform-name strings (rp2040, bk72xx), so the filter drops any key in `{p.value for p in Platform}` rather than relying on the key type. The result is a clean `{provider_key: [allowed_modes]}` map (19 providers: pca9554, pcf8574, mcp23xxx, sx1509, sn74hc165/595, tca9555, …).
- Emit `definitions/pin_registry_modes.index.json` (`_emit_pin_registry_modes_index`, atomic temp-then-replace, mirroring `sync_boards`'s featured-index emit). Skipped on a `--limit-component` debug run whose partial registry would clobber the committed artefact.
- `definitions/__init__.py` gains `load_pin_registry_modes_index()` (missing / malformed -> `{}`, mirroring the board / featured loaders).
- `ComponentCatalog.load()` caches it; `components/get_pin_registry_modes` exposes it. Documented in `docs/API.md`. Added to the wheel `package-data`.

The committed `pin_registry_modes.index.json` is generated by the script; the nightly sync owns regeneration going forward (the catalog index/bodies are intentionally not regenerated in this PR).

- Backend portion of https://github.com/esphome/device-builder-frontend/issues/584 (frontend scoping is a companion `device-builder-frontend` PR)

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: scopes the Mode checkboxes against `components/get_pin_registry_modes` (links this PR)

## Checklist

## Testing

- `tests/test_sync_components_pin_registry_modes.py` — `_pin_schema_mode_mapping` unwrap/scalar cases, `_pin_registry_allowed_modes` flag-key extraction, `_build_pin_registry_modes` against the installed esphome (pca9554 / pcf8574 input-output, mcp23xxx +pullup, sn74hc595 output-only, sn74hc165 input-only), and exclusion of native target platforms (enum- and string-keyed).
- `tests/test_definitions_loader.py` — `load_pin_registry_modes_index` missing / corrupt / reads-json.
- `tests/controllers/test_components.py` — endpoint returns the cached map; `load()` populates it.
- `ruff check` + `ruff format --check` clean; adjacent suites (`test_sync_components_emit_split`, `test_event_payload_contracts`, `test_sync_components_pin_long_form`) green.


bdraco:
Thanks @copilot-pull-request-reviewer — both addressed in 583caf3c:

- **Loader shape validation:** `load_pin_registry_modes_index()` now returns `{}` when the payload isn't a mapping, and drops any entry whose value isn't a list (keeping only string flags), so a malformed artefact degrades to "show every flag" instead of crashing startup. Added `test_load_pin_registry_modes_tolerates_unexpected_shapes` covering `[1,2,3]` and a mixed-type mapping.
- **Test module docstring:** reflowed so the content starts on the line after the opening `"""` per CLAUDE.md.

PR #1175, opened 2026-06-04
Scope find_by_pio_board to the device's platform

bdraco:

Scopes `BoardCatalog.find_by_pio_board` to the device's ESPHome platform.

`find_by_pio_board` matched purely on the PlatformIO board string (`esphome.board`), which is platform-blind. nRF52 and rp2040 both ship a board called `adafruit_itsybitsy` (ESPHome's `nrf52` `BOARDS_ZEPHYR` and `rp2040` `BOARDS` each define it). So a device with `nrf52:\n  board: adafruit_itsybitsy` resolved to the **rp2040** catalog entry and the editor served its `GPIOn` pins — which ESPHome's nRF52 pin validator rejects (`components/nrf52/gpio.py::_translate_pin`). That violates the project's "never generate invalid configs" principle.

This adds an optional `platform` filter to `find_by_pio_board` and passes it at the two callers (`devices/metadata.py`, `devices/mutations_create.py`) — both already parse the platform from the YAML via `parse_platform_from_yaml`, so it's just wiring an existing value through. A scoped miss returns `None`, so the existing fallback ladder runs (`find_by_platform_variant`, then a free-text pin field) rather than handing back a wrong-platform entry. Unscoped calls are unchanged (back-compat).

Surfaced by the review on #1174 (nRF52 board pins). No board generation change here; this is the resolution-side fix that lets a shared PlatformIO board string resolve correctly per platform.

- fixes <link to issue>

## Types of changes

## Frontend coordination

None. The frontend resolves a device's board via the backend-supplied `board_id` → `getBoard(board_id)` (id-keyed); it does no pio-string matching of its own, and `renderPinField` already falls back to a free-text pin field when there's no board/pins. So a scoped miss degrades correctly with no frontend change.

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/boards.py
+++ b/esphome_device_builder/controllers/boards.py
@@ -122,13 +122,25 @@ class BoardCatalog:
                 return board
         return None
 
-    def find_by_pio_board(self, pio_board: str, pio_variant: str = "") -> BoardCatalogIndex | None:
+    def find_by_pio_board(
+        self,
+        pio_board: str,
+        pio_variant: str = "",
+        platform: Platform | str | None = None,
+    ) -> BoardCatalogIndex | None:
         """
         Find a board by its PlatformIO board id, preferring a matching variant.
 
         Returns the slim index entry; the caller fetches the full
         body via :meth:`get_board` when it needs pins /
         featured_components / default_components.
+
+        ``platform`` scopes the match to one ESPHome platform. nRF52 and rp2040
+        both ship a PlatformIO board called ``adafruit_itsybitsy``; without the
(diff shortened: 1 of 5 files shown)


bdraco:
Thanks @esphbot — addressed the consistency nit in 8ab1b989. Went with the "drop the redundant guard" direction rather than mirroring it: `BoardEsphomeConfig.platform` is a required, non-nullable field, so `find_by_platform_variant`'s `b.esphome.platform and ...` was dead defensive code. Removed it so both platform filters read identically (`b.esphome.platform.value == platform`) and neither implies the field can be `None`. 435 tests still pass.

PR #1179, opened 2026-06-04
Extract DevicesController collaborator wiring into _wiring.py

esphbot:

**What:** Extract `DevicesController`'s background-collaborator construction
out of the 995-line controller god-module into a dedicated
`_wiring.wire_collaborators` free function.

**Why:** `controllers/devices/controller.py` is the package's largest module
and well over the 800-line soft cap. The single biggest non-delegator block
was the ~110-line `__init__` wiring graph — construction order plus the long
comments justifying each handle. Isolating it makes the controller body the
command surface it should be and gives the wiring one cohesive home.

**How:**
- New `_wiring.py` builds the scanner / state-monitor / reachability-tracker
  / MQTT-coordinator / build-size-refresher / metadata-store graph; `__init__`
  is now its own state plus a single `wire_collaborators(self)` call.
- Construction order and the bound-method capture
  (`is_ignored=self.state.ignored_devices.__contains__`) are preserved exactly.
- Constructed handles are declared as class-body annotations so the type
  checker still sees the controller's attribute surface; the collaborator
  imports move under `TYPE_CHECKING`.
- Drop a dead duplicate of `_REGEN_FAILURE_TTL_SECONDS` (the live copy lives
  in `storage_regen.py`).

No API, runtime, or behaviour change. `controller.py` 995 → 897 lines.

## Types of changes

## Frontend coordination

## Checklist

---

Tested: full `tests/controllers/devices/` (409 passed) plus
`tests/controllers/firmware/` + `tests/e2e` (506 passed; the 2 `e2e`
`*_compile_download` failures are the network-dependent compile tests that
only run in CI). `ruff` and `mypy` clean on the changed files.


bdraco:
@esphbot rr

bdraco (changes requested):
This is not a good way to split this module

bdraco (changes requested):
No the mixin was rejected and it was split like the other modules in this rep

Issue #1190, opened 2026-06-04
[Bug] Compiler errors in the preview version


wendefeuer:
### What happened?

When I compile this code in the preview version, I get error messages. See the log.
Everything works fine in the regular version.

------------------------------------------------------------------------------------------------------
```
esphome:
  name: esphome-buero-schrank-technik
  friendly_name: ESPHome_Buero_Schrank_Technik

esp8266:
  board: d1_mini

# Enable logging
logger:
  baud_rate: 0
  
# Enable Home Assistant API
api:
  encryption:
    key: "xxxxxxxyyyyyyzzzzzzzxxxxxxxyyyyyyzzzzzzzxxxxxxxyyyyyyzyzzzzzzz"

ota:
  - platform: esphome
    password: "xxxxxxxyyyyyyzzzzzzzxxxxxxxyyyyyyzzzzzzz"

wifi:
  ssid: !secret wifi_ssid
  password: !secret wifi_password
  manual_ip:
    # Set this to the IP of the ESP
    static_ip: 192.168.0.96
    # Set this to the IP address of the router. Often ends with .1
    gateway: 192.168.0.220
    # The subnet of the network. 255.255.255.0 works for most home networks.
    subnet: 255.255.255.0
  # Enable fallback hotspot (captive portal) in case wifi connection fails
  ap:
    ssid: "Esphome-Buero-Schrank-Technik"
    password: "xxxxxxxyyyyyyzzzzzzz"

captive_portal:


sensor:
  - platform: dht
    model: DHT22 
    pin: GPIO13      # D8
    temperature:
      name: "Temperatur"
      icon: mdi:thermometer
      device_class: "temperature"
      state_class: "measurement"
      unit_of_measurement: "°C"
    humidity:
      name: "Humidity"
      icon: mdi:water-percent
      device_class: "humidity"
      state_class: "measurement"
      unit_of_measurement: "%"
    update_interval: 60s  
```

### Dashboard version

2026.5.2

### How did you install it?

Home Assistant App (preview toggle)

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell
INFO ESPHome 2026.5.2
INFO Reading configuration C:\Users\User\esphome\.esphome\.remote_builds\Nc7uJKFUh3U0o6DKioxJFsfpZwWoN5Ws\esphome-buero-schrank-technik\esphome-buero-schrank-technik.yaml...
[...]

PR #1195, opened 2026-06-04
Shorten new dashboard_id from 32 to 8 chars

bdraco:

The `dashboard_id` is a correlation / isolation token, not an auth credential (pairing authenticates via the X25519 Noise XX handshake), so its 192-bit (24-byte `token_urlsafe`) entropy was far more than needed. It also appears in the receiver-side remote-build subtree path (`.remote_builds/<dashboard_id>/`), where 32 chars eat into the Windows `MAX_PATH` budget (issue #1190).

New ids are now minted with `secrets.token_urlsafe(6)` -> exactly 8 base64url chars (48 bits), the same length the mDNS hostname suffix already derives from the id. 48 bits is ample for the handful of dashboards on a LAN (collision needs ~16M concurrent dashboards). Existing 32-char ids stay as-is -- regenerating would break paired peers that pinned the old id -- and the validation pattern (`[A-Za-z0-9_-]`) + 64-char cap already accept both lengths, so nothing else changes. This is a smaller, dependency-free alternative to the ULID idea (a ULID's leading timestamp would also have weakened the mDNS suffix it shares).

- refs https://github.com/esphome/device-builder/issues/1190

## Types of changes

## Frontend coordination

## Checklist

- [x] The code change is tested and works locally. (`tests/test_dashboard_identity.py` updated + passing)
- [ ] Architecture-level changes are reflected in `docs/ARCHITECTURE.md` and/or `docs/API.md`. (n/a)

### DIFF
--- a/esphome_device_builder/helpers/dashboard_identity.py
+++ b/esphome_device_builder/helpers/dashboard_identity.py
@@ -42,7 +42,13 @@ from typing import Any
 from ..controllers.config import metadata_transaction
 from .peer_link_identity import PeerLinkIdentity, PeerLinkIdentityStore
 
-_DASHBOARD_ID_BYTES = 24
+# 6 bytes of entropy -> exactly 8 base64url chars (no padding), the
+# same length as the mDNS hostname suffix. Short keeps the
+# remote-build subtree path (``.remote_builds/<dashboard_id>/``)
+# under Windows MAX_PATH; the id is a correlation / isolation token,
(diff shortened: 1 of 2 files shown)


bdraco:
Superseded: shortening the dashboard_id itself is the wrong lever. Replacing with a PR that shortens only the on-disk remote-build directory (first 8 chars), leaving the identity token intact.

PR #1208, opened 2026-06-04
Exit cleanly on SIGTERM received during startup

bdraco:

A SIGTERM that arrives while the dashboard is still starting up was hitting the OS default disposition and exiting 143; the HA Supervisor reports that as "App ESPHome Device Builder did not handle SIGTERM and was terminated by the default signal handler (exit code 143)". aiohttp only installs its SIGTERM handler once the server is actually serving, so a stop landing during the multi-second cold start (esphome import, catalog load, mDNS bring-up, the firmware CLI sanity subprocess) was unguarded. This traps SIGTERM at the very top of `main`, before any of that work, so a startup-window stop exits 0 cleanly; once serving begins, `web.run_app` swaps in its own graceful handler as before.

A steady-state SIGTERM already exited 0 on both old and new code, so this is purely the startup window. Verified empirically on macOS: a SIGTERM sent 0.1 to 0.3s after launch goes from 143 before the change to 0 after.

- N/A

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/__main__.py
+++ b/esphome_device_builder/__main__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import argparse
 import logging
 import os
+import signal
 import sys
 import threading
 from contextlib import suppress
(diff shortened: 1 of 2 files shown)


bdraco:
Addressed the test-isolation finding in 18c92852:

- Replaced the fixed `time.sleep(0.3)` with a readiness gate: the test now signals on the first log line. Logging is configured below the SIGTERM trap in `main`, so any output proves the trap is armed, and the server binds ~1s later, so the signal lands in the startup window deterministically on both slow and fast runners.
- Switched `tempfile.mkdtemp` to the `tmp_path` fixture, no leaked dir.
- Wrapped the body in `try/finally` that kills and reaps the child, so a failed assertion can't orphan a running dashboard.

Confirmed it still guards the regression; against the pre-fix `main` the test fails with `returncode -15` (shell 143), and passes after the fix.

bdraco:
Thanks; both Copilot points were against the pre-hardening commit:

- **tempfile.mkdtemp leak** (L60): already resolved in the current head; the integration test now uses the `tmp_path` fixture and a `try/finally` that kills and reaps the child, so nothing leaks and a failed assertion can't orphan the process.
- **Module docstring**: trimmed to a terse summary plus a one-line why, dropping the narrative.

Current head is d9abe77a.

bdraco:
Both suggestions applied in db8ea0f0:

- **Timeout budget**: tightened the inner deadlines so they sum under the outer cap; readiness gate is now 15s and `communicate` 30s (45s total < the 60s `@pytest.mark.timeout`), so the graceful `pytest.fail` surfaces before pytest-timeout would.
- **Docstring**: trimmed `test_main_installs_sigterm_handler_before_parsing` to a one-line contract; the module docstring was already trimmed for Copilot's note.

PR #1189, opened 2026-06-04
Preserve operator pairing toggle on re-pair to APPROVED

esphbot:
## What
Preserve the operator-set `enabled` toggle (and last-known `esphome_version`) when a re-pair short-circuits to APPROVED.

## Why
`request_pair` rebuilds the `StoredPairing` from scratch. When a user re-runs Pair against a receiver whose row is still APPROVED, the receiver short-circuits the inbox dance and replies APPROVED immediately — so the freshly-constructed row overwrote the existing one with `enabled=True` (the dataclass default) and `esphome_version=""`.

`esphome_version` self-heals on the next session-open, but `enabled` is purely operator-set and never refreshes. The net effect: re-pairing a receiver silently re-enabled transparent-install routing the operator had deliberately turned off via `set_pairing_enabled`.

Deferred sibling of #1188 (the silent sweep path).

## How
Before the dict overwrite, look up any prior same-pin row (same pin = same receiver identity) and carry `enabled` + `esphome_version` forward. PENDING and APPROVED both benefit; the pin equality guarantees it's the same identity.

## Testing
- New regression `test_request_pair_approved_preserves_operator_enabled_and_version`: seeds an APPROVED row with `enabled=False` + a version, mocks the wire to return APPROVED, asserts both survive the re-pair.
- Full remote_build suite: 401 passed. `ruff` clean.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`

PR #1188, opened 2026-06-04
Fire removed event when sweeping stale pairings at an endpoint

esphbot:
## What
Fire `OFFLOADER_PAIR_STATUS_CHANGED` `"removed"` (and clear the
`offloader_remote_jobs` cache) when `sweep_stale_pairings_at_endpoint`
drops a stale pairing row.

## Why
Re-pairing the same `(hostname, port)` under a fresh pin — a receiver
rotating its identity, or a different receiver taking the hostname —
sweeps the old pairing. Unlike `unpair`, the sweep dropped the row from
RAM **silently**: no bus event and no remote-jobs cleanup. So a swept
(possibly APPROVED) row lingered in every connected client's `pairings`
snapshot until a full reload, and a ghost in-flight remote-job row
survived keyed to a pin that no longer exists — the exact cross-tab /
connected-client desync the snapshot-plus-events convention exists to
prevent.

## How
Mirror `unpair` inside the sweep loop: fire the `"removed"` event per
dropped pairing and drop the matching `offloader_remote_jobs` entries.
No persistence change — the caller (`request_pair`) still owns the save
on the approval path.

## Testing
- `test_sweep_stale_pairing_fires_removed_and_clears_remote_jobs` —
  swept APPROVED row fires `"removed"` with the right payload and its
  remote-job entry is gone.
- `test_sweep_keeps_matching_pin_silent` — re-confirming the same pin
  sweeps nothing and fires no event.
- Full remote-build suite (controller + peer-link + e2e identity
  rotation + event-payload contracts): 404 passed. `ruff` + `mypy` clean.

## Types of changes
- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`

## Frontend coordination
- [x] No frontend change needed

---
### Quality Report

**Changes**: 2 files changed, 86 insertions(+), 1 deletion(-)

**Code scan**: clean

**Tests**: failed ([Errno 13] Permission denied: 'pytest')

**Branch hygiene**: clean

_Generated by [Kōan](https://koan.anantys.com)_

PR #1211, opened 2026-06-05
Fire OFFLOADER_PAIRING_ADDED on request_pair for cross-tab sync

esphbot:
## What
Fire a new `OFFLOADER_PAIRING_ADDED` bus event when `request_pair` creates an offloader pairing, so connected clients that didn't issue the command build the row from the event.

## Why
`request_pair` persisted the `StoredPairing` and returned a `PairingSummary` in the command response, but fired **no** bus event for the creation. The originating tab seeds its own row from that response (see the frontend `onPairRequestSent`), but a second connected tab gets nothing — and the later `OFFLOADER_PAIR_STATUS_CHANGED` only *flips* an already-known row (`patchOffloadPairing` no-ops on a missing key). So a second tab's pairings list stayed stale until a full reload. This is the snapshot-plus-events cross-tab desync the architecture explicitly avoids (same vein as #1188 / #1189).

## How
- New `EventType.OFFLOADER_PAIRING_ADDED` + `OffloaderPairingAddedData` TypedDict carrying every `PairingSummary` field, so a subscriber builds the row from the event alone — mirroring the receiver-side `REMOTE_BUILD_PAIR_REQUEST_RECEIVED`.
- Fired from `request_pair` for both the PENDING and the re-pair-short-circuit APPROVED branches, off the single-source-of-truth `_pairing_summary_for` projection.
- Needs the matching frontend handler (separate PR against `device-builder-frontend`) to insert the row.

## Testing
- New `test_controller_request_pair_fires_pairing_added_event` (end-to-end through the WS-command shell) asserts the event fires once with the full payload.
- `pytest tests/test_remote_build_peer_link_client.py tests/test_remote_build_controller.py tests/test_subscribe_events_cleanup.py tests/models` → 426 passed. ruff + mypy clean.

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#637

---
### Quality Report

**Changes**: 7 files changed, 106 insertions(+), 3 deletions(-)

**Code scan**: clean

**Tests**: failed ([Errno 13] Permission denied: 'pytest')

**Branch hygiene**: clean

_Generated by [Kōan](https://koan.anantys.com)_


bdraco:
Looks spot on . real gap

bdraco:
@esphbot rr

bdraco (approved):

PR #1229, opened 2026-06-05
Distribute remote compiles across all paired build servers

### LINKED ISSUE #1225: Add parallel dispatch with real time re-routing with when adding new build servers
### What happened?

I have successfully paired with two headless build servers but it seems that the dashboard always only uses one/the first to distribute compile tasks to.

It also only distributes tasks to build hosts that have already been available at the time of submitting the tasks. If I add another build host while the (very long) list of compile tasks is being worked on, the new build host is not included on the fly.

### Dashboard version

2026.5.3

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

A paired offloader now spreads remote compiles across every connected build server at once instead of only the first, and it chooses the server when a slot frees, so a host paired or freed mid queue gets used without a fresh install. Local builds are untouched; a local only setup is just a pool of one. If a build server drops mid build the compile reroutes to the next free worker, with bounded retries.

The first commit adds a busy_build_server_pins filter to the scheduler with no behaviour change; the second adds the pool plus the dispatch loop.

- fixes https://github.com/esphome/device-builder/issues/1225

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#645

## Checklist


bdraco:
Thanks for the review. Addressed in 320ee392:

**🟡 Important**

1. **JOB_STARTED re-fires on re-route** — confirmed the frontend is idempotent: `handleJobEvent` routes `job_started` to `upsertJob`, which does `_firmwareJobs.set(job.job_id, job)` (a Map keyed by `job_id`), so a repeat re-enters RUNNING on the same row rather than creating a duplicate or resetting derived state. No idempotency change needed. Companion frontend PR esphome/device-builder-frontend#645 adds the `REMOTE_PENDING` enum value + a "Waiting for a build server" hint (the source line is otherwise blank until a server binds); the frontend-coordination box is now ticked.

2. **Architecture docs** — added a "Remote build-server pool" section to `docs/ARCHITECTURE.md` (REMOTE_PENDING resolution, the dispatch loop beside the two lanes, `pick_dispatch_target`'s WAIT outcome, `_MAX_SERVER_LOSS_RETRIES`, restart re-dispatch) and updated the CLAUDE.md two-lanes note. Docs checkbox ticked.

**Re-route reset (summary flag)** — `_return_to_pool` now calls a new `FirmwareJob.clear_run_state` instead of `reset()`, so a server-loss re-route no longer stamps the log with a "dashboard restarted mid-build" notice. `reset()` is now `clear_run_state` + that notice. Regression assertion added.

**🟢 Suggestions**

1. **Cancel race in the in-flight setup window** — already handled, and now noted: `run_remote_job` registers its `cancel_event` and immediately replays `if job_id in cancel_requested: cancel_event.set()` (remote_runner.py:178), so a cancel that lands before the event exists is honored. Same code as the lane path; covered by `test_remote_compile_cancel_before_runner_registers_event_still_fires`.

2. **pairing-None re-pass** — now re-arms `wake` in the `get_pairing is None` branch so a missed peer-link-close event can't strand the compile; the next pass re-snapshots (without the unpaired pin) and resolves, so it's bounded to one extra pass. Regression assertion added (`test_unpair_race_leaves_job_pending`).

bdraco:
Thanks @copilot — both addressed in 741b0317.

Both comments describe a cancel landing in the window between `_dispatch_to_server` spawning the driver and `_drive_remote` stamping RUNNING. Worth noting that under the production scheduler this window doesn't actually open: `create_background_task` is **eager** (`create_eager_task`), so `_drive_remote`'s `begin_run` prologue (RUNNING + JOB_STARTED) runs synchronously and returns control to `_dispatch_to_server` *before* `start()` records the in-flight binding, with no `await` between — so the job goes QUEUED → RUNNING+in-flight atomically from any other task's view, and a cancel sees the RUNNING/in-flight branch (which `run_remote_job` then honors via its `cancel_requested` replay).

That said, relying on eager timing is fragile, so I added the guards you suggested to make it correct under *any* scheduling:

1. **`_drive_remote` terminal guard** — at the start of the driver, if the job is already terminal it frees its pool slot and returns without starting the build, so a late cancel can't be overwritten back to RUNNING or submitted remotely.
2. **Cancel releases the in-flight binding** — the QUEUED cancel branch now calls a new `RemoteDispatchState.discard` (drop + release), clearing both the waiting entry and any in-flight binding so the server pin is freed.

Tests added: `test_drive_remote_skips_build_when_already_terminal` (driver no-ops + frees the slot, `run_remote_job` never called) and `test_cancel_queued_but_bound_in_flight_frees_the_server` (cancel clears the binding, `busy_pins()` empty).

bdraco:
Good catch koan — real bug, fixed in 31e00f24.

`_drive_remote` only caught `RemoteServerLostError`, so any other exception escaping `run_remote_job` ran the `finally` without finalizing (status still RUNNING → `end_run`/`finalize_bookkeeping` no-op, `JOB_FAILED` never fires), leaving the job spinning forever. This was exactly the lane-vs-pool divergence the begin_run/end_run refactor was meant to close, but the *error* finalization hadn't been shared yet.

Fix mirrors the lane runner's guarantee and shares it:
- Extracted `lifecycle.finalize_unexpected_error(controller, job, exc)` — cancel intent wins (`cancel_requested` → CANCELLED), else `job.error = str(exc)` + `_finalize_terminal(FAILED)`, with the `logging.exception`.
- Both `runner.execute_job`'s `except Exception` and `_drive_remote` now call it, so an uncaught raise always produces a terminal event on either path. `_drive_remote` also re-raises `CancelledError` (run_remote_job already finalized it on the way out).

Test added: `test_drive_remote_finalizes_failed_on_unexpected_exception` — a generic raise out of `run_remote_job` lands `job.status is FAILED`, fires `JOB_FAILED`, and frees the server (`busy_pins()` empty).

bdraco:
Thanks koan. Acknowledging the startup-grace suggestion and keeping it as-is by design.

The 30s grace does also delay a fresh interactive install submitted in the first 30s after the process starts. That's an accepted trade for simplicity: nobody is realistically logging in and kicking off a build within the first 30 seconds of the dashboard starting up, so the window where it matters is effectively empty. Racing the grace against an early-exit (peer-link-open + non-empty pending) or splitting the grace to only gate the restored-job fall-to-local path both add control-flow complexity to cover a case that doesn't occur in practice. If it ever proves to be a real annoyance we can revisit, but the simple unconditional sleep is the right call for now.

Everything else (off-lane terminality, single-router pool transitions, busy-pin re-derive, WAIT vs NO_COMPATIBLE_PEER, eager-task cancel window) you've already confirmed. The commits after this review's HEAD are pure cleanup — pushing the run-state mutations onto the model (mark_running / revert_to_pending_remote), an error= kwarg on finalize_terminal, and dropping a dead tuple half in the scheduler.

PR #1238, opened 2026-06-05
Cover the off-lane CancelledError and pool-retry server-loss raise

bdraco:

Adds the two lines of test coverage Codecov flagged on #1229 after it merged. Both are real branches the higher-level pool tests bypassed by mocking `run_remote_job`, so the internal raise/re-raise never executed:

* `remote_dispatch._drive_remote`'s `except asyncio.CancelledError: raise` — a test now raises `CancelledError` out of `run_remote_job` and asserts it propagates (shutdown-cancel) while the `finally` still frees the pool slot.
* `remote_runner._await_terminal`'s `raise RemoteServerLostError` — a mid-build session loss with `retry_on_server_loss=True` now drives the real `run_remote_job` (not a mock), asserting it raises rather than finalising the job locally, so the dispatch pool can re-route it.

Tests only; no production change.

- follows up https://github.com/esphome/device-builder/pull/1229

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/tests/controllers/firmware/test_remote_dispatch.py
+++ b/tests/controllers/firmware/test_remote_dispatch.py
@@ -373,6 +373,36 @@ async def test_drive_remote_finalizes_failed_on_unexpected_exception(
     assert pool.busy_pins() == frozenset()
 
 
+async def test_drive_remote_reraises_cancelled_error_and_frees_slot(
+    firmware_controller_factory: FirmwareControllerFactory,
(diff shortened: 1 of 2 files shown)


bdraco:
Good catch — fixed in 7b212e26. Collapsed the second test's docstring to a single line (the dispatch-pool re-route rationale was the production story, which belongs in the commit, not the docstring); the assertions and inline comments carry it. Now matches the sibling CancelledError test.

PR #1260, opened 2026-06-06
Skip storage regen for secrets.yaml; centralize the secrets-file check

bdraco:

Saving secrets.yaml kicked off a storage regen the same way a device config does, but secrets.yaml has no build dir, so storage_regen logged a stale config hash warning on every secrets write. It now skips secrets.yaml at the regen entry point, so the warning is gone.

While here I pulled the secrets.yaml special case into one place, `constants.is_secrets_file` and `constants.SECRETS_FILENAME`; the literal was duplicated across the version history seed, the bundle import, the config writer, the mqtt coordinator, and the config-save path, so now only constants.py holds it. It lives in constants.py rather than secrets_state.py so lightweight modules can import it without pulling in ruamel.

- N/A; caught in the logs of a running dashboard

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/constants.py
+++ b/esphome_device_builder/constants.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from importlib.metadata import PackageNotFoundError, version
+from pathlib import Path
 
 
 def _resolve_version() -> str:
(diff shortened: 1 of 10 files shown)


esphbot (approved):
No blocking issues found.

PR #1272, opened 2026-06-06
Exit cleanly on a stop signal received during dashboard startup

### LINKED ISSUE #1270: Flakey SIGTERM test needs to be fixed in CI
https://github.com/esphome/device-builder/actions/runs/27075424004/job/79911863241?pr=1267

bdraco:

**Production bug: the dashboard could hang or exit non-zero when it received a stop signal during startup.** A SIGTERM from the supervisor (systemd, Docker, the Home Assistant add-on) arriving while the dashboard is still cold-starting could leave the process wedged or make it exit unclean instead of shutting down. The flaky `test_sigterm_during_startup_exits_zero` was the symptom that surfaced it.

Two faults in the startup signal trap (`_exit_cleanly_on_signal`):

**1. Hang.** It raised `SystemExit` inline. When the signal landed in the window after aiohttp's event loop starts but before `AppRunner.setup` arms its own SIGTERM handler, that `SystemExit` was raised inside the live loop; `run_app` does not catch `SystemExit`, so it fell into its `finally`, re-entered the loop to clean up, and blocked in `select` forever. The supervisor then waits out its grace period and SIGKILLs (exit 137). The trap now raises `SystemExit(0)` only when no loop is running (the genuine pre-loop cold start), and otherwise defers aiohttp's own `GracefulExit` through `loop.call_soon_threadsafe`, so `run_app` catches it and drains `on_cleanup` exactly as a stop received while serving does.

**2. Unclean exit.** It also called `logging.info()`. A signal handler runs between bytecodes on the main thread, frequently while the no-auth startup banner is mid write, so logging from there re-enters the stream or queue. Under synchronous logging that is a hard `RuntimeError: reentrant call inside <_io.BufferedWriter>`; under the production async queue handler it corrupts or raises and the traceback is lost because the process exits before the listener thread flushes, surfacing as a bare non-zero exit. The handler now does no logging and minimal work; the stop is logged from `_raise_graceful_exit`, which runs in safe loop context.

## History

Several iterations. The first cut hard-exited via `os._exit`, which was blunt and needed a platform branch; the second deferred `GracefulExit` and fixed the hang, after which a rare non-zero exit surfaced on Linux 3.14. That second fault was sub 0.1 percent and would not reproduce until the tests captured the child transcript and drove many in-loop signals under contention; a diagnostic run that disabled the async log queue finally exposed the reentrant logging error and pinned the cause.

## Testing

- `_exit_cleanly_on_signal` and `_raise_graceful_exit` have unit tests for the no-loop, with-loop, and raise paths.
- Subprocess tests drive a real dashboard: SIGTERM pre-serving exits 0 (transcript captured on failure), SIGTERM while serving drains `on_cleanup` (drain marker asserted), and CTRL_BREAK on Windows drains.
- A deterministic crack test widens the loop-start window via a sitecustomize shim so CI exercises the exact race every run, exiting 0 instead of hanging.
- During the hunt, the bug was reproduced on CI under contention, root caused via a synchronous-logging diagnostic that exposed the reentrant logging error, and the verification run (the same diagnostic harness that reliably hung before) is now green on all five platforms. The heavy diagnostic harnesses were removed; the lean regression set above remains.

- fixes https://github.com/esphome/device-builder/issues/1270

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/__main__.py
+++ b/esphome_device_builder/__main__.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import argparse
+import asyncio
 import logging
 import os
 import signal
@@ -100,10 +101,31 @@ def _setup_logging(log_level: str, log_file: str | None = None) -> None:
     activate_log_queue_handler()
 
 
-def _exit_cleanly_on_signal(signum: int, _frame: object) -> None:
-    """Exit 0 on a stop signal the event loop isn't trapping itself."""
-    logging.getLogger(_LOGGER_NAME).info("Received stop signal %s; shutting down cleanly", signum)
-    raise SystemExit(0)
+def _exit_cleanly_on_signal(_signum: int, _frame: object) -> None:
+    """Exit 0 on a stop signal the event loop isn't trapping itself.
+
+    Runs in signal context (between bytecodes on the main thread, possibly
+    mid-logging-write), so it must do no logging and no real work — that
+    would re-enter the stream/queue and deadlock or raise (a stop landing
+    while the startup banner is being emitted hit
+    ``RuntimeError: reentrant call inside <_io.BufferedWriter>``). With no
+    running loop (pre-serving cold start) raise ``SystemExit(0)``;
+    otherwise hand off to the loop, which runs ``_raise_graceful_exit`` at
+    a safe point so ``run_app`` drains ``on_cleanup``.
+    """
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        raise SystemExit(0) from None
+    loop.call_soon_threadsafe(_raise_graceful_exit)
+
+
+def _raise_graceful_exit() -> None:
+    """Loop-context callback: log the stop and raise ``GracefulExit`` so ``run_app`` drains."""
+    from aiohttp.web import GracefulExit  # noqa: PLC0415
+
+    logging.getLogger(_LOGGER_NAME).info("Received stop signal; shutting down cleanly")
+    raise GracefulExit
 
 
 def main() -> None:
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

esphbot (commented):
LGTM — the mechanism is right.

Walked the three states and they each line up:

- **No running loop** (true cold start, before `web.run_app` exists) → `SystemExit(0)`. Nothing to drain, interpreter unwinds cleanly instead of the OS 143 default.
- **Loop running, POSIX startup crack** (after the loop starts but before aiohttp arms its own `add_signal_handler`) → defer `GracefulExit` via `call_soon_threadsafe`. `run_app` runs in the main thread and wraps the serve in `except GracefulExit`, so the deferred raise drains `on_cleanup` exactly as a serving-time stop does. The old inline `SystemExit` unwound through `run_app`'s loop and hit the re-enter-to-cleanup path that hangs in `select` — that's the flake.
- **Windows** → no `add_signal_handler`, so the `signal.signal` trap stays live while serving; same defer path drains correctly, and SIGBREAK is wired the same way.

Two things I liked:
- `get_running_loop()` is the right discriminator here — the signal handler runs in the main thread, which is where `run_app`'s loop lives, so it correctly returns the live loop in the crack and `RuntimeError`s pre-loop. `call_soon_threadsafe` (not `call_soon`) is also correct: it pokes the self-pipe to wake a parked loop.
- Keeping the `aiohttp.web.GracefulExit` import inside the deferred `_raise_graceful_exit` callback rather than the signal handler itself — the handler stays minimal.

Tests cover all three branches behaviorally (no-loop → exit 0, loop → defers the exact callback without raising inline, callback → `GracefulExit`). Empirical crack-widening verification in the PR body is the convincing part.

No blockers.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1276, opened 2026-06-07
Resolve cross-domain id references with schema-derived component provides

### LINKED ISSUE #1275: [Bug] Sensor platform ct_clamp can not be added from Add component wizard
### What happened?

When trying to add a CT Clamp component using the Add component wizard, the field Sensor is marked as required, but the form does not allow to select an ADC Component already added to the configuration. If the Add voltage_sampler option is selected, the Choose component wizard filters components by "voltage_sampler", but no component matches the search. Even if the filter is cleared, no ADC component is valid.

Image

### Dashboard version

Latest Beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Adding a ct_clamp sensor needs a voltage_sampler reference, but the picker found nothing even with an ADC sensor already configured; the providers live under `sensor:`, not a `voltage_sampler:` block, so the frontend's top level scan never matched them.

The catalog now records what interfaces each component can be referenced as. A component's `provides` is derived from its id class plus transitive parents in the schema bundle, matched against the classes that some `use_id` field actually references, keeping only interfaces whose namespace differs from the component's own domain (same domain references already resolve through the existing top level scan). This is exactly how ESPHome resolves `use_id` itself (the `inherits_from` check), so it is deterministic and needs no curated table; matching on the full class, not the namespace, keeps near misses like `cst226::CST226ButtonListener` out.

`get_components` gains a `provides` filter so the Add component picker can list valid targets; `provides=voltage_sampler` returns the ADC family.

The catalog `provides` values were produced by the sync logic against the 2026.5.3 schema and merged so the diff is provides only; stripping `provides` back out is byte identical to the committed catalog, and a full `script/sync_components.py` run reproduces the same values.

A companion frontend PR consumes `provides` for the reference dropdown and the Add picker.

- fixes https://github.com/esphome/device-builder/issues/1275

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#683

## Checklist

- [x] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (the `provides` values are generated by `script/sync_components.py`'s own logic and merged so the diff stays provides only).


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1287, opened 2026-06-07
Own the stop signal across startup so a mid-startup quit exits 0

bdraco:

`test_sigterm_during_startup_exits_zero` flaked on macOS/Windows CI with exit 1 instead of 0. The root cause is a real, if rare, shutdown bug: aiohttp arms its own SIGTERM handler inside `runner.setup()` before `on_startup` runs, and it runs `on_startup` outside its cleanup try/finally. So a stop signal landing mid-startup cancels aiohttp's main task while a controller is half-started; that controller's teardown can raise a non-`CancelledError` which escapes `run_app`, escapes `main`, and gets logged to the async log queue that is then lost on interpreter shutdown; the process exits 1 with no visible traceback, which is exactly what the CI transcript showed.

The fix runs `web.run_app(handle_signals=False)` so our `__main__` SIGTERM/SIGBREAK trap stays the sole handler for the whole lifecycle, instead of being replaced by aiohttp's mid-startup. The trap records that a stop was requested, and `main` treats a `run()` exception as a clean exit when a stop is pending; a genuine startup crash with no stop pending still propagates. The serving-time path is unchanged; our trap defers `GracefulExit` to the loop, so `run_app` still drains `on_cleanup`.

I reproduced the exit-1 under Python 3.14 with aiohttp 3.14, added a regression test that pins it (it fails on `main`, passes here), and reworked the existing startup-crack test since `handle_signals=False` closes the aiohttp handoff window its old shim relied on. Full suite is green (4628 passed).

- fixes flaky test in https://github.com/esphome/device-builder/actions/runs/27085856656/job/79940071051

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/__main__.py
+++ b/esphome_device_builder/__main__.py
@@ -27,6 +27,7 @@ from .helpers.logging import activate_log_queue_handler
 
 if TYPE_CHECKING:
     from .controllers.config import DashboardSettings
+    from .device_builder import DeviceBuilder
 
 _FORMAT = "%(asctime)s.%(msecs)03d %(levelname)s (%(threadName)s) [%(name)s] %(message)s"
 _DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

PR #1292, opened 2026-06-07
Consolidate duplicated list-handler write-path scaffold

### LINKED ISSUE #1291: Consolidate duplicated list-handler write-path scaffold (component / light / device on_*)
### Summary

The list-shaped automation write paths share a recurring scaffold that is still duplicated across three function pairs, even after #1290 factored out the inner insert-or-replace step into `writing_lists.apply_list_entry_upsert`. This is the drift risk previously noted around #1274.

### The duplicated shape

Each upsert/delete pair does: load YAML, locate the handler's list, refuse when the existing value is a single mapping ("convert it to a list first"), `drop_after_block_comment`, mutate at index (now `apply_list_entry_upsert`), then re-emit the whole list block, removing the key when it empties.

Copies today:

- `upsert_component_on_entry` / `delete_list_entry` — component instance, re-spliced via `_resplice_list_block` -> `upsert_inline_handler` / `remove_inline_handler` (`controllers/automations/writing_lists.py`)
- `upsert_light_effect` / `delete_light_effect` — light instance, same re-splice (`writing_lists.py`)
- `_upsert_device_on_entry` / `_delete_device_on_entry` — the singleton `esphome:` block, re-spliced via `_upsert_under_top_key` / `_delete_under_top_key` (`controllers/automations/writing.py`)

### Why it is not trivially unified

The two axes that differ are (1) **container access** — a list item located by `id` under `<domain>:` vs a fixed key under the singleton `esphome:` block — and (2) the **re-splice primitive** (`_resplice_list_block` vs `_upsert_under_top_key`). A single shared function would need those passed in as callbacks; the prior DRY review judged that the explicit parallel form was defensible, so #1290 only consolidated the index step rather than the whole scaffold.

### Suggested direction (non-blocking)

Extract the scaffold into one helper parameterised by a small "list container" strategy (resolve entries, re-splice, remove-when-empty) so the single-mapping guard, comment handling, and empty-list removal can't drift between the three sites. Tracking separately so it stays a behavior-free refactor rather than riding on the #1283 bugfix (#1290).

### Type

- refactor

bdraco:

The three list-shaped automation write-path pairs (component on_*, light effects, device on_*) each carried a copy of the same scaffold; load YAML, locate the list container, refuse a single mapping, drop block comments, mutate at index, then re-emit the block. The bodies had already started to drift.

This pulls the shared skeleton into two generic functions, `upsert_list_entry` and `delete_list_entry_for`, parameterized by a small `ListContainerStrategy` holding the two things that actually vary; how the container is located and how the block is re-spliced (plus the per-container not-present message). Component strategies bind their config with `functools.partial`; the device strategy is a module constant. No behaviour change; error messages, error codes, single-mapping refusal, and empty-list removal are all preserved.

- fixes https://github.com/esphome/device-builder/issues/1291

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/automations/writing.py
+++ b/esphome_device_builder/controllers/automations/writing.py
@@ -62,12 +62,13 @@ from .writing_layout import (
     _locate_top_list_item,
 )
 from .writing_lists import (
-    apply_list_entry_upsert,
+    ListContainerStrategy,
     delete_light_effect,
     delete_list_entry,
-    drop_after_block_comment,
+    delete_list_entry_for,
     upsert_component_on_entry,
     upsert_light_effect,
+    upsert_list_entry,
     wrap_handler_list_block,
 )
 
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1296, opened 2026-06-07
Dispatch peer-link app frames via a table on the receiver side

bdraco:

The receiver-side peer-link receive loop (`peer_link/session.py:_receive_loop`) hand-branched four app-frame types (`submit_job`, `submit_job_chunk`, `cancel_job`, `download_artifacts`) in an if-chain. The client side already does this with a dispatch table (`peer_link_client/client.py:_build_sync_frame_dispatch`); this brings the receiver into parity.

Frame types are now keyed in a module-level `_APP_FRAME_DISPATCH` constant built once at import, so there is no per-session or per-frame allocation. PING / PONG / TERMINATE stay branched in the loop body since they touch session-local state or close the loop. The non-string and unknown `msg_type` paths behave as before; a non-string short-circuits the lookup and falls through to the debug-log branch.

Behaviour is unchanged; the submit/artifacts accessors still resolve per-frame inside the handlers, so the table build never touches state that only exists after `ReceiverController.start()`.

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/remote_build/peer_link/session.py
+++ b/esphome_device_builder/controllers/remote_build/peer_link/session.py
@@ -273,38 +273,63 @@ async def _receive_loop(session: PeerLinkSession, controller: ReceiverController
             # the WS will drain via the next ``CLOSE`` frame.
             session._closing = True
             return
-        if msg_type == AppMessageType.SUBMIT_JOB.value:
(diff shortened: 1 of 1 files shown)


esphbot (approved):
No blocking issues found.

PR #1302, opened 2026-06-08
Surface id-less component instances as automation targets

bdraco:

Adding an automation to a component that has no `id:` failed with "No triggers available for this target"; a named gpio binary_sensor or an uptime/template sensor could not be targeted at all. The trigger catalog already had the triggers, and the parser and writer already handled id-less components via a synthetic `<domain>_<idx>` id, but the target-surfacing step `_scope_component_instances` never emitted those instances, so the wizard had nothing to point at.

This surfaces an id-less leaf the same way, keyed on that same synthetic id, so it round-trips through parse, upsert, and delete. An id-less container (a multi-entity platform with no id of its own) stays out, since its sub-entities are the targets; that is a separate pre-existing gap. The frontend already computes the same synthetic id and looks it up in the returned devices list, so no frontend change is needed.

- No filed issue; reported in conversation against the live dashboard.

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/automations/controller.py
+++ b/esphome_device_builder/controllers/automations/controller.py
@@ -379,15 +379,17 @@ def _scope_component_instances(
         catalog_id = parsing.catalog_id(domain, item.get("platform"))
         subs = list(parsing.iter_subentities(domain, item))
         comp_id = item.get("id")
-        if comp_id:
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

PR #1307, opened 2026-06-08
Exclude component action-fields from the automation trigger catalog

bdraco:

The "+ Add automation" trigger picker offered component action-fields (Number.Template -> Set Action, Cover.Template -> Open Action, Climate.Thermostat -> Heat Action, ...) as if they were triggers; picking one failed with `invalid_args: Unknown trigger id 'number.set_action'`. Those fields are not event triggers; the component performs the user's action list on command, and they are already edited correctly through the component form's "Edit actions" surface.

Root cause: the catalog generator emitted every `type: "trigger"` schema field as a trigger, but action-fields are `type: "trigger"` in the schema too, so 81 of 379 catalogued triggers were actually action-fields. The codebase already classifies real triggers by the `on_` prefix in the component-form skip and the runtime parser; the generator's trigger extraction never applied it. This hoists the `on_` classifier into one shared helper (`helpers/automation_keys.py`), routes every `startswith("on_")` site through it, gates trigger extraction on it, and regenerates the automations catalog to drop the 81 action-field entries.

The 44 legitimate platform-scoped `on_*` triggers (ezo, nextion, touchscreen, speaker, ...) are preserved; actions, conditions, light effects, and filters are byte-identical. Note: the `api`/`script`/`interval` `then` and `opentherm.before_*` entries also leave the picker; they were never component event triggers, and the opentherm hooks remain wireable via the component action-list surface.

- No filed issue; reported in conversation against the live dashboard.

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1310, opened 2026-06-08
Preserve the !lambda tag when round-tripping automations

### LINKED ISSUE #1306: Editing a !lambda automation drops the !lambda tag and flattens multi-line bodies
### Summary

Editing an automation that contains a `!lambda` value corrupts it on round-trip; the `!lambda` YAML tag is dropped and a multi-line lambda body is flattened to one line. Dropping `!lambda` turns the C++ lambda into a plain literal string, which silently breaks the firmware (the source is sent as raw bytes instead of being executed).

### Reproduction

A `number.template` (or any component) with a `set_action:` containing an inline lambda:

```yaml
number:
  - platform: template
    name: Blockade Time
    id: Blockade_Time
    optimistic: true
    set_action:
      - uart.write: !lambda
          std::string blockades = "setInhibit " + str_sprintf("%.0f",x) + "\r\n";
          return std::vector<uint8_t>(blockades.begin(), blockades.end());
```

Parse it via `automations/parse` and re-render via `automations/upsert` (what the editor does on save of an action-list edit).

### Actual

```yaml
      - uart.write: |
          std::string blockades = "setInhibit " + str_sprintf("%.0f",x) + "\r\n"; return std::vector<uint8_t>(blockades.begin(), blockades.end());
```

Two regressions:

1. `!lambda` is replaced with a literal block scalar `|`; the tag is gone, so ESPHome treats the value as a string literal, not a lambda.
2. The two-line body is flattened onto one line.

### Expected

The `!lambda` tag is preserved and the body keeps its original line breaks; round-tripping an unedited automation is a no-op.

### Notes

- Confirmed against the parser/writer: the parsed tree captures the value as `{'data': {'_lambda': '...'}}` with the body already flattened (newlines joined with spaces) at parse time, and the renderer emits `|` rather than `!lambda`.
- Affects any automation action whose value is a `!lambda`, not just `set_action`; reproduced via a component action-field but the lambda emit path is shared.

bdraco:

Editing an automation that contains a `!lambda` value dropped the `!lambda` tag on save, and re-emitted a bare `|` block; that turned the C++ lambda into a plain string literal and silently broke the firmware (the source was sent as raw bytes instead of being executed). It also rewrote `delay: !lambda return 0;` into an invalid `delay: |` string.

The fix carries the tag through the parse to upsert wire sentinel as an optional `_tag` key, so the emitter re-emits `!lambda`; block style `!lambda |-` bodies keep their line breaks. Untagged bare `|` lambdas are left unchanged, so the common `lambda: |-` idiom still round trips as a no-op.

A plain (non block) multiline `!lambda` has its newlines folded to spaces by YAML at parse time; that fold is not recoverable from the parsed tree, so it re-emits as a single line `!lambda`, which compiles identically.

- fixes https://github.com/esphome/device-builder/issues/1306

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#696

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/automations/_decompose.py
+++ b/esphome_device_builder/controllers/automations/_decompose.py
@@ -231,18 +231,20 @@ def _render_value(value: Any) -> Any:
     """
     Convert a ruamel-parsed value to its JSON-wire shape.
 
-    Lambda block scalars (``|`` or ``!lambda`` tagged) become the
-    ``{"_lambda": "<source>"}`` sentinel; ruamel maps and lists
-    become plain dicts/lists, recursively. Tagged scalars from
-    ruamel are not JSON-serialisable on their own, so any
-    unrecognised tag falls back to its plain string value.
+    Lambda block scalars become the ``{"_lambda": "<source>"}``
+    sentinel; an ``!lambda``-tagged value additionally carries
+    ``"_tag": "!lambda"`` so the emitter re-emits the tag (dropping it
+    turns the C++ lambda into a plain string literal). ruamel maps and
+    lists become plain dicts/lists, recursively. Tagged scalars from
+    ruamel are not JSON-serialisable on their own, so any unrecognised
+    tag falls back to its plain string value.
     """
     if isinstance(value, LiteralScalarString):
         return {"_lambda": str(value)}
     if isinstance(value, TaggedScalar):
         tag = getattr(value.tag, "value", "") if value.tag is not None else ""
         if tag == "!lambda":
-            return {"_lambda": str(value)}
+            return {"_lambda": str(value), "_tag": "!lambda"}
         return str(value)
     if isinstance(value, dict):
         return {k: _render_value(v) for k, v in value.items()}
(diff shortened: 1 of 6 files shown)


esphbot (changes requested):

bdraco:
Addressed the frontend-coordination concern: opened companion PR esphome/device-builder-frontend#696 and flipped the coordination box to link it.

That PR preserves `_tag` across the three frontend paths that previously reconstructed a bare `{_lambda}`: the lambda editor carries an existing tag forward across body edits, the literal/lambda toggle sets `_tag: "!lambda"` (a templatable value field also accepts a literal, so the lambda form needs the explicit tag), and the YAML serializer emits `!lambda |-` when tagged. Untagged `lambda: |-` stays bare. So the user-facing edit flow keeps the tag end to end, not just the no-edit round-trip. Empirically verified plus the full frontend suite (2656 tests) green.

On the minor note: single-line `!lambda |` blocks normalizing to plain `!lambda` is intentional and harmless; the docstring on `_encode_lambda` documents the block-vs-plain rule.

esphbot (approved):
No blocking issues found.

PR #1337, opened 2026-06-08
Promote template-entity control fields out of advanced

### LINKED ISSUE #1324: [Bug] Optimistic/turn_*_action should be promoted to core properties in template switch
### What happened?

When adding a template switch, optimistic/turn_*_action are hidden under show advanced. They should be promoted to core properties because, without them, an error is immediately raised, so the user have to go looking for those options under advanced settings.

Image

Note: Other template components could be affected by this issue

### Dashboard version

Latest Beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

A `template` entity (template switch, cover, number, and so on) only works once its control fields are set; `optimistic` and the `*_action` handlers are conditionally required, but ESPHome marks them optional, so the catalog generator buried them under "Show advanced". Adding a template switch then raised an immediate validation error with the fix hidden away under advanced.

This adds a scoped generator pass that surfaces `optimistic` and the `*_action` fields on the main form for `.template` components, and regenerates the eight affected catalog bodies. The scope is limited to `.template` ids so it cannot promote, for example, `climate.thermostat`'s many optional `fan_mode_*_action` handlers, or a non template `optimistic` like `switch.hbridge`; `lambda` stays advanced. The body changes are the deterministic output of the new generator pass applied to the affected components; the nightly full sync reproduces them.

- fixes https://github.com/esphome/device-builder/issues/1324

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1346, opened 2026-06-08
Invalidate editor validate cache on config-dir writes

### LINKED ISSUE #1332: YAML editor lint stays stale after secrets.yaml (or any included file) write
## Summary

After a secret is written while the YAML editor is open (inline create/edit from the secret picker, or the secrets editor), the code editor keeps showing the stale `Secret '<name>' not defined in "<file>"` squiggle on the `!secret` line. It only clears once the 60s validate cache expires or the device YAML is edited (which changes the cache key). The structured editor already self-heals; the YAML lint does not.

Found while building the missing-`!secret` UX in the frontend: esphome/device-builder-frontend#706.

## Root cause

`EditorController.validate_yaml` caches the result per session keyed **only on the device-YAML content hash** (`controllers/editor.py`, `_CachedValidation` / `_VALIDATE_CACHE_TTL = 60.0`):

```python
content_hash = fnv1a_32(content.encode("utf-8"))
cached = session.cached
if cached is not None and cached.is_fresh_for(content_hash):
    return cached.result
```

Writing `secrets.yaml` (or any `!include` / `external_components` source) does not change the device YAML, so the hash is unchanged and the cached `{validation_errors: [...]}` is returned for up to 60s. The warm `esphome vscode --ace` subprocess would read the fresh `secrets.yaml` if re-run, but the cache short-circuits it. So even forcing a client re-lint returns the stale result.

This is a latent correctness bug beyond secrets: editing any externally-referenced file (`!include`d fragment, `packages:`, `external_components`) leaves dependent device validations stale for the TTL.

## Proposed fix (coordinated, two repos)

**Backend (this repo):** invalidate the editor session validate cache when a config file is written. Cleanest hook is the write path that persists `secrets.yaml` / device YAML (the `editor`/config write command behind `updateConfig`): clear `session.cached` for all `_EditorSession`s (a referenced file can affect any open device's validation). Add an `EditorController.invalidate_cache()` and call it on any config-dir write. Keep it coarse (clear all) — validations are cheap to recompute and correctness beats a 60s stale window.

**Frontend (device-builder-frontend):** on the `secrets-saved` window event (already dispatched by the secret write path), force the open editor to re-lint (`forceLinting(view)` from `@codemirror/lint`) so the now-invalidated result is recomputed promptly instead of on the next keystroke.

## Repro

1. Open a device whose YAML has `api: { encryption: { key: !secret foo } }` where `foo` is absent from `secrets.yaml`.
2. The code editor squiggles `Secret 'foo' not defined`.
3. Create `foo` via the secret picker's inline create (or add it in the secrets editor) — the structured editor clears.
4. The code editor keeps the squiggle for ~60s despite `foo` now being defined.

bdraco:

The editor's `validate_yaml` result is cached per session keyed only on the device YAML content hash (60s TTL). Writing `secrets.yaml` (via `config/set_secret`, the whole-file secrets editor, onboarding, or a bundle import) or any `!include`d file doesn't change the device YAML, so the device session's hash is unchanged and the stale `Secret '<name>' not defined` result was served for up to 60s. This adds `EditorController.invalidate_cache()` and calls it from every config-dir write path. It clears all sessions, not just the written file, since the stale result lives on the device session, not on `secrets.yaml`; a referenced file can affect any open device's validation. Validations are cheap to recompute, so clearing coarsely is the right trade.

Pairs with the frontend change that forces the open editor to re-lint promptly.

- fixes #1332

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#724

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/config/controller.py
+++ b/esphome_device_builder/controllers/config/controller.py
@@ -17,7 +17,6 @@ from ...helpers.secrets_state import (
     is_valid_secret_key,
     read_secrets_yaml,
     write_secret,
-    write_secrets_locked,
 )
 from ...helpers.storage_path import resolve_storage_path
 from ...models import ErrorCode, UserPreferences
@@ -145,8 +144,9 @@ class ConfigController:
             raise CommandError(ErrorCode.INVALID_ARGS, "overwrite must be a boolean")
         config_dir = self._db.settings.config_dir
         try:
-            created = await write_secrets_locked(
-                self._db.secrets_write_lock,
+            # Funnel through DeviceBuilder so the editor's now-stale !secret
+            # lint is dropped along with the write (see write_secrets_locked).
+            created = await self._db.write_secrets_locked(
                 partial(write_secret, config_dir, key, value, overwrite=overwrite),
             )
         except SecretsContentError as err:
--- a/esphome_device_builder/controllers/devices/controller.py
+++ b/esphome_device_builder/controllers/devices/controller.py
@@ -867,6 +867,10 @@ class DevicesController(  # noqa: PLR0904 (grandfathered; new public methods nee
         async with self._yaml_write_lock(configuration):
             await self._write_yaml_atomic_async(self._db.settings.rel_path(configuration), content)
             await self._commit_history(configuration, message or f"Update {configuration}")
+        # A write here (device YAML, or the whole-file secrets.yaml editor)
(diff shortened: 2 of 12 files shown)


esphbot (changes requested):

esphbot (approved):
No blocking issues found.

PR #1354, opened 2026-06-09
Stop preferences borrowing the esphome core page title

bdraco:

The component catalog listed `preferences` with the name "ESPHome Core Configuration", a duplicate of the real `esphome` core entry; the navigator showed "ESPHome Core" twice and the editor opened the preferences section under that title.

Root cause is in `script/sync_components.py`: `clean_docs` derives a component's name from its `See also` link text, but the schema's `preferences` doc links to a section of the esphome core page (`components/esphome#preferences-component`), so the link text is the parent page's title. A new `build_catalog` post-pass (`_fix_borrowed_page_titles`) detects when an entry's docs page is owned by a different component and its own stem is unrelated to that page slug, and falls back to the stem-derived name. Here `preferences` becomes "Preferences". The rule was validated against the full catalog and changes only the `preferences` entry (variant families like `pn532_spi` -> `pn532` are explicitly preserved).

The catalog was regenerated with `script/sync_components.py --version 2026.5.3`; unrelated upstream drift in a few other bodies was left out so the diff is just the `preferences` rename.

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist


esphbot (changes requested):

bdraco:
Addressed in 5e9f2acc: added direct unit tests for `_fix_borrowed_page_titles` (rewrite, same-family variant kept, own-page kept, no-owner kept) and dropped the redundant `slug != cid` guard the earlier checks already guarantee.

esphbot (approved):
No blocking issues found.

bdraco:
Trimmed in 356b164: the comment now just states the same-page/same-family rationale, no reference to the dropped guard.

PR #1367, opened 2026-06-09
Advertise dashboard over mDNS in HA addon mode

bdraco:

The dashboard now advertises itself over mDNS when running as a Home Assistant addon, so the ESPHome Desktop app can discover that a builder is running on the network. Previously the advertise was skipped entirely in addon mode.

The addon runs with host networking, so the announce carries the host LAN IP and reaches peers. The one wrinkle is that the container also sees the Supervisor `hassio` bridge (`172.30.32.0/23`); that address is host internal, so it is now filtered out the same way `docker0` already was, and we never advertise an unreachable `172.30.32.x` address.

The advertise is discovery only; the peer-link receiver stays default off on the addon, so the announce carries `server_version` and `esphome_version` but no `pin_sha256` or `remote_build_port`. The companion frontend PR uses that absence to hide the addon dashboard from the send builds list, since it cannot accept builds.

- No related issue

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#743

## Checklist

### DIFF
--- a/esphome_device_builder/device_builder.py
+++ b/esphome_device_builder/device_builder.py
@@ -346,24 +346,20 @@ class DeviceBuilder:
         # Reuses the state monitor's zeroconf instance so the
         # responder count stays at one per process.
         #
-        # Skipped in two cases:
-        #   * Zeroconf failed to bind — device discovery already
-        #     fails soft here, the advertise follows the same rule.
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

PR #1373, opened 2026-06-10
Resolve dotted-subdomain extends refs in component sync

### LINKED ISSUE #1372: [Bug] "internal_error: Command failed: devices/add_component" while adding "Speaker Audio Media Player component"
### What happened?

"internal_error: Command failed: devices/add_component" while adding "Speaker Audio Media Player component"

Image

Image

Base config:
```
# Board: Waveshare ESP32-S3-Zero (Waveshare)
# Definition: definitions/boards/waveshare_esp32_s3_zero/manifest.yaml

esphome:
  name: mediaplayer-test
  friendly_name: Mediaplayer Test

esp32:
  variant: esp32s3
  flash_size: 4MB
  framework:
    type: esp-idf

logger:

api:
  encryption:
    key: "QA5x8ZU6I3KbjTfcX77H+MeDU6lHvBiPnej/O7gi0F4="

ota:
  - platform: esphome

wifi:
  ssid: "!secret wifi_ssid"
  password: "!secret wifi_password"
  ap:
    ssid: Mediaplayer Tes Fallback Hotspot
    password: "Jr99zIDjk3gQ"

captive_portal:

```

### Dashboard version

v0.1.0b109

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Adding the "Speaker Audio Media Player" component failed with `internal_error: Command failed: devices/add_component`. Two backend bugs combined.

First, the catalog generator could not resolve a nested field whose schema extends a platform sub-namespace ref like `speaker.media_player.PIPELINE_SCHEMA`; `_lookup_schema_ref` split the ref on the first dot and looked for `media_player.PIPELINE_SCHEMA`, but the bundle keys that entry whole as `speaker.media_player` with the schema named bare `PIPELINE_SCHEMA`. The field then shipped with empty `config_entries`, so the editor rendered no form for it and the required `announcement_pipeline` could never be supplied. The fix resolves an extends ref by its bare schema name, preferring the exact dotted-domain key; this restores the dropped sub-fields for the speaker media player pipelines plus ~25 other multi-channel sensor and climate sub-schemas.

Second, once the pipeline fields render, the YAML generator injected an invalid `name` into the `announcement_pipeline` block; the nested sub-block autofill assumed every `platform_type` group was an entity sub-reading needing name and id. It now only fills a field the sub-schema actually declares, so the pipeline gets an id and the user's speaker, no spurious name. Verified end to end against `esphome config`.

The last commit regenerates the catalog against schema 2026.5.3 so the fix can be verified in the editor; it is meant to be reverted before merge so the nightly sync regenerates the real catalog.

- fixes https://github.com/esphome/device-builder/issues/1372

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/helpers/yaml/component.py
+++ b/esphome_device_builder/helpers/yaml/component.py
@@ -152,21 +152,25 @@ def generate_component_yaml(  # noqa: C901, PLR0912
     # empty. ESPHome multi-sensor parents (HLW8012, BME280, ...)
     # expose their readings as ``platform_type``-tagged ConfigEntry
     # blocks; an unnamed sub-sensor won't surface in HA, and one
-    # without an id can't be referenced from automations.
+    # without an id can't be referenced from automations. Only fill a
+    # field the sub-schema declares: config sub-blocks like the speaker
+    # media_player ``announcement_pipeline`` are ``platform_type``-tagged
+    # too but take no ``name``.
     for entry in component.config_entries:
         if not entry.platform_type or not entry.config_entries:
             continue
         sub = fields.get(entry.key)
         if not isinstance(sub, dict):
             continue
+        sub_keys = {c.key for c in entry.config_entries}
         if sub.get("name") and sub.get("id"):
             continue
         # Build a fresh dict with name/id at the front so the emitted
         # YAML reads naturally (humans put name/id first).
         autofill: dict[str, Any] = {}
-        if not sub.get("name"):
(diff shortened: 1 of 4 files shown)


esphbot (changes requested):

esphbot (changes requested):

PR #1394, opened 2026-06-12
Derive structured editors for typed_schema fields

### LINKED ISSUE #1389: [Bug] Structured editor for font.file is incomplete
### What happened?

The structured editor for font.file appears to be incomplete. References to gfonts are not recognized.

Docs: 
https://esphome.io/components/font/

File:
[Sendspin_Zero-Display-Digital.yaml](https://github.com/user-attachments/files/28848836/Sendspin_Zero-Display-Digital.yaml)

Screenshot:
Image

### Dashboard version

v0.1.0b110

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The component catalog dropped every `typed_schema` field down to a plain string, so the visual editor could not edit them; `font.file` is the case in the linked issue. The sync now converts a schema `typed` node into a nested group with a type selector plus the per variant fields gated by `depends_on`, so the structured editor works. This also covers `animation.file`, `image.file`, `external_components.source`, `micro_wake_word.model`, and the display menu items.

`font.file` is special; a wrapper hides its `typed_schema` from the published schema bundle, so the bundle carries only a string. The bundle does still keep the shared `font.EXTERNAL_FONT_SCHEMA` (weight, italic, refresh), so the sync supplies a small static local/gfonts/web skeleton and pulls those shared fields back in via `extends`, then feeds the result through the same generic converter. No esphome import is needed; this is the permanent path, since the published schema cannot express the field.

While here, `_lookup_schema_ref` is now cached; that removes repeated JSON reads on the extends path and cuts a full sync from about 15s to 11s.

This is code only; the generated component catalog is left out so the nightly sync regenerates it on the standard schedule.

- fixes https://github.com/esphome/device-builder/issues/1389

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

script/sync_components.py
```diff
@@ -2203,6 +2218,169 @@ def _resolve_extends_maybe(ref: str, schema_dir: Path) -> str | None:
...
+        else:
+            for type_name, child in occurrences:
+                variant_children.append(
+                    {**child, "depends_on": typed_key, "depends_on_value": type_name}
+                )
+
+    return [discriminator, *_sort_entries(variant_children)]
+
+
+def _recover_font_file_node(schema_dir: Path) -> dict | None:
```
bdraco: need to iterate _recover_font_file_node a bit more .. still a bit more complex than I would like

bdraco:
Type isn't showing in the UI but I think it's a separate bug

script/sync_components.py
esphbot: `_recover_font_file_node` gone in current diff. Replaced by static `_FONT_FILE_NODE` plus `_RAW_NODE_OVERRIDES`. No closure, no esphome import. Skeleton supplies local/gfonts/web; `extends` pulls shared knobs from the bundle. Simpler than the version this comment landed on.

esphbot (approved):
No blocking issues found.

PR #1409, opened 2026-06-12
Derive structured editors for top-level typed schemas

### LINKED ISSUE #1397: Ethernet lacks a structured editor
Since https://github.com/esphome/bluetooth-proxies does use ethernet and people do change these, it would be nice to have an editor there since we do change fields (`clk` recently) and it would make it easier for users to adapt to new schema

bdraco:

Components whose whole `CONFIG_SCHEMA` is a `cv.typed_schema` union got no structured editor; the generator only understood plain schema nodes, so ethernet, spi, usb_uart, i2s_audio, template.output and about twenty more shipped empty bodies and fell back to YAML only.

The typed converter from #1394 is now wired into the top-level extraction path, and its gating is set-valued instead of per-variant copies. A field defined identically across a subset of variants is emitted once with the new `depends_on_value_any` list on `ConfigEntry`; the frontend shows it only while the selected type is in the list. Without the dedupe, ethernet's 15 PHY types sharing the RMII and SPI base schemas exploded to 117 entries and 417KB of JSON; with it the body is 23 entries. The component id now also threads through the typed path, so deprecated field filtering and per-component overrides apply there too.

The catalog is not regenerated here; the nightly catalog sync picks it up. Until the bumped frontend ships, an older frontend ignores the new field and shows subset-shared fields ungated, which is cosmetic and transitional.

Verified live with the companion frontend: ethernet renders a structured editor, switching the type from LAN8720 to W5500 swaps the RMII pin fields for the SPI ones, and the YAML round-trips flat. Live testing also caught `ethernet.clk`: the schema says optional but a final-validate step requires it for RMII PHYs (the deprecated `clk_mode` migrates into it, removal 2026.7.0), so a `_FIELD_OVERRIDES` row marks it required on the main form while `clk_mode` stays behind the advanced toggle.

- fixes https://github.com/esphome/device-builder/issues/1397

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#783

## Checklist

### DIFF
--- a/esphome_device_builder/models/common.py
+++ b/esphome_device_builder/models/common.py
@@ -466,9 +466,11 @@ class ConfigEntry(DataClassORJSONMixin):
     suggestions: list[ConfigPrimitive] | None = None
 
     # === conditional visibility ===
-    # `depends_on_value` and `depends_on_value_not` are mutually
-    # exclusive — set at most one. Frontend hides the entry when the
-    # predicate fails.
+    # `depends_on_value`, `depends_on_value_not` and
+    # `depends_on_value_any` are mutually exclusive — set at most one.
+    # Frontend hides the entry when the predicate fails. New catalog
+    # output emits only `depends_on_value_any`; the scalar forms remain
+    # for catalogs generated before it existed.
 
     # Key of another entry in the same component this entry depends on.
     # When None the entry is always visible.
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

PR #1413, opened 2026-06-12
Match the existing list indent when splicing a component

### LINKED ISSUE #1411: Add component into a non 2 space indented block emits invalid YAML
Backend twin of #1403 (the frontend side was fixed in esphome/device-builder-frontend#786). The line based YAML splice helpers in `esphome_device_builder/helpers/yaml/component.py` assume the canonical 2 space layout when inserting into an existing block, so "Add component" on a config that uses a column 0 (zero indented) or otherwise non 2 space list produces YAML that fails to parse.

**1. `_splice_into_domain_block` (~line 275) inserts the new item at the canonical indent regardless of the existing items' dash indent.** Verified repro:

```python
existing = "switch:\n- platform: template\n  name: a\n"
block = "switch:\n  - platform: gpio\n    pin: GPIO11\n"
_splice_into_domain_block(existing, "switch", block)
```

produces

```yaml
switch:
- platform: template
  name: a
  - platform: gpio
    pin: GPIO11
```

which PyYAML rejects ("while parsing a block mapping"). The same applies to a 4 space list, the spliced dash sits at a different indent than the sequence it should join. `_splice_into_multi_conf_block` shares the path via `_splice_into_domain_block`.

Fix shape: derive the dash indent from the existing block's first list item line and re indent `block_lines[1:]` to match before splicing, mirroring how the frontend now derives the item indent from the section's first content line.

**2. `_mapping_body_to_list_item` (~line 338) anchors the dash with a `startswith(ESPHOME_YAML_INDENT * 2)` check.** A mapping body indented with a single space never satisfies it, the dash is silently never inserted, and the subsequent splice mixes mapping keys with a list item under the same key. Verified:

```python
_mapping_body_to_list_item([" sda: GPIO21", " scl: GPIO22"])
# ['   sda: GPIO21', '   scl: GPIO22']   <- no "- " anywhere
```

Bodies at 3 or more spaces happen to survive (the output is valid though oddly aligned). Fix shape: strip the body's actual leading indent and re emit at the canonical level instead of prefix arithmetic against a fixed constant.

Note `_find_top_level_block_bounds` and `block_body_is_list` are already indent agnostic, only the two writers above assume the canonical layout.

bdraco:

The line based splice helpers assumed the canonical 2 space layout, so "Add component" into a zero indented or 4 space list emitted YAML that fails to parse, and a 1 space mapping body silently lost its dash during multi_conf normalisation.

The splice now derives the dash indent from the existing block's first list item and re indents the generated item to match, mirroring the frontend fix in esphome/device-builder-frontend#786; mapping bodies convert to list form by stripping their actual indent instead of prefix arithmetic against the canonical constant.

- fixes #1411

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/helpers/yaml/component.py
+++ b/esphome_device_builder/helpers/yaml/component.py
@@ -272,12 +272,34 @@ def _find_top_level_block_bounds(file_lines: list[str], key: str) -> tuple[int, 
     return block_start, block_end
 
 
+def _leading_ws(line: str) -> str:
+    """Leading whitespace of *line*."""
+    return line[: len(line) - len(line.lstrip())]
+
+
+def _list_item_indent(file_lines: list[str], header_idx: int, end_idx: int) -> str:
+    """
+    Dash indent of the first list item in a block body.
+
+    Falls back to the canonical indent when the body holds no item.
+    """
+    for idx in range(header_idx + 1, end_idx):
+        stripped = file_lines[idx].strip()
+        if not stripped or stripped.startswith("#"):
+            continue
+        if stripped.startswith("- ") or stripped == "-":
+            return _leading_ws(file_lines[idx].rstrip("\n\r"))
+    return ESPHOME_YAML_INDENT
+
+
 def _splice_into_domain_block(existing: str, domain: str, block: str) -> str | None:
     """
     Insert the platform-list item from *block* under an existing ``<domain>:``.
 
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1429, opened 2026-06-13
Capture bus constraints from FINAL_VALIDATE_SCHEMA into the catalog

### LINKED ISSUE #1425: Adding an I2C component with a frequency constraint (ags10) yields an invalid bus frequency
### What happened?

Adding a component that constrains its I²C bus frequency (e.g. `ags10`) reuses the existing `i2c:` bus but never sets a compatible `frequency:`, so the generated config fails ESPHome's final validation.

Repro:
1. Config with a plain `i2c:` bus (no `frequency:`, so it defaults above 15kHz).
2. Add component → sensor → AGS10 → Add.
3. Editor squiggle on `i2c:`: `Component ags10 cannot be used with a frequency of over 15khz for the I2C bus.`

The add itself produces structurally valid YAML (`sensor: - platform: ags10\n  i2c_id: i2c_1\n  tvoc:\n    name: TVOC`); the incompatibility only surfaces at compile-time, because the dashboard has no awareness of the bus-frequency constraint.

### Root cause

`ags10` (and other slow I²C devices) declare the limit via a final-validation against the bus, not a regular schema field:

```python
# esphome/components/ags10/sensor.py
FINAL_VALIDATE_SCHEMA = i2c.final_validate_device_schema("ags10", max_frequency="15khz")
```

`i2c.final_validate_device_schema` (esphome/components/i2c/__init__.py:297) injects a `cv.Required(frequency)` Range onto the *hub* schema. The component catalog is built from the per-component schema and doesn't carry these cross-component (device → bus) constraints, so neither the add-component dependency flow nor the form knows the existing bus needs `frequency: 15kHz`.

### Possible directions

- Sync-script: introspect `FINAL_VALIDATE_SCHEMA` to capture min/max bus frequency (and timeout) constraints into the catalog entry.
- Add-component flow: when the chosen component constrains a bus it's attaching to, either set a compatible `frequency:` on the bus or warn the user before Add.

Separate from the `ags10` prefill fix (esphome/device-builder-frontend#805), which is what surfaced this while testing.

### Dashboard version

dev (main)

bdraco:

Components can constrain the bus they attach to via ESPHome's final_validate_device_schema; ags10 caps i2c at 15kHz, 30 uart devices pin an exact baud rate, parity or stop bits, and spi/uart devices require specific pins. The catalog carried none of this, so a bus added through the dependency flow came up with an invalid default (the #1425 repro).

The sync script now reads the helper's literal kwargs from the component module's source with ast and emits them as a `bus_constraints` field on the catalog entry; 45 components gain it (regenerated from the same schema version, 2026.5.3). The frontend dep-add prefill that consumes the field lands separately.

- fixes #1425 (backend half)

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#807

## Checklist


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1430, opened 2026-06-13
Queued offline updates

rwalker777:

Add detection when device is offline to local build and add a queue flag that triggers an OTA when the device wakes.

- fixes <link to issue>

## Types of changes

- [X] New feature (non-breaking change which adds functionality) — `new-feature`

## Frontend coordination

- [X] Companion frontend PR: esphome/device-builder-frontend#911

## Checklist

- [X] The code change is tested and works locally.
- [X] Pre-commit hooks pass (`ruff`, `codespell`, yaml/json/python checks).
- [X] Tests have been added or updated under `tests/` where applicable.
- [X] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (regenerate via `script/sync_components.py` if a sync is needed).
- [X] Architecture-level changes are reflected in `docs/ARCHITECTURE.md` and/or `docs/API.md`.


rwalker777:
Will submit frontend to show queued status and option to cancel.  Feedback appreciated.

esphbot (changes requested):

esphbot (changes requested):

rwalker777:
This is dependent on this ESPHome PR - https://github.com/esphome/esphome/pull/16358

esphbot (changes requested):

esphbot (changes requested):

rwalker777:
@codspeedbot fix this regression

esphbot (changes requested):

jesserockz:
Hi @rwalker777, Just commenting to let you know that this is a feature that we want. Please address the above bot comments and conflict and mark as ready for review and then someone will take a better look.

Jesse

rwalker777:
@jesserockz Understood, but we need this first.  Else we run the risk of bricking a device that goes back to sleep too quickly.
https://github.com/esphome/esphome/pull/16358.  I had pinged bdraco to review.  Will leave it to you guys, but I think it should default to 3 min or 5 min, I put it at zero so there was no behavior change...

rwalker777:
@jesserockz Took a stab at the frontend here, but could use some help - https://github.com/esphome/device-builder-frontend/pull/911.  The backend looks good.

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -69,11 +72,43 @@ def __init__(self, device_builder: DeviceBuilder) -> None:
...
+
+    def _handle_device_wake(self, event: Event) -> None:
+        """Intercept device wake to trigger queued updates and prevent flapping."""
+        if event.data["state"] != DeviceState.ONLINE.value:
+            return
+
+        config = event.data["configuration"]
+        device = self._device_for_configuration(config)
+
+        if device and getattr(device, "queued_update", False):
```
bdraco: getattr looks like an issue here

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -69,11 +72,43 @@ def __init__(self, device_builder: DeviceBuilder) -> None:
...
+            _LOGGER.info("Device %s woke up. Triggering queued offline update.", config)
+            if self._db.devices:
+                self._db.devices.set_queued_update(device.name, is_queued=False)
+
+            create_eager_task(self.upload(configuration=config, port="OTA"))
```
bdraco: check if this is a dangling task

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -166,6 +201,16 @@ async def upload(self, *, configuration: str, port: str = "", **kwargs: Any) ->
     async def clean(self, *, configuration: str, **kwargs: Any) -> FirmwareJob:
         return await clean_mod.clean(self, configuration=configuration)
 
+    @api_command("firmware/clear_queued_update")
+    async def clear_queued_update(self, *, configuration: str, **kwargs: Any) -> None:
+        """Manually clear the queued_update flag for a device."""
+        await self._validate_configuration_boundary(configuration)
+
+        device = self._device_for_configuration(configuration)
+        if device and getattr(device, "queued_update", False) and self._db.devices:
```
bdraco: getattr is a bit of a code smell here

esphome_device_builder/controllers/firmware/controller.py
esphbot: Agreed. `queued_update` now a declared `Device` field. `getattr(device, "queued_update", False)` default never fires; masks wiring bugs as no-ops. Same smell at `_device_for_configuration`'s `getattr(d, "configuration")`, `hasattr(self, "_state_monitor")`, and `getattr(job, "is_deferred_install", False)`. Use direct attribute access.

esphbot (changes requested):
Blocking issues found.

- Deferred install silently dropped if device wakes during compile
- queued_update cleared before OTA runs — failed flash loses the queue with no retry
- Per-event O(n) device scan on the DEVICE_STATE_CHANGED hot path

esphome_device_builder/controllers/firmware/controller.py
esphbot: Valid. Returned task unretained. Coroutine suspends at `_validate_configuration_boundary` await. GC can drop it mid-flight, install silently lost. Retain via `_db.create_background_task` or hold in a `set`, discard on done. Same at the `_execute_job` ONLINE branch.

esphome_device_builder/controllers/firmware/controller.py
esphbot: Agreed. `is_deferred_install` now declared on `FirmwareJob`. Default never fires. Masks wiring bugs as no-ops. Use `job.is_deferred_install`. Same smell remains at the upload-clear `getattr(device, "queued_update", False)`.

esphbot (changes requested):
Blocking issues found.

- Wake handler re-fires OTA on every flap, interrupting an in-flight flash
- Fire-and-forget OTA task is unretained — can be GC'd mid-flight

esphome_device_builder/controllers/devices/controller.py
```diff
@@ -337,6 +338,20 @@ def get_ota_address_cache_args(self, configuration: str, port: str | None) -> li
...
+    def set_queued_update(self, name: str, *, is_queued: bool) -> bool:
+        """Public API to mutate the queued update flag for a device."""
+        if self._state_monitor:
+            return self._state_monitor.apply_queued_update(name, is_queued=is_queued)
+        return False
+
+    def _on_queued_update_change(self, name: str, is_queued: bool) -> None:  # noqa: FBT001
+        """Handle offline queued update flag transitions and persist."""
+        for device in self.get_devices():
+            if device.name == name:
```
bdraco: nit: reverse condition and continue to save indent

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -68,12 +71,54 @@ def __init__(self, device_builder: DeviceBuilder) -> None:
...
+    def _device_for_configuration(self, configuration: str) -> Any | None:
+        """Resolve a Device by its configuration filename."""
+        if self._db.devices is None:
+            return None
+
+        # Direct, typed call. Test stubs must implement get_devices().
+        return next(
+            (d for d in self._db.devices.get_devices() if d.configuration == configuration),
+            None,
+        )
```
bdraco: nit: linear search should be avoided

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -68,12 +71,54 @@ def __init__(self, device_builder: DeviceBuilder) -> None:
...
+            return
+
+        config = event.data["configuration"]
+
+        if config not in self._armed_deferred_installs:
+            return
+
+        device = self._device_for_configuration(config)
+
+        if device and device.queued_update:
```
bdraco: nit: early return if there is nothing to do

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -166,6 +211,17 @@ async def upload(self, *, configuration: str, port: str = "", **kwargs: Any) ->
     async def clean(self, *, configuration: str, **kwargs: Any) -> FirmwareJob:
         return await clean_mod.clean(self, configuration=configuration)
 
+    @api_command("firmware/clear_queued_update")
+    async def clear_queued_update(self, *, configuration: str, **kwargs: Any) -> None:
+        """Manually clear the queued_update flag for a device."""
+        await self._validate_configuration_boundary(configuration)
+
+        device = self._device_for_configuration(configuration)
+        if device and device.queued_update and self._db.devices:
```
bdraco: nit: early return if there is nothing to do

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -388,6 +455,47 @@ async def _run_queue(self) -> None:
...
+        ):
+            device = self._device_for_configuration(job.configuration)
+            if device and device.queued_update and self._db.devices:
+                if job.status == JobStatus.COMPLETED:
+                    self._db.devices.set_queued_update(device.name, is_queued=False)
+                    self._armed_deferred_installs.discard(job.configuration)
+                else:
+                    # Attempt failed or got cancelled (e.g. by a manual
+                    # retry) while still queued — re-arm for the next wake.
+                    self._armed_deferred_installs.add(job.configuration)
```
bdraco: this logic is now very complex. It should get some named constants/frozenset and refactoring to reduce the complexity

esphbot (changes requested):
Blocking issues found.

- Armed deferred-install state is memory-only — lost on restart, defeating the persisted flag

esphome_device_builder/controllers/firmware/controller.py
rwalker777: This doesn't get called hundreds or thousands of times where it would matter right?

bdraco: Its avoidable though so we should fix it

esphbot: Correct. Not hot. Call gated behind `_armed_deferred_installs` membership check. Runs only when a deferred install is armed for that config — not per `DEVICE_STATE_CHANGED` event. Still avoidable though, see below.

esphbot: Agreed. `scanner.get_by_configuration` is already O(1) keyed — used in `firmware_sync.py:171,227`. Expose it on `DevicesController`, replace the `next(...)` scan. Conftest stub then implements `get_by_configuration` instead of `get_devices`.

esphome_device_builder/controllers/firmware/controller.py
esphbot: Addressed. Combined method split into `_handle_deferred_compile_completion` + `_handle_ota_upload_completion`. Terminal statuses now `_TERMINAL_UPLOAD_STATUSES` frozenset. Each branch single-concern.

esphbot (changes requested):
Blocking issues found.

- FirmwareController.stop() is never invoked — listener teardown is dead code

esphome_device_builder/controllers/firmware/controller.py
esphbot: Valid. `_handle_device_wake` already guard-clauses; collapse remaining nested checks the same way. Non-blocking.

esphome_device_builder/controllers/firmware/controller.py
esphbot: Valid. `clear_queued_update` can early-return on `not device or not device.queued_update`. Already structured that way — confirm no residual nesting.

esphome_device_builder/controllers/devices/controller.py
esphbot: `_on_queued_update_change` already uses `if device.name != name: continue`. Loop fans out across all name-matches by design (mDNS dedupe). No further reversal needed.

esphbot (changes requested):
Blocking issues found.

esphome_device_builder/controllers/firmware/controller.py
rwalker777: Feel free to tackle that one.

bdraco: https://github.com/esphome/device-builder/pull/1778

bdraco: Available once you merge in main. needs to be implemented here still

esphbot (changes requested):
Blocking issues found.

- Deferred install silently lost if the immediate OTA upload fails (online-at-compile-completion path)

esphbot (changes requested):
Blocking issues found.

- Deferred install silently lost when the online-at-compile-completion OTA upload fails

esphbot (approved):
No blocking issues found.

rwalker777:
At 917 lines this is SUPER done!  With a bow and cherry on top.  I feel the AI thinks it will *DIE* if it doesn't find SOMETHING on every review.  I even fixed documentation formatting... I don't even do that in my job!

esphome_device_builder/controllers/devices/controller.py
```diff
@@ -343,6 +344,21 @@ def get_ota_address_cache_args(self, configuration: str, port: str | None) -> li
             return []
         return self.get_address_cache_args(configuration)
 
+    def set_queued_update(self, name: str, *, is_queued: bool) -> bool:
+        """Public API to mutate the queued update flag for a device."""
+        if self._state_monitor:
+            return self._state_monitor.apply_queued_update(name, is_queued=is_queued)
+        return False
```
bdraco: Hmm.. maybe this should be by configuration and not name since thats the unique key.. Need to think about collisions here

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -119,8 +164,17 @@ async def start(self) -> None:
...
+        self._rehydrate_armed_deferred_installs()
         self._runner_task = self._db.create_background_task(self._run_queue())
 
+    def _rehydrate_armed_deferred_installs(self) -> None:
+        """Re-arm devices whose ``queued_update`` survived a restart."""
+        if self._db.devices is None:
+            return
+        for device in self._db.devices.get_devices():
+            if device.queued_update:
+                self._armed_deferred_installs.add(device.configuration)
```
bdraco: Need to check this linear search

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -213,6 +280,17 @@ async def install(
...
         await self._validate_configuration_boundary(configuration)
+
+        if port == "OTA":
+            device = self._device_for_configuration(configuration)
+            # Gated ONLY on OFFLINE, avoiding UNKNOWN startup states
+            if device and device.state == DeviceState.OFFLINE:
+                _LOGGER.info("Device %s is offline. Queuing compile-only job.", configuration)
+                build_source = self._resolve_install_source(force_local=True)
+                job = self._create_job(configuration, JobType.COMPILE, build_source=build_source)
+                job.is_deferred_install = True
```
bdraco, line 326: We should probably move `is_deferred_install` into `_create_job` so it has a single setter

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -383,6 +461,78 @@ async def _run_queue(self) -> None:
...
+        touch the offline-queue machinery just because the device happens to
+        also have ``queued_update`` set for an unrelated reason.
+        """
+        is_terminal_ota_upload = (
+            job.job_type == JobType.UPLOAD
+            and job.port == "OTA"
+            and job.status in _TERMINAL_UPLOAD_STATUSES
+        )
+        if not is_terminal_ota_upload:
+            return
```
bdraco: Might want to put this on job.XXX instead so we have it in one place

esphome_device_builder/controllers/firmware/controller.py
```diff
@@ -383,6 +461,78 @@ async def _run_queue(self) -> None:
...
+        path (``is_deferred_install``) does anything here — a plain compile,
+        or one that failed, has nothing to act on.
+        """
+        is_deferred_compile_success = (
+            job.job_type == JobType.COMPILE
+            and job.status == JobStatus.COMPLETED
+            and job.is_deferred_install
+        )
+        if not is_deferred_compile_success:
+            return
```
bdraco: Might want to put this on job.XXXX like the others so we have it in one place. need to think about it

esphome_device_builder/controllers/devices/controller.py
esphbot: Confirmed — real collision. Flag mutation keys on name; everything else (`_armed_deferred_installs`, wake handler, metadata store) keys on configuration. Two configs sharing an `esphome.name` (copy, `dashboard_import` sibling): `set_queued_update(device.name, ...)` fans out to both. Only the compiled config gets armed. Sibling shows queued, has no artifact, never flashes, never self-clears. Actual flash stays safe — wake handler uploads `event.data["configuration"]`, per-config. Harm limited to persisted/displayed flag on siblings. Fix: config-keyed apply path mirroring `get_by_configuration`, not the mDNS-style name fan-out (that pattern fits mDNS broadcasts, not a per-config firmware job).

esphbot (changes requested):
Blocking issues found.

- Queued-update flag keyed on name, not the unique configuration — sibling-config collision

esphome_device_builder/controllers/_device_state_monitor/controller.py
```diff
@@ -511,3 +530,14 @@ def get_cached_dns_addresses(self, host_name: str) -> list[str] | None:
...
+    def _find_device_by_configuration(self, configuration: str) -> Device | None:
+        """
+        Linear scan for the one device with this unique *configuration*.
+
+        No configuration-keyed index exists on the monitor (only the
+        name-keyed ``_get_devices_by_name``), and this path isn't hot
+        enough — one queued-update flip per completed compile — to
+        warrant adding one.
+        """
+        return next((d for d in self._get_devices() if d.configuration == configuration), None)
```
bdraco: need to see if we can avoid linear search here as well.

bdraco:
This feels like its just about ready.  Need to do functional testing with it

esphome_device_builder/controllers/_device_state_monitor/controller.py
esphbot: Confirmed avoidable. Caller `set_queued_update` already resolves the device O(1) via `scanner.get_by_configuration` (dict lookup). `apply_queued_update` then re-scans linearly only for the dedup check. Pass the resolved device in, or dedup in the caller — drops the scan entirely.

esphome_device_builder/controllers/firmware/controller.py
esphbot: Startup-only path. One pass over all devices at `start()`. No index avoids iterating each device once here. Cold, bounded — acceptable as-is.

esphome_device_builder/controllers/firmware/controller.py
esphbot: Already done. `_create_job` now takes `is_deferred_install` and sets `job.is_deferred_install`. Single setter.

esphome_device_builder/controllers/firmware/controller.py
esphbot: Already on the model. `is_terminal_ota_upload` is a property on `FirmwareJob`. Logic lives in one place.

esphome_device_builder/controllers/firmware/controller.py
esphbot: Already on the model. `is_deferred_compile_success` is a property on `FirmwareJob`. Consolidated with the other job predicates.

esphbot (changes requested):
Blocking issues found.

- Redundant linear device scan — caller already resolved the device O(1)

bdraco:
Brought this up to date with main through the rename decomposition (#1819): resolved the conflicts in models/firmware.py (both property sets kept) and the controller import block, and quieted the FBT001 ruff hit on _on_queued_update_change (the monitor's Callable contract forces the positional bool). Full suite (5764), mypy, and ruff are green locally; the offline-queue machinery and the rename chain stay disjoint since is_terminal_ota_upload only matches UPLOAD jobs with port OTA and rename tails carry the old device address in port. One edge worth a look while reviewing: a device with a queued update that gets renamed before it wakes leaves _armed_deferred_installs keyed on the old configuration, so the arm goes stale until a restart rehydrates from the migrated queued_update flag (or firmware/clear_queued_update).

esphbot (changes requested):
Blocking issues found.

bdraco:
All three review findings are addressed on the branch. d6d1bd8 removes _armed_deferred_installs entirely rather than re-keying it: Device.queued_update (which the rename migration already carries to the new filename) is the single arm state, the wake handler derives from it with an active-flash check as the flap guard, and the rehydrate step disappears since the persisted flag is the arm. The same commit collapses the set_queued_update round-trip — the dedup now happens in DevicesController.set_queued_update against the O(1)-resolved device, and the monitor's apply_queued_update, callback plumbing, and linear _find_device_by_configuration are gone (queued_update is dashboard-internal, not broadcast-driven, so it never belonged on the monitor). 320a278 fixes the stranded-arm loop: firmware/clean disarms the wiped configuration and firmware/reset_build_env disarms every armed device, so a clean-while-armed device no longer OTA-fails on every wake. Full suite, ruff, and mypy green.

esphbot (changes requested):
Blocking issues found.

- Flap re-entrancy window: wake handler can dispatch a second OTA that supersedes the first

bdraco:
Three more commits on the branch. 47e254f adds an e2e test (tests/e2e/test_offline_queue_wake.py) that stands up a real DeviceBuilder with an esp8266 config, fakes the esphome CLI with an argv recorder, installs while the device is OFFLINE, and drives a fake mDNS announcement through the production monitor path — asserting the deferred compile-only job, the armed flag, the wake-triggered upload argv (upload <config> --device OTA), and the flag clearing; all waits are event-driven off the bus, no polling. 6e5c2b6 closes the flap re-entrancy window from the last review: the wake now creates the UPLOAD job synchronously (bus listeners don't yield, so a flap's second wake sees it in active_jobs) and only the enqueue is deferred — no tracking set, no optimistic flag games. The same commit removes force_local from the deferred compile so it rides the normal remote-eligibility scheduler (a 40-minute local build on an HA Green vs minutes on a build server defeats having the firmware ready at wake), which required moving the completion handling onto a JOB_COMPLETED bus listener — pool-dispatched remote compiles never pass through the lane runner's _execute_job, so the old hook placement would have silently never armed a remote deferred build.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

bdraco (approved):
Thanks @rwalker777

PR #1435, opened 2026-06-13
Surface board combobox for variant-less platforms in the editor

bdraco:

The structured editor stripped `board` from every platform component, so on the variant-less platforms (esp8266, nrf52, bk72xx, rtl87xx, ln882x) there was no way to pick the board; the rtl87xx/bk72xx sections rendered empty even when the YAML already had a `board:`. Those platforms have no `variant` to substitute for the board, so `board` is their required sole selector.

How they got there: `_DEPRECATED_FIELDS` was introduced in commit `28bf49f4` ("Add MAP entry type, skip deprecated board field, surface framework", an early direct commit with no associated PR), which added `board` for every target platform at once with the blanket rationale that the dashboard feeds esphome the `variant` only. That holds for esp32 and rp2040, which carry a `variant` (the schema enforces `has_at_least_one_key(board, variant)`); it was over-applied to the variant-less platforms, where `board` is the required sole selector and nothing replaces it. nrf52 was never added to the set, so it already showed `board`, just as a bare string with no catalog dropdown.

This carves the variant-less platforms out of the deprecation (esp32 and rp2040 keep it), and emits `board` with the platform's catalog board ids as `options` plus `allow_custom_value`, so the field renders as a pick-or-type combobox pre-filled with the current board; free-text is preserved since board is `cv.string_strict`. The five affected catalog bodies were regenerated via `script/sync_components.py`.

Note: `sync_components.py` now reads the committed `boards.index.json`. The board sync and the component sync are independent workflows, so these options are read as committed and can trail a board add/remove by one component-sync cycle; `allow_custom_value` covers the gap since any board id can still be typed.

- fixes <link to issue>

## Types of changes

## Frontend coordination

The frontend already renders this shape; a string entry with `options` + `allow_custom_value` routes to the existing combobox renderer (same path `unit_of_measurement` uses).

## Checklist


esphbot (approved):
No blocking issues found.

bdraco:
The combo box we render on the frontend is not good, and https://webawesome.com/docs/components/combobox/ is not open source. However we already built a combo box for the command palette so we can use that if we extract if first

esphbot (approved):
No blocking issues found.

PR #1449, opened 2026-06-14
Add per-platform UART combobox for logger hardware_uart

bdraco:

The logger `hardware_uart` field rendered as a free-text box, so users had to know the valid UART name for their chip. The valid set varies by chip variant (ESP32 offers UART0/1/2; ESP32-C3 swaps in USB_SERIAL_JTAG / USB_CDC) and comes from a custom `uart_selection` validator the schema bundle can't carry, which is why it was never a dropdown.

This introspects ESPHome's `UART_SELECTION_*` maps into a new `ConfigEntry.platform_options`, keyed the same way as `platform_defaults`. The component controller resolves it into `options` per request, chip variant first then base platform, exactly like it already resolves `platform_defaults` into `default_value`; combined with `allow_custom_value` the frontend renders the existing pick-or-type combobox, so no frontend change is needed. Resolution is now variant aware, which also fixes the ESP32-C3 default that previously fell back to the base esp32 `UART0`.

Only `definitions/components/logger.json` changed in the catalog; it was regenerated via `script/sync_components.py`, not hand-edited.

The variant-first resolution is global, not logger-only: it also activates variant `platform_defaults` keys the old platform-only lookup never reached. For example an ESP32-C3 board now resolves `remote_transmitter.rmt_symbols` to 48 instead of the base esp32 64; this is a latent fix and is pinned by a regression test.

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist


esphbot (changes requested):

bdraco:
Thanks, all three addressed in ea7e3332:

- **Important #1 (broadened scope):** confirmed and pinned. Added a regression test asserting an ESP32-C3 board resolves `remote_transmitter.rmt_symbols` to 48 (S2 stays 64), and called out the global variant-first behavior in the PR description.
- **Suggestion #1 (duplicated normalizer):** deduped rather than commented. The sync script now imports the runtime `_variant_to_key`, so build-time keys and the controller lookup are the same function; a test pins `sync._variant_to_key is controller._variant_to_key`.
- **Silent failure (logger import):** narrowed to `ImportError` with a warning; any other failure now propagates and fails the build instead of silently dropping the combobox.

esphbot (approved):
No blocking issues found.

bdraco:
Good call; done in f0c3d779. Renamed `_variant_to_key` to public `variant_to_key` and re-exported it from `controllers/components`, mirroring the `INTERNAL_COMPONENT_IDS` pattern; the sync script and tests import it from the package now. No behavior change.

PR #1459, opened 2026-06-14
Accept underscores in esphome.name so mDNS matches

### LINKED ISSUE #1453: [Bug] MAC-Adress and running ESPHome-Version not recognized
### What happened?

The ESPHome Device Builder cannot detect either the MAC address or the deployed ESPHome version for one of my ESPHome devices (it is the same device as in https://github.com/esphome/device-builder/issues/1414).

I can't figure out why right now—are there specific conditions that need to be met for this to work?

[hf-display-logs.txt](https://github.com/user-attachments/files/28930458/hf-display-logs.txt)
[hf-display-simplified.yaml](https://github.com/user-attachments/files/28930459/hf-display-simplified.yaml)

Image

Image

### Dashboard version

v0.1.0b116

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

A device whose `esphome.name` contains an underscore (e.g. `hf_display`) was never recognized: its MAC, version, config hash and IP stayed empty, even though `aioesphomeapi-discover` and `esphome logs` resolve it and the legacy dashboard matches it.

ESPHome allows underscores in `name:` (`esphome.const.ALLOWED_NAME_CHARS`) and broadcasts the name verbatim over mDNS, but our `_VALID_ESPHOME_NAME_RE` was `[a-z0-9-]+`, missing the underscore. So `_is_valid_esphome_name("hf_display")` was false and `load_device_from_storage` fell back to the filename stem, keying the device under the wrong name. The mDNS browse then looked up the verbatim `hf_display`, missed, and bailed, so nothing was applied. Our mDNS side was already correct (it preserves the underscore on purpose).

The fix adds the underscore to the regex to mirror ESPHome. Genuinely-wrong fields (a dotted package id, a friendly_name with spaces/uppercase) still fail, so the stem fallback they were guarding still works.

- fixes esphome/device-builder#1453

## Types of changes

## Frontend coordination

The drawer copy that surfaces an undiscovered device (device-builder-frontend#833) is a separate UX improvement and stays open.

## Checklist

### DIFF
--- a/esphome_device_builder/helpers/device_yaml/_loading.py
+++ b/esphome_device_builder/helpers/device_yaml/_loading.py
@@ -152,7 +152,7 @@ def load_device_from_storage(
     fallback_name = configuration_stem(filename)
     storage_name = storage.name if storage else None
     # Pick the first valid ESPHome slug from (yaml_name, storage_name).
-    # Real ESPHome ``esphome.name`` values are ``[a-z0-9-]+`` — a parsed
+    # Real ESPHome ``esphome.name`` values are ``[a-z0-9_-]+`` — a parsed
     # value with dots / spaces / uppercase is something else (a package
     # id like ``ratgdo.esphome``, a friendly_name leaked through, etc.).
     # ``device.name`` is the key the state monitor uses to match mDNS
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Good call, applied. The regex is now built from the source of truth so it can't desync again:

```python
_VALID_ESPHOME_NAME_RE = re.compile(rf"\A[{re.escape(const.ALLOWED_NAME_CHARS)}]+\Z")
```

`re.escape` keeps the trailing `-`/`_` class-safe. Added a `test_is_valid_esphome_name_covers_allowed_name_chars` invariant so a future hand-edit that drops a char fails. The reject-the-wrong-field heuristic still holds: ESPHome can't add `.`/space/uppercase to a name (DNS-invalid), so package ids and friendly_names stay rejected.

esphbot (approved):
No blocking issues found.

PR #1480, opened 2026-06-15
Persist the expert_mode preference the frontend already writes

bdraco:

The frontend's Expert Mode toggle writes `expert_mode` via `config/set_preferences`, but the backend `UserPreferences` model never had the field, so mashumaro dropped it on write and `get_preferences` never returned it; the toggle reset to off on every reload. Adds `expert_mode: bool` to the model so the existing get/set path persists it. Backend only; the frontend already reads and writes the field.

- N/A

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/models/preferences.py
+++ b/esphome_device_builder/models/preferences.py
@@ -46,6 +46,10 @@ class UserPreferences(DataClassORJSONMixin):
     navigator_visible: bool = True
     yaml_diff_button: bool = False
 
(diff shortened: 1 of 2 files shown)


bdraco:
folded into https://github.com/esphome/device-builder/pull/1445

final wire shape to be BEGINNER = “beginner” or EXPERT = “expert” and drop  UI = “ui”, / YAML = “yaml”

PR #1489, opened 2026-06-15
Backfill esp32 board variants so every board can be created

### LINKED ISSUE #1486: Shelly 1 Gen 4 board can't be created
Image

bdraco:

The Shelly 1 Gen4 board (and 15 other esp32 boards) could not be created. Their catalog entries carried only a PlatformIO `board:` id with no `variant:`, and the device-YAML generator emits `variant:` for esp32 but never `board:`, so the generated `esp32:` block had neither key; ESPHome rejects that (the esp32 schema needs at least one of board or variant) and create failed. The missing variant also made the board picker tag these as bare ESP32 instead of their real sub variant (ESP32-C6, etc.).

The variant is fully derivable from the PlatformIO board id, so `sync_boards.py` now backfills it from ESPHome's `BOARDS` table for any esp32 board missing it; the catalog is regenerated, which fixes both creation and the picker classification. The 16 affected board bodies plus the index pick up their variant.

Tests: a fast check that every board generates a platform block create accepts, and a slow `tests/e2e/slow/boards` sweep that runs real ESPHome validation on every board, each in its own forked worker since ESPHome accumulates module global state across validations that `CORE.reset()` cannot fully clear.

- fixes https://github.com/esphome/device-builder/issues/1486

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1495, opened 2026-06-16
Track latest beta esphome schema in the catalog syncs

bdraco:

New chip variants from an esphome beta did not show up in the dashboard, because the component catalog is built from the stable schema release; the nightly sync pinned the version to the installed stable esphome. This makes the nightly sync resolve the latest schema release including prereleases, then install that exact esphome so the live introspection lines up with the beta schema. The device-catalog sync now introspects against the latest prerelease too, and both jobs install with uv to match the rest of CI.

The schema repo already publishes a release per beta (b1/b2/b3 all exist and the bundles serve fine), and the latest one already carries the new variants; for example the `2026.6.0b3` `esp32` schema lists `ESP32H4`, `ESP32H21`, `ESP32S31` in its `variant` enum. We just were not consuming it; the dashboard was pinned to stable `2026.5.3`, whose variant set tops out earlier, which is why those chips did not show up. So this is the complete device-builder side fix; once the sync targets the latest beta, those variants flow into `esp32.json` automatically.

This PR only changes which schema the sync targets; it does not regenerate the catalog. The nightly job (or a manual workflow_dispatch with an empty version) opens that catalog PR separately once this lands.

The `include_prereleases` dispatch input is gone since beta tracking is the default now; a stable build is still reachable on manual dispatch by pinning an explicit `version`. The align step installs the exact resolved esphome and fails hard if that wheel is not on PyPI yet, so a schema/PyPI skew retries the next run rather than proposing a catalog built with mismatched introspection.

- N/A

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

bdraco:
Thanks for the review. Addressed in 38cae42:

**🟡 Align step fails if the beta esphome isn't on PyPI yet** — kept as a hard fail on purpose. Introspecting a beta schema with a stale esphome would regress the catalog (`introspect_component` quietly degrades to empty multi_conf/platform_defaults/refined_types for components the installed esphome can't load), and we don't want a degraded catalog proposed for merge. A red nightly that retries the next day is the better failure mode; added a comment saying so. The skew window is small in practice; for a recent beta the schema tag and the PyPI version match exactly (`esphome==2026.6.0b3` resolves cleanly).

**🟢 Unauthenticated GitHub API in the resolve step** — `_http_get` now attaches `Authorization: Bearer $GITHUB_TOKEN` when the env var is set, scoped to `api.github.com` only so the token is never sent to the schema CDN download. The resolve step passes `secrets.GITHUB_TOKEN`, lifting it off the 60 req/hr cap.

**🟢 Dropped `include_prereleases` dispatch input** — intended; beta tracking is the default now. A stable build is still reachable on manual dispatch by pinning an explicit `version`. Noted in the PR description.

PR #1496, opened 2026-06-16
Tag WiFi-capable RP2040 boards so the picker shows a WiFi chip

bdraco:

Most RP2040 boards have no WiFi; only a few (Pico W, Pico 2 W, and the Pimoroni/SparkFun/Waveshare W variants) do, but the board picker card showed only the `RP2040` platform pill, so users didn't know which board to pick for a WiFi build.

This adds a `WiFi` board tag and injects it in `_augment_rp2040_boards` from esphome's own per-board `wifi` flag, so the 10 WiFi-capable RP2040 boards get a WiFi chip on their card; esp32/esp8266/libretiny are universally WiFi, so they don't get the chip (it would be noise). The flag is the same data behind the wizard's wifi inference, so the chip and the generated config can't disagree.

- N/A

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#860

## Checklist


bdraco:
We still have a bug where the boards aren't appearing unless I search but its outside of scope here

PR #1499, opened 2026-06-16
Distinguish RP2040 and RP2350 by chip in the board catalog

bdraco:

ESPHome lumps RP2040 and RP2350 under one `rp2040` platform, so the board catalog had no way to tell the two chips apart; the picker showed a single combined filter and badged every board "RP2040". We should have modeled this like ESP32, which carries its chip variant on each board; this adds the equivalent for rp2040.

Carries ESPHome's per-board chip series onto every rp2040 entry as `esphome.mcu` ("rp2040" / "rp2350"), backfilled at sync time from `esphome.components.rp2040.boards.BOARDS` (the `mcu` field; not inferred from pin count, which is ambiguous). `boards/get_boards` gains an `mcu` filter so the frontend can split the one chip into separate RP2040 / RP2350 filters and badge each board's real chip. Only rp2040 entries gain the field; other platforms keep using `variant`.

The `mcu` field is generic: it can be reused to split the libretiny families (e.g. bk72xx chips) the same way if we want chip-level filters there in the future.

- Companion frontend PR: esphome/device-builder-frontend#861

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#861

## Checklist


esphbot (changes requested):

esphbot (approved):
No blocking issues found.

PR #1503, opened 2026-06-16
Break extends cycles and prime CORE in the catalog sync

bdraco:

Now that the catalog sync tracks the latest beta esphome (2026.6.0b3), the run surfaced two defects the stable schema never hit; both were silently swallowed, so the catalog degraded without failing the sync.

`lvgl` was dropped entirely with a `RecursionError`. Its widgets extend `lvgl.WIDGET_TYPES` (a widget can contain a widget list) and `lvgl.STYLE_SCHEMA`, so the converter followed the nested widget schema and re-expanded `WIDGET_TYPES` forever. The raw JSON is shallow, so this is a reference cycle, not deep data; a depth cap would expand the cycle and balloon the 1.6 MB schema. The fix threads a `_seen_refs` set through the convert recursion so an `extends` target already being expanded up the stack is not re-expanded, which bounds the nested widget list.

`usb_uart` shipped without introspection (`KeyError: 'target_platform'`). In beta it evaluates a platform-gated schema (`CORE.is_esp32`) at import time, and the sync imported it with a `CORE` whose `KEY_CORE` dict lacked a `target_platform` slot. The loader now gives `CORE` that slot (set to `None`) when first resolved, so the lookup returns False instead of raising.

Verified locally against 2026.6.0b3: the sync log no longer shows the `lvgl` RecursionError or the `usb_uart` load failure; `lvgl` builds bounded (depth 6, 194 top-level entries) and `usb_uart` carries its introspected fields.

Heads-up for the catalog refresh: with `lvgl` restored, its body file is large (~14 MB; b3's lvgl inlines the full style schema on every widget) versus the 341-byte stub stable produced. It is lazy-loaded per-id, but it is a real wheel-size bump worth a look when the b3 catalog PR regenerates.

- Follow-up to the beta-tracking work (#1495, #1500)

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/script/sync_components.py
+++ b/script/sync_components.py
@@ -2230,17 +2230,25 @@ def _promote_template_controls(component_id: str, entries: list[dict]) -> None:
             entry["advanced"] = False
 
 
-def _merge_extends_config_vars(schema_node: dict, schema_dir: Path) -> dict[str, Any]:
+def _merge_extends_config_vars(
+    schema_node: dict, schema_dir: Path, _seen_refs: frozenset[str] = frozenset()
+) -> dict[str, Any]:
     """
     Deep-merge a schema node's ``extends`` ancestry with its local config_vars.
 
     Per-field, not whole-field: a partial override like ``{"default": "x"}``
     keeps the base's inherited ``type`` / enum / docs. Values are usually field
     dicts; a malformed schema can carry a non-dict, hence ``dict[str, Any]``.
+
+    ``_seen_refs`` carries the ``extends`` targets already being expanded up the
+    convert stack; a ref in it is skipped to break self-referential cycles
+    (lvgl widgets extend ``lvgl.WIDGET_TYPES``, which contains a widget list).
     """
     config_vars = dict(schema_node.get("config_vars") or {})
     extended: dict[str, dict] = {}
     for ref in schema_node.get("extends") or []:
+        if ref in _seen_refs:
+            continue
         extended.update(_resolve_extends(ref, schema_dir))
     merged: dict[str, Any] = {}
     for key in {*extended, *config_vars}:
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Thanks both. Addressed in 926a658:

**Copilot — explicit `frozenset` (L2277):** `seen_refs = _seen_refs | frozenset(...)` now, so the type stays unambiguously immutable.

**Copilot — docstring format (L2275):** moved the `_convert_config_vars` summary onto the line after `"""` per CLAUDE.md.

**Copilot / Kōan — test global mutation:** the priming test now uses `monkeypatch` (replaces `CORE.data` with a copy and `setitem`s the loader-cache entries), so the forced re-resolution and CORE mutation unwind on teardown — no order-dependent leakage.

**Kōan #1 — over-pruning a base reused at multiple depths:** verified empirically against `2026.6.0b3` rather than relying on the top-level count. A widget's state-parts keep their inherited `STYLE_SCHEMA` fields — for `button`: `main` 131 fields, `checked`/`pressed`/`focused`/`disabled` 119 each. It's safe because widget schemas extend `None` (they don't accumulate `STYLE_SCHEMA` onto the path), and the parts re-extend `STYLE_SCHEMA` as independent siblings, so `_seen_refs` only ever carries the genuinely-cyclic `WIDGET_TYPES`. Agreed the path-wide ref prune is conservative for a hypothetical base reused down a non-cyclic path; no such case exists in today's catalog, so leaving the note here for a future schema.

**Kōan — `target_platform=None` / esp32-only fields:** `usb_uart`'s 28 config entries come from the schema bundle, not introspection, so the field set is complete regardless; the `None` platform only routes the import-time `CORE.is_esp32` validation-limit branch (it doesn't drop fields). Introspection only layers on `multi_conf` / `platform_defaults`.

esphbot (approved):
No blocking issues found.

PR #1505, opened 2026-06-16
Dispatch the catalog syncs on an ESPHome release

bdraco:

Adds a small workflow that esphome/version-notifier dispatches right after an ESPHome release, so the catalog refreshes immediately instead of waiting for the nightly. It only orchestrates; the existing `sync-component-catalog.yml` and `sync-device-catalog.yml` are untouched (#1495 already had them track the latest beta schema and esphome).

On dispatch it waits for the release's schema to publish (the schema build and this run fan out from the same event, so the bundle can lag a few minutes), then dispatches both sync workflows with the released version.

Companion trigger in esphome/version-notifier: esphome/version-notifier#19.

Note: a brand new chip variant can still need a manual `Esp32Variant` entry in `models/boards.py`, and bumping the bundled `esphome-device-builder==` pin in esphome's Dockerfile stays a separate manual step.

- N/A

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/.github/workflows/sync-on-esphome-release.yml
+++ b/.github/workflows/sync-on-esphome-release.yml
@@ -1,0 +1,76 @@
+name: Sync catalog on ESPHome release
+
+# Dispatched by esphome/version-notifier right after an ESPHome release. Waits
+# for the new schema to publish, then dispatches the existing component and
+# device/board catalog sync workflows so both refresh immediately instead of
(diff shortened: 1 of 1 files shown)


esphbot (changes requested):

esphbot (approved):
No blocking issues found.

PR #1518, opened 2026-06-17
Defer model mashumaro codegen to cut cold import

bdraco:

On a slow ARM target (HA Green) startup is dominated by importing our own models/controllers tree, which blocks the frontend for many seconds. Phase timing showed ~13.5s of import (`config` 9.5s + `builder` 4.0s) before the frontend is even served; the esphome package base is only 0.1s.

mashumaro generates `from_dict`/`to_dict` at class-definition (import) time. This adds a `DashboardModel` base that sets mashumaro `lazy_compilation`, switches the model dataclasses to it, and sets `lazy_compilation` on the shared catalog `Config`. Codegen moves to first use, so importing the models tree is about 2.4x faster locally (`config` 423ms to 148ms, `builder` 144ms to 83ms). Wire output is unchanged; the deferred first `from_dict`/`to_dict` compile costs under a millisecond per class.

The discriminated-union location variants are lazy like the rest. The one place that dispatched on them (`_LOCATION_DECODERS`) used to capture each variant's bound `from_dict`; under `lazy_compilation` that bound reference is a one-shot stub that recompiles the unpacker on every call, so the dispatch now maps each `kind` to its concrete type and resolves `from_dict` per call (compiled-once after first use). A regression test pins compile-once decode.

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist


esphbot (changes requested):

esphbot (approved):
No blocking issues found.

Issue #288, opened 2026-05-04
[Bug] Device builder is ignoring configuration from packages


randybb:
### What happened?

The Device Builder is using only partial configuration, not the final one merged with packages (in my case all devices are using a common package for ota, api, wifi,...), so it is showing incorrect information:
- OTA shows as "Not encrypted";
- "Platform" is empty;
- "Loaded Integrations" are empty.

### Dashboard version

dev

### How did you install it?

From source (script/setup)

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

randybb:
Have BLE beacons without configured wifi/ota/api and these are showing only 3 options for update, the 4th for manual upload is missing andthe first one - via wifi/network - would never work.
<img width="469" height="363" alt="Image" src="https://github.com/user-attachments/assets/76da880c-3630-4a5a-ab62-8f2aa4b4f941" />

example config: https://github.com/randybb/esphome-configs/blob/main/mcu-vehicle-honda-pcx125.yaml

PR #1519, opened 2026-06-17
Skip external package git fetch during the device scan

bdraco:

The startup device scan loads every YAML for metadata. For configs with a `packages:` block it runs esphome's package resolution, which by default refreshes each external repo with a `git fetch` (package default `refresh: "1d"`). On a fleet that is a wall of `[esphome.git] Updating https://github.com/...` lines on every startup and dominates the post-serve `controllers` phase (~10s on an 86-device install).

The legacy ESPHome dashboard never did this: `DashboardEntry.load_from_disk` reads `StorageJSON.load()` only, so its scan never parsed YAML, never resolved packages, and did no git work at all. Device Builder added in-process package resolution so metadata behind `packages:` is visible (the api/wifi/platform blocks for flag detection, #288), and that inadvertently pulled esphome's git refresh into the scan.

The long-lived process never needs an in-process fetch: metadata reads cached checkouts, and real builds run in esphome CLI subprocesses with their own `CORE`. So this sets `CORE.skip_external_update` once at dashboard startup (`DashboardSettings.parse_args`), where the dashboard already initializes `CORE`. esphome's git layer then treats an already-cloned package as a local no-op, so the scan merges from cached content with no fetch. Setting it once is also why this avoids toggling the process-global per resolve from shared executor threads (a racy pattern an earlier draft used); nothing in-process ever wants a real fetch, so the flag is just on for the process. A first-time clone still happens once (the content has to exist to merge), and real compiles are unchanged.

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/config/settings.py
+++ b/esphome_device_builder/controllers/config/settings.py
@@ -213,6 +213,11 @@ class DashboardSettings:
             host.strip().lower() for host in raw_trusted.split(",") if host.strip()
         ]
         CORE.config_path = self.config_dir / _DASHBOARD_SENTINEL_FILE
+        # The long-lived process never fetches external packages in-process:
+        # metadata resolution reads cached checkouts, and real builds run in
+        # esphome CLI subprocesses with their own CORE. Set once here so the
+        # device-scan package merge doesn't git-fetch per repo on every startup.
+        CORE.skip_external_update = True
 
     def rel_path(self, *parts: str) -> Path:
         """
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1523, opened 2026-06-17
Check the exact release tag, not a prefix match

bdraco:

The release workflow's "Reject already-existing tag" check used GitHub's plural refs endpoint, `git/refs/tags/<version>`, which does **prefix** matching. For a stable `X.Y.Z` whose `X.Y.ZbN` betas already exist, that query returns HTTP 200 with the beta refs, and the step reads any 2xx as "tag exists" and refuses to release.

This blocks the normal flow, since every stable release follows its prereleases. The 1.0.2 release hit exactly this: `1.0.2b1`..`1.0.2b4` existed, so the check false-positived on `1.0.2` even though no `1.0.2` tag was ever created. As a workaround I skipped 1.0.2 and released 1.0.3 (no `1.0.3bN` betas, so the prefix query 404'd); 1.0.2 was never tagged or published anywhere, so there's no gap.

The fix switches to the singular `git/ref/tags/<version>` endpoint, which matches the tag exactly and 404s when absent. The surrounding 404-capture logic is unchanged. Verified against the live repo: singular `git/ref/tags/1.0.2` returns 404 (would proceed) and `git/ref/tags/1.0.2b4` returns the object (would still reject an existing tag); beta releases are unaffected.

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,12 +63,16 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           set -euo pipefail
+          # Use the singular ``git/ref/tags`` endpoint: it matches the tag
+          # exactly and 404s when absent. The plural ``git/refs/tags`` does
+          # PREFIX matching, so an ``X.Y.Z`` query returns 200 with the
+          # ``X.Y.ZbN`` prerelease tags and false-positives "already exists".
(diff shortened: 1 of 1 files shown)


esphbot (approved):
No blocking issues found.

PR #1526, opened 2026-06-17
Strip comment after a quoted friendly_name or substitution value

### LINKED ISSUE #1525: [Bug] Comment recognized as part of the friendly_name
### What happened?

Reported on [discord](https://discord.com/channels/429907082951524364/1478125795355529326/1516734468264562759)

If a comment follows the `friendly_name`, it is displayed as part of the `friendly_name` in the GUI.

Image

Image

Same with substitution:

Image

Image

### Dashboard version

v1.0.3

### How did you install it?

Desktop App

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The lightweight `esphome:` meta parser stripped inline comments with a quote-unaware regex that only ran when the value did not start with a quote; a quoted value followed by a comment kept its quotes and the comment, so `friendly_name: "Test #1"  # c` showed up in the device title as `"Test #1" # c`. The same happened for a quoted substitution value feeding `friendly_name`.

The editor form field already renders these correctly because the frontend parses the scalar itself in `util/yaml-scalar.ts`; the title leaks because it reads the backend-supplied `friendly_name`. This ports the frontend's two primitives (`splitInlineComment`, `stripQuotes`) to the backend so both sides parse scalars identically: a quote-aware, escape-aware walk that opens a comment only outside quotes and whitespace-preceded, keeping a mid-scalar `Room#2` literal.

- fixes https://github.com/esphome/device-builder/issues/1525

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/helpers/device_yaml/_parsing.py
+++ b/esphome_device_builder/helpers/device_yaml/_parsing.py
@@ -8,6 +8,8 @@ from typing import NamedTuple
 from esphome import const
 from esphome.const import CONF_PACKAGES
 
+from ..yaml import _split_value_and_comment, _strip_yaml_quotes
+
 
 class EsphomeMeta(NamedTuple):
     """The user-facing ``esphome:`` meta fields; ``None`` for any the YAML omits."""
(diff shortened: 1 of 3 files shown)


esphbot (changes requested):

bdraco:
Addressed both reviews in `608aa565`:

**Kōan warning #1 (single-quote `''` escape):** good catch, that was the same title-vs-form divergence for single-quoted input. Fixed locally in `_parse_inline_value` (not in the shared `_strip_yaml_quotes`, since the rename gate / substitution-ref parser / `is_plain_literal_scalar` rely on its current behavior): the matched single-quoted case now unwinds `''` to `'`, so `'Bob''s Room'` resolves to `Bob's Room`. Added unit and parse-meta tests; the "parse identically" claim now holds.

**Copilot (issue-number refs):** dropped the `(#1525)` references from the test docstrings and the inline comment per CLAUDE.md.

esphbot (approved):
No blocking issues found.

bdraco:
Re Kōan suggestion #1 (double-quoted `\"`): confirmed the two sides parse identically, so leaving `\"` literal is correct.

The frontend's `stripQuotes` double-quote branch is just `s.slice(1, -1)` with no unescaping; `unescapeYamlDoubleQuoted` is only used by `parseFlowList` and `secret-eligibility`, not the `parseScalar` form-field path. So `"Say \"hi\""` parses to `Say \"hi\"` on both the backend title and the frontend form; unwinding `\"` here would *introduce* the divergence rather than remove it.

Pinned it with an assertion that `_parse_inline_value('"Say \"hi\""')` stays `Say \"hi\"`, noting it mirrors the frontend and matches the round-trip of the backend's own `_quote('Say "hi"')`. 163 tests pass.

PR #1563, opened 2026-06-18
Serve top-level frontend bundles without auth

### LINKED ISSUE #1560: [Bug] Device Builder inaccessible when USERNAME and PASSWORD are used
### What happened?

1. Upgraded docker container to 2026.6.0 which switched to using this new Device Builder instead of the old in-tree one.
2. Noticed that it wasn't honoring the old USERNAME and PASSWORD env vars anymore.  It would let anyone in.
3. Found and changed to the new ESPHOME_USERNAME and ESPHOME_PASSWORD env vars.  Noticed it properly prompted for login.
4. Did a forced browser refresh and got a blank screen.  Did another hard refresh and cleared cache, and still got a blank screen.
5. Opened developer tools and noticed that the app.[checksum].js file and favicon.ico are returning a 401.
6. Removed the ESPHOME_USERNAME and ESPHOME_PASSWORD env vars and verified that the device builder works again.

Seems like at least the app js and favicon need to be allowed through even if unauthenticated.

### Dashboard version

2026.6.0

### How did you install it?

### Where is it running?

Docker container

### Relevant logs / errors

```shell

```

### Anything else?

Image

bdraco:

When HTTP auth is enabled (ESPHOME_USERNAME + ESPHOME_PASSWORD), the dashboard shows a blank screen. The SPA shell loads at `/`, but the frontend's entry script and lazy chunks return 401, so the app never boots.

`index.html` loads its bundle with a flat `src`, so `app.<hash>.js` and the numbered chunks resolve to the deploy root (`/app.<hash>.js`), not under `/assets/`; the auth allowlist only covered `/assets/`, so every top-level bundle 401'd. The original intent (allowlist comment: "Frontend assets are public so the login page can load before login") was only half implemented.

This allows GET/HEAD requests to top-level content-hashed filenames through `auth_middleware`. The hash segment is what makes it safe; no API or legacy route is content addressed, so `/devices`, `/json-config`, `/compile`, `/upload`, `/api/firmware/download` stay gated. The shared `HASHED_FILENAME_RE` is the same pattern the frontend serving uses for its immutable cache headers.

Verified against a real server with a password set: `/app.<hash>.js` now returns 200 without credentials while `/devices` and `/json-config` still return 401, and a nested `/device/app.<hash>.js` stays gated.

- fixes https://github.com/esphome/device-builder/issues/1560

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/device_builder.py
+++ b/esphome_device_builder/device_builder.py
@@ -11,7 +11,6 @@ import asyncio
 import contextlib
 import html
 import logging
-import re
 from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor
 from functools import lru_cache
@@ -43,7 +42,7 @@ from .controllers.remote_build import OffloaderController, ReceiverController
 from .controllers.version_history import VersionHistoryController
 from .helpers.api import CommandHandler, collect_api_commands
 from .helpers.async_ import create_eager_task
-from .helpers.auth import auth_middleware
+from .helpers.auth import HASHED_FILENAME_RE, auth_middleware
 from .helpers.dashboard_advertise import DashboardAdvertiser
 from .helpers.dashboard_identity import get_or_create_identity as get_or_create_dashboard_identity
 from .helpers.event_bus import Event, EventBus, StreamControls, stream_events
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Addressed the review nits in c4fab106: trimmed the new test docstrings and the `HASHED_FILENAME_RE` comment to the terse repo style and dropped the issue cross-refs from code; the HEAD test now asserts 200 instead of just not-401. The security reasoning stays in the comment since it's the load-bearing invariant for why the bypass is safe.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1565, opened 2026-06-18
Stop exposing the no-auth HA Ingress site beyond the supervisor

bdraco:

The trusted HA Ingress site skips auth because the supervisor authenticates the request upstream, and it was binding `0.0.0.0`. The ESPHome add-on runs host network for mDNS, so `0.0.0.0` also bound the host's LAN interface; any device on the network could reach `http://<ha-ip>:<ingress_port>/` and get the full dashboard with no credentials. This has been present since the ingress site was introduced (#25); it only affects the host-network HA add-on, not the password-gated public port (6052).

Two changes, matching what the legacy add-on's nginx did:

- Bind the ingress site to loopback plus the supervisor gateway (`127.0.0.1` + `172.30.32.1`) instead of all interfaces. Loopback serves HA core's host-network ESPHome integration (it connects to `127.0.0.1:<ingress_port>` with no credentials by design); the gateway serves the supervisor's ingress proxy. The LAN interface is no longer bound; an explicit `--ingress-host` still overrides the bind.
- Add `ingress_peer_guard`: the trusted site returns 403 to any TCP peer other than loopback or the supervisor (`172.30.32.2`), so another hassio bridge add-on reaching the gateway cannot use it either. This mirrors the legacy nginx `allow 127.0.0.1; allow 172.30.32.2; deny all`.

The gateway and supervisor addresses are the supervisor's own fixed `172.30.32.0/23` constants (it reaches host-network add-ons at the gateway and connects from `.2`), so we bind and filter exactly where it talks to us. The peer guard still applies regardless of `--ingress-host`; the help text, the `ingress_bind_hosts` docstring, and a debug log of any rejected peer make that diagnosable.

- Found while investigating the auth behavior reported in #1560; reproduced directly against the add-on.

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

jesserockz (approved):

Issue #1567, opened 2026-06-19
[Bug] Cannot create any device without Wi-Fi secrets


smaeul:
### What happened?

I am trying to use ESPHome with a [WT32-ETH01 EVO](https://en.wireless-tag.com/product-item-59.html) board (ESP32-C3 + Ethernet). I am not using Wi-Fi. I want to start with a new blank configuration and fill it out in the editor.

1. When prompted to set up Wi-Fi credentials, choose "I don't use Wi-Fi — skip this step" (in other words, secrets.yaml is empty).
2. Attempt to create an empty configuration. This fails with the error below.
3. Attempt to import an empty YAML file. This also fails with the error below.
4. Attempt to use the wizard to create a device based on the "Generic ESP32-C3 Board". The wizard will not allow you continue without supplying Wi-Fi credentials.

### Dashboard version

2026.6.0

### How did you install it?

ESPHome container (preview toggle)

### Where is it running?

Docker container

### Relevant logs / errors

```shell
[SENDING] 
Object { command: "devices/create", message_id: "21", args: {…} }
​
args: Object { name: "dummy", board_id: "", config_type: "empty" }
​
command: "devices/create"
​
message_id: "21"
​
<prototype>: Object { … }
app.86aced6ff874c2ce.js:2178:26885

[RECEIVED] 
Object { message_id: "21", error_code: "internal_error", details: `Can't create — config doesn't validate: Secret 'wifi_ssid' not defined\n  in "/config/dummy.yaml", line 19, column 9. Please report this with a redacted snippet of just the esphome: / substitutions: blocks (strip Wi-Fi credentials, API keys, and static IPs) so the dashboard generator can be fixed.` }
​
details: `Can't create — config doesn't validate: Secret 'wifi_ssid' not defined\n  in "/config/dummy.yaml", line 19, column 9. Please report this with a redacted snippet of just the esphome: / substitutions: blocks (strip Wi-Fi credentials, API keys, and static IPs) so the dashboard generator can be fixed.`
​
error_code: "internal_error"
​
message_id: "21"
​
<prototype>: Object { … }
app.86aced6ff874c2ce.js:2178:24285
```

bdraco:
Currently the device builder does not have a mapping of ethernet pins per board so it doesn't support creating ethernet per board. 

It all has to be done manually in yaml or added in the ui field by field

smaeul:
[...]

PR #1604, opened 2026-06-19
Add onboard ethernet for ESP32-PoE, wESP32, Ethernet-Kit, and more

bdraco:

Adds onboard ethernet to five more common boards from the ESPHome ethernet docs that were missing: Olimex ESP32-PoE (the non-isolated sibling of the PoE-ISO we already ship), Silicognition wESP32, Espressif ESP32-Ethernet-Kit, M5Stack PoESP32, and the WT32-ETH01-EVO.

Pin data comes from the documented configs at esphome.io/components/ethernet; each generated config passes `esphome config`. The EVO is the ESP32-C3 board from the original issue #1567, so this closes the loop on that one. This also brings the RTL8201 and DM9051 PHYs into the catalog.

- builds on #1591

## Types of changes

## Frontend coordination

## Checklist


esphbot (changes requested):

bdraco:
Good catch, fixed. `esp32-poe` and `wesp32` already had auto-generated 34-pin bodies on main (both are in ESPHome's esp32 board list), and my partial manifests dropped them to 4 pins; there is no backfill for a manifested esp32 board, so that was a real pin-picker regression.

Restored the full 34-pin GPIO catalog for both (sourced from their prior generated bodies), with the ethernet GPIOs marked `available: false` / `occupied_by: Ethernet …` and the rest keeping their ADC/touch/strapping/feature metadata. The regenerated bodies are back to 34 pins, validator passes, and both still pass `esphome config`. The other three boards had no prior body, so no change needed there.

esphbot (approved):
No blocking issues found.

Issue #1529, opened 2026-06-14
Flashing ESP8266 over web serial fails


Spawnkiller1:
### The problem

I wanted to flash an ESP8266 on Windows using the beta backend; the flashing process works, but the ESP becomes unresponsive afterwards. However, if I download the file and flash it via web.esphome.io or the classic backend, it works fine. I am using the standard configuration yaml, and the same issue occurs with the Device Builder beta in Home Assistant.

### Which version of ESPHome Device Builder has the issue?

ESPHome Device Builder v0.1.0b116 ESPHome 2026.5.3

### What platform are you using?

Windows

### Anything in the logs that might be useful for us?

```txt

```

TheDelta:
Holy shit, good timing... - I'm facing the same issue and it almost drove me insane.
Was this your first flash or did you made multiple?
Mine was working fine for the first time but the flashes afterwards never worked and it kind of died - no logs no nothing.

Comparing both, it seems like the clearing is not done on esphome-desktop, because it goes straight to flashing (web version takes twice as long)

PR #1605, opened 2026-06-19
Reset the flashed chip into the app, not the stub bootloader

bdraco:

The dev flasher flashed correctly but the device **never booted into the app** afterwards, sitting in the ROM (reported on an ESP32-S3 over USB-Serial-JTAG: the log showed `Hard resetting via RTS pin...` then only the `ESP-ROM:esp32s3-...` banner, no firmware).

Root cause: `esploader.after()` resolves to esptool-js's `HardReset`, which only calls `setRTS(false)`. It doesn't pulse DTR/EN, so the chip stays in the stub bootloader and the just-flashed firmware never runs. This is the same trap `esphome/device-builder-frontend` already documents (#1529 there).

Fix: replace `esploader.after()` with the proven frontend strategy in a new `flasher/src/reset.ts`:
- **RTC-watchdog reset** for ESP32-S2/S3/C2/C3 (releases the boot-strap pin, then arms the WDT) — most reliable, works through UART bridges and native USB-Serial-JTAG alike.
- **`UsbJtagSerialReset`** for other native-USB (VID 0x303A) chips.
- **Classic EN pulse with GPIO0 released** for UART-bridge ESP32/ESP8266.

Dev/test flasher only; the postMessage hand-off contract is unchanged.

- Part of esphome/backlog#151

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/flasher/src/main.ts
+++ b/flasher/src/main.ts
@@ -1,5 +1,6 @@
 import { ESPLoader, Transport } from "esptool-js";
 import { validateEspImage } from "./image-magic";
+import { hardResetChip } from "./reset";
 import type {
   FirmwareMessage,
   OutboundMessage,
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #1609, opened 2026-06-19
Harden the dev flasher's postMessage receiver

### LINKED ISSUE #1607: Harden the dev flasher's postMessage receiver (parity with esphome/dashboard#923)
The web.esphome.io postMessage receiver review (esphome/dashboard#923) surfaced three receiver-side hardening items that the dev flasher (`flasher/src/main.ts`) shares, since both implement the same postMessage contract. The device-builder-frontend side is the sender (`src/util/usb-flasher.ts`) and is unaffected; these are receiver concerns.

1. **`post()` can throw on a malformed origin** (`flasher/src/main.ts:117-120`). `targetOrigin` comes from the hash `origin` param; a malformed value (e.g. `origin=null`) makes `postMessage` throw, which breaks the repeating ready interval. Wrap in try/catch and fall back to `"*"`.

2. **Malformed payload does not stop the ready retry** (`flasher/src/main.ts:169-177`). On a malformed payload the handler sets the error state and returns before `stopReadyRetry()`, so the opener keeps receiving `ready` frames for up to 10s. Call `stopReadyRetry()` on the failure path.

3. **`firmware.erase ?? true` truthy coercion** (`flasher/src/main.ts:510`). A non-boolean `erase` (e.g. the string `"false"`) is treated as truthy. Coerce explicitly (e.g. `firmware.erase !== false`).

Low priority: the dev flasher is the throwaway test page; web.esphome.io (now hardened in #923) is the production target. Worth doing for parity while the dev flasher is still the active test target.

bdraco:

Three receiver-side hardening fixes for the dev flasher's postMessage handler (`flasher/src/main.ts`), for parity with the web.esphome.io receiver review (esphome/dashboard#923) that shares the same contract. The device-builder-frontend sender (`src/util/usb-flasher.ts`) is unaffected; these are receiver concerns.

- **`post()` no longer throws on a malformed origin.** `targetOrigin` comes from the hash `origin` param, so a value like `origin=null` becomes `"null"` and the first synchronous `post()` threw before the ready-retry interval was set up, wedging the handshake. `post()` now wraps `postMessage` in try/catch, falls back to `"*"`, and pins `targetOrigin = "*"` so subsequent frames keep flowing.
- **A malformed payload stops the ready re-announce.** The `isFlashParts` failure branch set the error state and returned before `stopReadyRetry()`, so the opener kept receiving `ready` frames for up to 10s even though it had clearly attached and sent. It now calls `stopReadyRetry()` on that path, mirroring the accepted path.
- **`erase` is coerced explicitly.** `firmware.erase ?? true` only defaults on null/undefined, so a non-boolean `erase` (e.g. the string `"false"`) read as truthy. Changed to `firmware.erase !== false`, so erase stays the safe default unless the opener explicitly sent boolean `false`.

The puppeteer harness (`flasher/test/handshake.test.mjs`) gains assertions for the first two: a malformed `origin=null` still yields a `ready` frame, and a malformed payload stops the `ready` re-announce. The erase path only runs at flash time against real hardware, so it stays code-only (consistent with the harness's existing scope). I confirmed both new assertions go red when their fix is reverted.

- Fixes #1607

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/flasher/src/main.ts
+++ b/flasher/src/main.ts
@@ -117,7 +117,14 @@ const terminal = {
 function post(msg: OutboundMessage): void {
   // targetOrigin narrows from '*' to the opener's real origin once known; see
   // its declaration. Outbound frames carry no nonce (see protocol.ts).
-  opener?.postMessage(msg, targetOrigin);
+  try {
+    opener?.postMessage(msg, targetOrigin);
+  } catch {
+    // A malformed origin= hash param (e.g. origin=null) makes postMessage throw,
+    // which would wedge the ready handshake before the retry interval is set up.
+    targetOrigin = "*";
+    opener?.postMessage(msg, "*");
+  }
 }
 
 function setState(state: FlashState, detail: string): void {
@@ -167,6 +174,9 @@ window.addEventListener("message", (ev: MessageEvent) => {
   if (!data || data.type !== "esphome-web-flash:firmware") return;
   if (data.nonce !== nonce) return;
   if (!isFlashParts(data.parts)) {
+    // The opener has attached and sent — stop re-announcing ready even though
+    // the payload is unusable, mirroring the accepted path below.
+    stopReadyRetry();
     setState("error", "Received a malformed firmware payload.");
     return;
   }
@@ -504,7 +514,7 @@ installBtn.addEventListener("click", async () => {
       data: new Uint8Array(p.data),
       address: p.address,
     }));
-    await runFlash(files, firmware.erase ?? true);
+    await runFlash(files, firmware.erase !== false);
     return;
   }
   const file = fileInput.files?.[0];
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Thanks Kōan — addressed the broad-catch finding in 70c1250: `post()` now logs the caught error before falling back (so an unrelated failure isn't masked) and the fallback `postMessage` is wrapped too. I kept the `'*'` fallback rather than narrowing to `SyntaxError` because outbound frames carry no nonce, so the broader audience leaks nothing, and the in-browser error type for a bad origin didn't reliably match `SyntaxError instanceof DOMException` (the malformed-origin test went red with the narrowed check). The same fix is applied to the web.esphome.io receiver in esphome/dashboard#923.

esphbot (approved):
No blocking issues found.

PR #1612, opened 2026-06-19
Dedup generated boards that duplicate a curated manifest by name

bdraco:

The board picker showed two cards for some boards; e.g. a "WIZnet W6300-EVB-Pico2" twin where one was the curated entry (image, Dev-kit tag, onboard ethernet preset) and the other a bare generated entry with no ethernet. Picking the bare one created a device with no ethernet to auto-pull, which is what got reported.

Root cause is in the catalog generator (`sync_boards.py`): it emits a canonical entry for every ESPHome board key no manifest covers, deduping only on catalog `id`. The curated WIZnet manifests use `id: wiznet-w6300-evb-pico2` while their `esphome.board` is `wiznet_6300_evb_pico2`, so the generated twin (id == the board key) slipped past id-dedup and listed a second time under the same display name.

The fix extends each generator's id-dedup to also skip a board whose `(platform, display-name)` already exists; curated manifests load before any generated board is appended, so the curated entry always wins and only generation is ever suppressed. It can't dedup on `esphome.board` instead, because dozens of vendor products legitimately share a generic chip key (esp32dev, esp01_1m) with distinct names and must keep their own entries. A catalog sweep found the only `(platform, name)` collisions were 5 curated/generated twins (WIZnet x3, Freenove alias, Sonoff Basic); all are now collapsed to the curated entry, and a regression test pins that no two entries share a `(platform, name)`.

- 

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

bdraco:
Both suggestions applied:

1. Dropped the line-broken board-id example in the rp2040 docstring; the `(platform, display-name)` rule is already stated, so it reads clean now.
2. Added `_name_already_listed(...)` which emits a `_LOGGER.debug` when a generation is skipped on the name key (not the id key), mirroring the nrf52 warning pattern; wired into all five generators. Verified it fires for exactly the 5 known twins (WIZnet x3, Freenove alias, Sonoff Basic) and nothing else, so a future upstream board shadowed by a name collision surfaces in the sync logs instead of vanishing.

Logic-only change; regenerated artefacts are byte-identical, board tests green.

esphbot (approved):
No blocking issues found.

PR #1618, opened 2026-06-20
Add onboard W5500 ethernet to Waveshare ESP32-S3-ETH-8DI-8DO

bdraco:

The Waveshare ESP32-S3-ETH-8DI-8DO board has an onboard W5500 ethernet controller, but creating a device on it produced a bare config with no ethernet. Its sibling `waveshare_esp32_s3_eth_8di_8ro` (same product, relay outputs instead of digital outputs) already carries the full W5500 config in the catalog, so this just lands the same config on the 8DO variant; it now auto-pulls an `ethernet:` block on create like every other onboard-ethernet board.

I audited all esp32 boards in the catalog and this was the only genuine onboard-ethernet board missing the featured component (two ethernet-named stubs, yb_esp32s3_eth and cnrs_aw2eth, have no manifest and no in-repo config, so they're out of scope).

The 8DO manifest was imported from devices.esphome.io, and its upstream config has the `ethernet:` block commented out, so `sync_esphome_devices.py` skips it and a future re-sync would wipe a hand-added block. To keep the fix durable I dropped the `source:` block, converting it to a hand-curated board the sync leaves alone; this is the established pattern (see the `_extract_ethernet` docstring and `gl_inet_gl_s10_v2`). Pins GPIO12-16 are marked occupied (INT/MOSI/MISO/CLK/CS), mirroring the 8RO.

Verified: regenerated catalog, validator clean, the new W5500 config compiles to an esp32-s3 firmware image (esp-idf) end to end, and the sync skips the now-curated manifest.

- 

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1602, opened 2026-06-19
Add onboard ethernet for WIZnet W5500/W6300 and Waveshare ESP32-S3-ETH

bdraco:
# What does this implement/fix?

Follow-up to #1600, adds three more onboard ethernet boards: WIZnet W5500-EVB-Pico, WIZnet W6300-EVB-Pico2, and Waveshare ESP32-S3-ETH. Pins come from tested configs in my esphome checkouts, so the values are real; each generated config passes `esphome config`.

This brings two firsts to the catalog: the W6300 PHY, and the RP2350 platform (the W6300-EVB-Pico2 is a Pico 2 board on ESPHome's rp2040 platform).

I left the Waveshare ESP32-S3-ETH 8DI variants out for now; their input pins overlap the plain board's W5500 SPI pins, so the 8DI ethernet pinout is different and I do not have a tested config to source it.

**Related issue or feature (if applicable):**

- builds on #1600

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [x] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Frontend coordination

- [x] No frontend change needed

## Checklist

- [x] The code change is tested and works locally.
- [x] Pre-commit hooks pass (`ruff`, `codespell`, yaml/json/python checks).
- [x] Tests have been added or updated under `tests/` where applicable.
- [x] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (regenerate via `script/sync_components.py` if a sync is needed).
- [ ] Architecture-level changes are reflected in `docs/ARCHITECTURE.md` and/or `docs/API.md`.

PR #1621, opened 2026-06-20
Fix swapped W5500 clk/miso pins on Waveshare ESP32-S3-ETH

bdraco:

A device generated from the **Waveshare ESP32-S3-ETH** board failed to bring up its onboard W5500 at boot:

```
w5500.mac: W5500 version mismatched, expected 0x04, got 0x00
emac_w5500_init(883): verify chip ID failed
```

The board manifest had `clk_pin: GPIO12` and `miso_pin: GPIO13`, but Waveshare's pinout is SCLK GPIO13 and MISO GPIO12. With MISO listening on the SCLK line the chip-ID read came back all-zeros, so ethernet never initialized. This swaps the two pins (and the matching `occupied_by` pin annotations) and regenerates the catalog. MOSI/CS/INT/RST and 25MHz were already correct.

The transposition slipped through because both pins are valid GPIOs, so `esphome config` accepted it; only flashing real hardware surfaced it. Verified against Waveshare's ETH_DHCP pin table and confirmed on hardware (W5500 now initializes, link plus DHCP).

Reported via hardware flash; introduced in #1602.

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1625, opened 2026-06-20
Restore legacy USERNAME and PASSWORD env vars as a deprecated fallback

bdraco:

Restores backward compatibility for the deprecated bare `USERNAME` and `PASSWORD` env vars. The `ESPHOME_*` rename in #265 dropped the old names, so a dashboard that was protected with `USERNAME` and `PASSWORD` lost its authentication on upgrade and started open to the network. The bare pair is accepted again as a deprecated fallback so previously protected instances stay protected; a loud deprecation warning at startup tells operators to rename them to `ESPHOME_USERNAME` and `ESPHOME_PASSWORD`. The fallback is gated on `PASSWORD` being set and is only adopted as a pair, so the OS provided `USERNAME` is never read on its own and the collision footgun that motivated #265 stays closed; a lone bare `PASSWORD` still fails loud as a credential mismatch rather than starting unauthenticated.

- closes GHSA-rrxg-g2pf-6hh4

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -70,6 +70,8 @@ On connect, the server sends a [`ServerInfoMessage`](../esphome_device_builder/m
 
 When the dashboard is started with `--username`/`--password` (or `$ESPHOME_USERNAME`/`$ESPHOME_PASSWORD` env vars), every WebSocket connection on the public port must authenticate before any other command will be accepted.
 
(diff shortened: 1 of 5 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1628, opened 2026-06-20
Commit version history only on real YAML edits

bdraco:

The version history catch-all listened to `DEVICE_UPDATED`, which also fires on every mDNS runtime tick (version, config_hash, ip, mac, api_encryption) and on every post-flash scanner reload. None of those change the YAML, but each one queued a debounced git commit; the commit was a content no-op yet still grabbed the repo's `index.lock`, so a quiet device kept triggering pointless git work on every announce.

This adds a `DEVICE_YAML_UPDATED` event that fires only when the scanner sees a device's YAML change on disk, and splits the forced metadata re-read into `ScanChange.RELOADED`. Clients still get `DEVICE_UPDATED` for both cases; only a real edit fires `DEVICE_YAML_UPDATED`, which is now what version history commits on. The new event is internal, so it stays off the client broadcast.

- N/A; reported from production logs.

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/_device_scanner.py
+++ b/esphome_device_builder/controllers/_device_scanner.py
@@ -64,9 +64,9 @@ class ScanChange(StrEnum):
     ADDED = "added"
     UPDATED = "updated"
     REMOVED = "removed"
-    # A forced re-read whose YAML bytes did not change (binary mtime,
+    # A forced re-read with no YAML cache-key change (binary mtime,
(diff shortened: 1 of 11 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

thepointer1982 (commented):
G

PR #1629, opened 2026-06-20
Await peer-link client teardown before rebind respawn

bdraco:

A paired build server that restarts under a changed mDNS hostname could stay DISCONNECTED until the dashboard process itself was restarted.

On a restart the receiver re-announced under a different `.local` name (same machine, same IP), which fired an endpoint rebind. The rebind cancelled the old peer-link client and immediately spawned a new one without awaiting the old task; the old client's in-flight connect to the prior endpoint then completed a few milliseconds later, and since every peer-link client shares the offloader's single `dashboard_id`, the receiver superseded the freshly-spawned client and orphaned it for good (the orphan flag is one-shot).

The fix awaits the old client's teardown before spawning the replacement, so the old connection is fully gone (its terminate sent, the receiver's `dashboard_id` slot freed) before the new client connects; the new client is then the only registrant and stays connected.

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/remote_build/_shared.py
+++ b/esphome_device_builder/controllers/remote_build/_shared.py
@@ -12,6 +12,7 @@ Exposes:
 from __future__ import annotations
 
 import asyncio
+import logging
 from collections.abc import Callable, Iterable
 from contextlib import ExitStack
 from typing import TYPE_CHECKING, Any
@@ -23,12 +24,16 @@ if TYPE_CHECKING:
     from ...device_builder import DeviceBuilder
     from ...helpers.event_bus import Event, EventType
 
+_LOGGER = logging.getLogger(__name__)
 
-async def drain_tasks(tasks: Iterable[asyncio.Task[Any]]) -> None:
+
+async def drain_tasks(tasks: Iterable[asyncio.Task[Any]], *, log_exceptions: bool = False) -> None:
     """Cancel and await every task in *tasks*, swallowing exceptions.
 
-    Snapshots *tasks* to a list so the caller's post-drain
-    ``clear`` doesn't pull tasks out from under the gather.
+    With *log_exceptions*, a non-cancellation exception from a settled
+    task is logged at WARNING (still not propagated) instead of being
+    dropped silently. Snapshots *tasks* to a list so the caller's
+    post-drain ``clear`` doesn't pull tasks out from under the gather.
     Caller owns clearing the source collection.
     """
     tasks_list = list(tasks)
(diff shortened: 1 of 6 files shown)


esphbot (changes requested):
Blocking issues found.

- Identical supersede/orphan race remains in identity-rotation respawn

bdraco:
Good catch — fixed in e1f44460. `_refresh_identity_and_respawn_clients` had the identical structure: rotation keeps the `dashboard_id` and reconnects to the same receiver, so the non-awaiting cancel could let a late old registration supersede the fresh client. It now uses `await self._cancel_peer_link_client_and_wait(pin_sha256)`, and the rotation test asserts the awaiting cancel (`cancel.assert_awaited_once_with`).

esphbot (approved):
No blocking issues found.

bdraco:
Fixed in 2946ff2d. Switched the teardown wait from `with contextlib.suppress(asyncio.CancelledError): await handle.task` to `await asyncio.wait({handle.task})`. That addresses both this suggestion and a related hazard: `asyncio.wait` absorbs the task's own outcome (CancelledError *or* any teardown exception, so a settling-task error can't abort the respawn), while still propagating a cancellation of the *calling* task — so a shutdown-time cancel of the rebind/rotation task isn't silently swallowed. Added a regression test pinning that caller-cancellation propagates.

esphbot (approved):
No blocking issues found.

bdraco:
Switched to the in-house `drain_tasks` helper (c9c2d7a4) — the same idiom `stop()` uses. It cancels + awaits via `gather(return_exceptions=True)`, so the task outcome is retrieved (no "Task exception was never retrieved" at GC) and a non-cancel teardown error is absorbed rather than aborting the respawn, while a cancel of the calling task still propagates. The caller-cancellation regression test still passes. I kept the swallow-on-teardown behavior (consistent with `stop()`/`drain_tasks`) rather than adding a warning log, since the new client's connect supersedes any old session the failed terminate left behind.

esphbot (approved):
No blocking issues found.

bdraco:
Addressed the silent-failure (9ae39292): added an opt-in `log_exceptions` kwarg to `drain_tasks`. When set, a non-cancellation exception from a settled task is logged at WARNING (with `exc_info`) instead of being dropped — still absorbed, never propagated. `cancel_peer_link_client_and_wait` passes `log_exceptions=True` so a failed terminate/socket-close during teardown is now visible; `stop()` keeps the quiet default. Added tests for the logged-exception, no-log-on-plain-cancel, and silent-by-default cases.

esphbot (approved):
No blocking issues found.

PR #1632, opened 2026-06-20
Derive the advertised mDNS hostname from dashboard_id, not the OS hostname

bdraco:

The dashboard's peer-link mDNS SRV target was composed from `socket.gethostname()`, read live on every startup. macOS returns different values across reboots (`mac` one boot, `macbook-pro` the next), so the advertised hostname flipped even though the machine and its IP were unchanged; a paired peer saw the endpoint move and ran a needless rebind on every restart.

This derives the service-instance name and the SRV target purely from the persisted `dashboard_id` (`esphome-builder-<id>.local`), so the advertised identity is stable across reboots and the OS hostname can no longer flip it. The human machine label now travels in a new `friendly_name` TXT entry, surfaced on `RemoteBuildPeer`, so peers can still display the friendly name now that the instance name is an opaque identifier.

Existing pairings rebind once to the stable name on upgrade and then stay put; older receivers that don't broadcast `friendly_name` simply show the non-friendly identifier until they update.

- fixes <link to issue>

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#925

Must land lockstep: `name` becomes an opaque `esphome-builder-<id>` identifier and the human label moves to the `friendly_name` TXT field. The frontend ships prebuilt inside our wheel, so the bundle in any release carrying this change must include esphome/device-builder-frontend#925, or the discovered-build-server picker shows the opaque id instead of the machine name.

## Checklist


esphbot (changes requested):
Blocking issues found.

- TXT property table missing the new `friendly_name` entry
- `remote_build_host_added` event payload doc omits `friendly_name`

bdraco:
Addressed in 69a158a8:

- **🟡 1 + 2 (docs/API.md):** added `friendly_name` to the TXT-property table and to the `remote_build_host_added` payload doc (noting `name` is now the opaque `esphome-builder-<id>` identifier and `friendly_name` is the display label, `""` for older receivers).
- **🟢 1 (dead fallbacks):** dropped both unreachable branches now that `build_mdns_hostname` always returns a non-empty name — `self._name = host.split(".", 1)[0]` and `server` derives straight from `self._hostname` — and removed the inaccurate fallback comment.

bdraco:
@copilot good catch — fixed in c5c11cb0. Added `friendly_name: str` to the `RemoteBuildHostAddedData` TypedDict so it matches the `peer.to_dict()` payload the event actually fires; `tests/test_event_payload_contracts.py` passes.

esphbot (changes requested):
Blocking issues found.

- Discovered-host picker shows opaque IDs until the frontend follow-up lands

bdraco:
Agreed — corrected the Frontend coordination section to couple esphome/device-builder-frontend#925 and flagged it as lockstep. Since the frontend ships prebuilt in the wheel, the release carrying this change must bundle #925, otherwise the discovered-host picker shows the opaque `esphome-builder-<id>` until then. #925 is already approved.

PR #1634, opened 2026-06-20
Refresh deployed version after a flash so Update Available clears

### LINKED ISSUE #1630: [Bug] "Update Available" won't reset/clear after update was completed.
### What happened?

Following updating my ESP devices, the status still shows "Update Available" that I'm running 2026.06.01 and the most recent is 2026.06.02 (even after running the install and upgrade).  The ESP devices show 2026.06.02, but not being reported to the UI.

### Dashboard version

1.0.12

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

After a successful install or upload the dashboard pinned only the config hash, so the orange dot cleared but the deployed version stayed at its pre-update value; the device row kept showing the old version and an Update Available badge that never cleared even though the device was running the new build. This is permanent in deployments where mDNS cannot reach the dashboard (Docker Desktop on macOS is the textbook case), since deployed_version only ever updates from an mDNS announce or a Native API probe that stops once a version is already present.

This pins deployed_version from StorageJSON at flash time alongside the config hash, so both badges clear immediately. A successful flash means the new binary is on the device, so the compiled version is what it now runs; to catch a rollback or failed boot where mDNS can't tell us, it also schedules a one-shot Native API re-probe 60s after the flash, which confirms or corrects the running version.

- fixes https://github.com/esphome/device-builder/issues/1630

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -409,12 +409,20 @@ against legacy behaviour before assuming the simpler version suffices.
   `apply_api_encryption` both lean on the empty-string-means-plaintext
   distinction; a nullable boolean would lose it.
 - **Optimistic post-flash sync** in
-  `DevicesController._sync_deployed_hash_after_flash` pre-pins
-  `deployed_config_hash = expected_config_hash` after a successful
-  UPLOAD/INSTALL via `DeviceStateMonitor.apply_config_hash`, so the dot
-  clears immediately instead of waiting on the rebooted device's mDNS
-  announce. If the OTA silently failed, the next real announce pushes the
-  truth back through the same callback.
+  `DevicesController._sync_deployed_state_after_flash` pre-pins both
+  `deployed_config_hash = expected_config_hash` *and*
+  `deployed_version = StorageJSON.esphome_version` after a successful
+  UPLOAD/INSTALL via `DeviceStateMonitor.apply_config_hash` /
+  `apply_version`, so the dot and the "update available" badge clear
+  immediately instead of waiting on the rebooted device's mDNS announce.
+  A real announce later overwrites either through the same callback. mDNS
+  is dark in some deployments (Docker-bridge), so `refresh_after_job`
+  also arms a `loop.call_later` timer (`_schedule_version_reprobe`,
+  tracked in `_reprobe_timers`, cancelled in `stop()`) that ~60s later
+  forces one Native-API version probe via `request_version_reprobe` —
+  the only signal of a rollback / failed boot where the announce never
+  arrives. The forced probe still honours `_is_due`'s `priority_for !=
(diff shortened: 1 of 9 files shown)


esphbot (changes requested):
Blocking issues found.

- CLAUDE.md invariant references the renamed method and stale behavior

bdraco:
Addressed (59d0d28b): updated the CLAUDE.md optimistic-post-flash-sync invariant to the renamed `_sync_deployed_state_after_flash`, noting it now pins `deployed_version` (from `StorageJSON.esphome_version`) alongside `deployed_config_hash`, and documenting the `loop.call_later` re-probe path for mDNS-dark rollback detection. Confirmed no other stale references to the old method name remain.

esphbot (changes requested):
Blocking issues found.

- Stale invariant: renamed method + broadened contract not reflected in CLAUDE.md

bdraco:
Thanks, all three addressed:

- 🟡 CLAUDE.md invariant: updated in 59d0d28b (renamed `_sync_deployed_state_after_flash`, dual hash+version pin, and the call_later re-probe as the only rollback signal in mDNS-dark deployments).
- 🟢 `_read_compiled_esphome_version` coverage: the real body is exercised by `test_read_compiled_esphome_version_{reads_storage,missing_storage_is_empty,blank_version_is_empty}` (added in 8d74f12e, writing a real sidecar, no monkeypatch) so the `StorageJSON is None` / empty-version guards run; the codecov drop referenced was from the first commit before those landed.
- 🟢 Forced re-probe semantics: fixed in 67343532; a forced probe that connected and confirmed the existing version now returns instead of `_record_failure`, so a healthy re-probe no longer records a spurious cooldown / counts toward systemic-health. The normal path still cools down on a no-progress connect. Added a test.

esphbot (approved):
No blocking issues found.

bdraco:
Addressed the Copilot + koan follow-ups (096fc90d):

- Copilot (x3): the multi-line docstrings on `schedule_version_reprobe`, `_fire_version_reprobe`, and `_select_targets` now start their content on the line after `"""`, matching the repo convention.
- koan 🟢 (forced flag consumed before the dial/key early-returns): kept the up-front consume and documented the trade-off in a comment. Consuming on dispatch instead would let a device with an unresolvable Noise key stay forced and re-probe every sweep (forced bypasses cooldown); such a device can't be API-verified anyway, so its rollback check is deliberately skipped rather than churning.

esphbot (approved):
No blocking issues found.

PR #1646, opened 2026-06-22
Resolve ESPHome substitutions in MQTT broker discovery

### LINKED ISSUE #1643: MQTT discovery treats unresolved substitutions as broker host
### The problem

Device Builder's MQTT discovery monitor appears to read `mqtt.broker` from the raw YAML path and accepts an unresolved ESPHome substitution as the broker hostname.

With a config like:

```yaml
substitutions:
  mqtt_host: 192.0.2.10

mqtt:
  broker: ${mqtt_host}
  port: 1883
```

Device Builder starts its dashboard-side MQTT discovery against the literal substitution token instead of the rendered broker host:

```text
MQTT discovery starting — broker=${mqtt_host}:1883
MQTT broker ${mqtt_host}:1883 unreachable ([Errno -2] Name or service not known) — reconnecting in 5s
```

The ESPHome configuration itself is valid and should render `mqtt.broker` to `192.0.2.10`. The issue seems limited to Device Builder's own MQTT discovery monitor.

### Expected behavior

Device Builder should either:

1. resolve ESPHome substitutions before creating the MQTT monitor, or
2. treat unresolved `$var` / `${var}` values as unresolved in the fast parser so it falls back to the fully rendered `load_device_yaml(...)` path.

### Actual behavior

Device Builder creates a monitor for the literal hostname `${mqtt_host}` and then loops on DNS resolution failure.

### Likely cause

In `esphome_device_builder/controllers/_device_mqtt_coordinator.py`, `_resolve_broker(...)` first calls `parse_mqtt_block(yaml_content, secrets_map)`. `parse_mqtt_block(...)` resolves `!secret` values, but it does not resolve ESPHome substitutions. `_broker_from_block(...)` then accepts the non-empty string `${mqtt_host}` as a valid broker host, so `_resolve_broker(...)` returns early and never reaches the existing slow path:

```python
resolved = load_device_yaml(yaml_path)
broker = _extract_broker_from_config(resolved)
```

That fallback looks like it would handle the rendered config correctly, but it is bypassed because the unresolved substitution is treated as a valid host.

### Environment

Observed in a running Device Builder container with:

- `esphome-device-builder`: `1.0.12`
- `esphome-device-builder-frontend`: `0.1.174`
- `esphome`: `2026.6.2`
- Python: `3.13`

### Workaround

Using a literal broker hostname/IP in `mqtt.broker` avoids the dashboard-side MQTT discovery error, but it defeats normal ESPHome substitution usage.

bdraco:

Device Builder runs its dashboard side MQTT discovery by reading each device's `mqtt.broker` value straight from the YAML. The fast parser resolved `!secret` references but not ESPHome substitutions, so a config using `broker: ${mqtt_host}` started a monitor against the literal token `${mqtt_host}`, then looped forever on DNS failure.

This resolves `${var}` / `$var` substitutions from the file's own `substitutions:` block in both the fast parser and the slow path that expands packages. A broker that is still unresolved after both paths now starts no monitor and logs the existing once per device warning, instead of looping on DNS.

- fixes https://github.com/esphome/device-builder/issues/1643

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/_device_mqtt_coordinator.py
+++ b/esphome_device_builder/controllers/_device_mqtt_coordinator.py
@@ -19,7 +19,12 @@ import yaml
 from esphome.core import EsphomeError
 
 from ..constants import SECRETS_FILENAME
-from ..helpers.device_yaml import load_device_yaml
+from ..helpers.device_yaml import (
+    _UNRESOLVED_SUBSTITUTION_RE,
+    _extract_resolved_substitutions,
+    _resolve_substitutions,
+    load_device_yaml,
+)
 from ..helpers.yaml import FastestSafeLoader, load_yaml_fast_then_esphome
 from ..models import Device
 from ._device_mqtt_monitor import (
@@ -32,6 +37,10 @@ from ._device_mqtt_monitor import (
 _LOGGER = logging.getLogger(__name__)
 
 _DEFAULT_PORT = 1883
+
+# ``mqtt:`` fields read into an MqttBrokerConfig, each resolved through
+# the same !secret + substitution pipeline.
+_BROKER_FIELDS = ("broker", "port", "username", "password")
 
 
 class DeviceMqttCoordinator:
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Addressed the `port` suggestion from @Copilot and Kōan: `mqtt.port` is now resolved in both paths the same way as broker/username/password, so it honors both `!secret` and `${var}` substitutions instead of falling back to 1883. Added tests pinning port resolution via secret, via substitution (fast and slow paths), end-to-end at the coordinator, and the unresolved-port fallback. (`182205af`)

esphbot (approved):
No blocking issues found.

PR #1647, opened 2026-06-22
Honor a startup stop signal whose SystemExit was swallowed

bdraco:

`test_sigterm_during_startup_exits_zero` flakes on CI: "dashboard did not exit within 30s of SIGTERM". The child transcript shows the cause:

```
Exception ignored while calling weakref callback <function _get_module_lock.<locals>.cb ...>:
  File ".../__main__.py", line 131, in _exit_cleanly_on_signal
    raise SystemExit(0) from None
SystemExit: 0
```

The SIGTERM landed while Python was running an importlib module-lock weakref callback (during `main()`'s lazy imports). CPython ignores exceptions raised in weakref callbacks, so the handler's `raise SystemExit(0)` was swallowed; `_stop_requested` was set, but nothing acted on it, so startup completed and the dashboard served forever.

The handler always sets `_stop_requested` reliably (a plain assignment, never swallowed); only the `SystemExit` raise is best-effort. This adds a pre-serve check at the top of `_serve_until_stop`: if a stop is already pending, log and return before serving. That recovers the swallowed case at a deterministic checkpoint (the swallow window is the import phase, which finishes before `_serve_until_stop` runs). The signal handler is unchanged and stays log-free.

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/__main__.py
+++ b/esphome_device_builder/__main__.py
@@ -391,7 +391,15 @@ def _serve_until_stop(device_builder: DeviceBuilder) -> None:
     try/finally — so a half-started controller's teardown can surface a
     non-``CancelledError`` that escapes ``run_app``. With a stop pending
     that's a clean exit; a crash with no stop pending propagates.
+
+    The pre-serve check covers a stop signal whose ``SystemExit(0)`` was
+    swallowed (signals raised inside an import-machinery weakref callback are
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1649, opened 2026-06-22
Disable version history instead of crashing on an unusable .git

bdraco:

When `/config` is bind-mounted from a git submodule, `/config/.git` is a pointer file (`gitdir: ...`) whose target lives in the parent repo's `.git/modules` tree; inside the container that tree is not mounted, so `git init /config` follows the pointer, finds nothing, and aborts. The resulting `GitCommandError` is a `CalledProcessError`, not an `OSError`, so it slipped past `discover_or_init`'s catch and crash-looped startup; the dashboard never came up.

This makes version history honor its documented "soft, optional" contract:

- Detect a `.git` that git can't use here (broken submodule or worktree pointer, corrupt repo) before attempting init, and disable cleanly with an INFO log.
- Broaden the setup catch to any git failure (`OSError` plus `subprocess.CalledProcessError`), matching the module's existing `GIT_COMMIT_ERRORS` pattern, so a startup git error disables the feature rather than taking down the server.

- Related: esphome/device-builder#1710

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/version_history/git_repo.py
+++ b/esphome_device_builder/controllers/version_history/git_repo.py
@@ -191,8 +191,19 @@ class GitRepo:
                     self.config_dir,
                     toplevel,
                 )
+            elif (self.config_dir / ".git").exists():
+                # rev-parse found no work tree yet ``.git`` is present: an
+                # unusable git dir (a submodule / worktree pointer whose
+                # target isn't mounted, or a corrupt repo). Re-initialising
+                # over it fails, so disable rather than crash on init.
+                _LOGGER.info(
+                    "Config dir %s has a .git git can't use here (likely a submodule or "
+                    "worktree whose git dir isn't mounted); version history disabled",
+                    self.config_dir,
+                )
+                return
             self._init_repo()
-        except OSError as exc:
+        except (OSError, subprocess.CalledProcessError) as exc:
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Addressed the review feedback in 40cd1798:

- Reuse the existing `GIT_COMMIT_ERRORS` constant in `discover_or_init`'s catch instead of re-spelling the tuple, so the swallow-as-best-effort git error set stays in lockstep (Kōan esphome/esphome#1).
- Reset `enabled`/`toplevel`/`managed` in the `except` block so a git failure after the adopt branch enabled the repo can't leave it half-enabled (Copilot).
- Fixed the two test docstring grammar nits, and strengthened the failure-path tests to assert the repo is fully disabled (`enabled` False, `toplevel` None, `managed` False), plus a new test that the adopt-then-fail path rolls all three back.

esphbot (approved):
No blocking issues found.

bdraco:
Closed suggestion esphome/esphome#1 in f19f6470: extracted a `_disable()` helper and call it on both the unusable-.git early return and the except path, so neither leaves the repo half-enabled.

esphbot (approved):
No blocking issues found.

PR #1657, opened 2026-06-22
Guard sync_components against a mismatched installed ESPHome

### LINKED ISSUE #1654: sync_components.py can stamp a mismatched esphome_schema_version (no installed-ESPHome guard)
## Summary

`script/sync_components.py` can silently stamp a `esphome_schema_version` that does not match the ESPHome it actually derived platform metadata from. That stamped version is the pin `script/sync_boards.py` now trusts (via the guard added in #1653), so a mismatch here poisons the value the boards guard relies on.

## Background

PR #1653 added `_require_matching_esphome()` to `sync_boards.py`: the boards sync refuses to run unless the installed `esphome.const.__version__` equals `esphome_schema_version` in `components.index.json`. That closes the *consumer* side of the invariant.

`sync_components.py` is the *producer* of that pin, and has no equivalent guard:

- It takes `--version <tag>` (`script/sync_components.py:872`) to fetch a schema bundle, defaulting to the latest GitHub release (`resolve_latest_release`, `:899`).
- It also imports platform metadata from the **installed** ESPHome package, e.g. `from esphome.components.esp32.const import VARIANTS` (`:3088`), `esphome.components.libretiny`, `esphome.components.rp2040.boards`, `esphome.components.wifi` (`:3089-3091`), and more (`:5628`, `:5840`, ...).
- It writes `"esphome_schema_version": version` into the index from the `--version` tag (`:4375`, `:4421`).

Nothing asserts that the installed `esphome` equals the `--version` being fetched and stamped. So `sync_components.py --version 2026.6.2` run against an installed `esphome==2026.4.3` fetches the 2026.6.2 schema, derives platform metadata from 2026.4.3, and stamps `esphome_schema_version: 2026.6.2`. The boards guard then trusts that stamp.

## Proposed fix

Extract a shared `assert_installed_esphome(expected: str)` helper (natural home: `script/_catalog_split.py`, already the shared import surface both sync scripts use) that exits with clear install instructions when `esphome.const.__version__ != expected`, then wire both call sites:

- `sync_boards.py`: `expected = components.index.json["esphome_schema_version"]` (replacing the inline `_require_matching_esphome`).
- `sync_components.py`: `expected = version` (the resolved `--version` / latest tag), checked right after the version is resolved and **before** it imports `esphome.components.*` or stamps the index.

The assertion body is identical for both; only the source of `expected` differs (consumer matches what was written; producer matches what it is about to write).

## Notes

- Surfaced by the `/simplify` altitude review on #1653; deliberately left out of that PR to keep its scope to the boards-side tooling.
- Related: the `_write_index` helper in `sync_boards.py` duplicates the index validate/stage/atomic-replace logic in `swap_split_catalog_in` (`script/_catalog_split.py`). If the shared-helper extraction above happens, consider also factoring an index-only `swap_index_in` out of `swap_split_catalog_in` and reusing it from both. Minor; bundle only if convenient.

bdraco:

`sync_components.py` derives platform metadata (variants, board tables, wifi sets) from the **installed** ESPHome, but stamps `esphome_schema_version` from the fetched `--version` tag, with nothing asserting the two agree. So running it against a mismatched esphome ships one release's metadata labelled as another's version.

This adds a shared `assert_installed_esphome()` (`script/_esphome_version.py`) and calls it right after the version resolves, before any `esphome.components` import. `sync_boards.py`'s existing single-board guard now routes through the same helper; the only difference is the match mode (the board catalog matches canonically since board tables are base-stable, components match exactly since the schema bundle is version-specific).

No CI job runs `sync_components.py` against a mismatched esphome (only `sync-component-catalog.yml`, which already installs `esphome==${version}`), so this surfaces contributor mistakes without touching automation.

- fixes #1654

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/script/_esphome_version.py
+++ b/script/_esphome_version.py
@@ -1,0 +1,46 @@
+"""Shared ESPHome-version guard for the catalog sync scripts.
+
+``sync_boards.py`` and ``sync_components.py`` both derive platform metadata
+from the *installed* ESPHome, so each must run against the version its output
+records, or it ships metadata from one ESPHome stamped with another's version.
+This centralizes the "installed == expected, else exit with install
+instructions" check both perform.
+"""
+
+from __future__ import annotations
+
+import sys
+from collections.abc import Callable
+
+
+def assert_installed_esphome(
+    expected: str,
+    *,
+    what: str,
+    normalize: Callable[[str], str] | None = None,
+    alt_fix: str | None = None,
+) -> None:
+    """
+    Exit unless the installed ESPHome matches *expected*.
+
+    *what* names the caller for the message; *normalize* (e.g. beta -> base) is
+    applied to both sides before comparing; *alt_fix* appends an extra
+    remediation line to the mismatch message.
+    """
+    canon = normalize or (lambda v: v)
+    install = f"    uv pip install 'esphome=={expected}'   # or: pip install 'esphome=={expected}'"
+    try:
+        from esphome.const import __version__ as installed
+    except ImportError:
+        raise SystemExit(
+            f"{what}: ESPHome is not importable in this interpreter ({sys.executable}).\n"
+            f"To fix, install it into the project venv and re-run:\n{install}"
+        ) from None
+    if canon(installed) != canon(expected):
+        msg = (
+            f"{what}: needs ESPHome {expected}, but {installed} is installed.\n"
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

Issue #1675, opened 2026-06-24
[Bug]


wesselah:
### What happened?

After the latest update to 2026.6.2 the online offline does not function anymore , when devices are of line they are still online in the dashbord. can be easy tested disconnect online device and it will stay online

### Dashboard version

2026-6.2

### How did you install it?

Home Assistant App (preview toggle)

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell
no errors in the log
```

### Anything else?

[5c53de3b_esphome_2026-06-24T12-44-38.876Z.log](https://github.com/user-attachments/files/29293176/5c53de3b_esphome_2026-06-24T12-44-38.876Z.log)

breti:
See https://github.com/esphome/device-builder/issues/1524
mDNS TTL is 4500 seconds == 75 minutes

wesselah:
I did read #1524 and it does not fit my problem even after 75 minutes the devices are still online and when i open the sidebar i see that mDNS keeps refreshing . Is this somewhere in mDNS, is it a setting ? Because mDNS does not seem to timeout

breti:
The mDNS messages are sent by the ESP devices. When they're offline, mDNS messages can't really come through anymore. 

Could you maybe use Wireshark to check what's going on?

wesselah:
Just a bit of background and i am also running matter/ thread on the latest release 9.02 and as far as i know the tbr is publish mDNS DNS-SD discovery packets , which shoudl only be for the thread/ matter nodes , i am just wondering. Never had a issue with the online offline status in the esphome dashboard but i upgraded matter and esphome on the same day and then this issue did happen. To be honest never new this online offline thing was a mdns thing

wesselah:
> The mDNS messages are sent by the ESP devices. When they're offline, mDNS messages can't really come through anymore.
> 
> Could you maybe use Wireshark to check what's going on?

I will hookup wireshark tomorrow

PR #1682, opened 2026-06-24
Expose the mDNS PTR record TTL on the reachability snapshot

bdraco:

People keep filing "offline device still shows online" reports because mDNS is passive and that isn't obvious. An ESPHome device's online state comes from its zeroconf PTR record; when a device powers off without a goodbye the record only ages out after its TTL (commonly around 75 minutes). We deliberately don't actively probe every device on a short interval, since that would flood the network with mDNS traffic.

This is the backend half of surfacing that in the UI. It adds `mdns_ptr_ttl_remaining_seconds` to the reachability snapshot (`MdnsCacheInfo` and the `reachability_state` event); it is the PTR record's own remaining TTL, the countdown to the zeroconf `Removed` event that flips an mDNS owned device offline, kept distinct from the existing freshest-record `mdns_ttl_remaining_seconds` (which tracks the ~120s A record the refresh loop renews). It also adds a one shot `devices/get_reachability` command so the open drawer can re-read the live value; a pure PTR TTL refresh is deduped by zeroconf and never pushed, so polling is the only way to catch a healthy device re-announcing. The read is scoped to a single device, not fleet wide.

The frontend renders the countdown and does the polling; this PR only exposes the data.

- part of https://github.com/esphome/device-builder/issues/1675

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#969

## Checklist


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1692, opened 2026-06-25
Resolve substitutions when reading the API encryption key

### LINKED ISSUE #1691: [Bug] Asking for view the API key shows the substitution pattern.
### What happened?

I did a pull (as of right now) on latest image ("image: ghcr.io/esphome/esphome:latest").

My device YAML has this:

```
packages:
  connect: !include includes/connections.yaml
```
connections.yaml start with:
```
substitutions:
  api_key: !secret api_key
```
and later in that same connections.yaml file:
```
api:
  encryption:
    key: ${api_key}
```
But when I "Show API Key" for the device's API key on the UI it shows this:

Image

Or if I click reveal:

Image

### Dashboard version

esphome-device-builder 1.0.12 (esphome 2026.6.2)

### How did you install it?

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

"Show API Key" returned the literal substitution token `${api_key}` instead of the resolved key when the key is built from a substitution, e.g.:

```yaml
substitutions:
  api_key: !secret api_key
api:
  encryption:
    key: ${api_key}
```

`devices/get_api_key` reads the key in-process first via `load_device_yaml`, which resolves `!secret` / `!include` / `packages:` but not `${...}` substitutions. The raw read returned the non-empty literal `${api_key}`, which is truthy, so the `if key:` gate short-circuited and the `esphome config --show-secrets` fallback (which would resolve it) never fired. The resolved secret was already sitting in the merged config's `substitutions:` block; we just never applied it.

The same read also feeds the **Native-API mac/version fallback** (`get_api_connection`), used to fill `mac_address` / `deployed_version` when mDNS is dark. With a substituted key it handed the literal `${api_key}` to the Noise handshake, which failed the wrong-key path, so those fields stayed blank for affected devices. So this is not just a display glitch; it broke the API connection itself for substitution-based keys.

This adds `get_resolved_api_encryption_key`, mirroring the #1646 MQTT broker fix: it expands the key against the merged `substitutions:` map using the existing `_resolve_substitutions` / `_extract_resolved_substitutions` helpers, and returns `""` when the token can't be resolved in-process (e.g. a Jinja-templated package) so the `esphome config` subprocess fallback still runs. The in-process read sites in `controllers/devices/api_key.py` (Show API Key and the Native-API info probe) now use it.

- fixes https://github.com/esphome/device-builder/issues/1691

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1693, opened 2026-06-25
Surface the template lambda field on the main form

bdraco:

For a template entity the lambda is the field that produces its value, yet the catalog generator flagged it advanced, so it was hidden under "Show advanced settings". This surfaces the lambda on the main form alongside `optimistic` and the action handlers, which #1324 already promoted there. The switch template docs even lead with a lambda example (https://esphome.io/components/switch/template/), so it is plainly a primary field rather than an advanced one.

`_promote_template_controls` now also clears the advanced flag on `lambda` for `.template` ids; the promotion stays scoped to template ids, so a regular component's lambda is untouched. The catalog was regenerated from schema 2026.6.2; the only change in the bodies is the `lambda` entry dropping its advanced flag across the 11 template components that carry one (binary_sensor, cover, datetime, lock, number, select, sensor, switch, text, text_sensor, valve).

- Reported via dashboard feedback (Template Text lambda hidden under "Show advanced"); no separate tracking issue.

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1694, opened 2026-06-25
Skip wrapped JSX element blocks in MDX description fallback

bdraco:

Some component descriptions showed raw MDX markup in the Add component dialog instead of prose; for example the Lite-On LTR sensors rendered the literal `<Figure src={ltr501FullImg} alt="" caption="..." layout="constrained" />` text.

When the schema has no description, the catalog falls back to the first prose paragraph of the docs MDX. That fallback skipped a tag's opening line but not its continuation lines, so a `<Figure>` wrapped across several lines leaked its `src`, `alt`, `caption` attribute lines into the description. This hardens the fallback to skip the whole wrapped element, and adds a final guard so residual attribute markup can never be returned.

The six affected entries (ltr501, ltr_als_ps, he60r, sprinkler, kamstrup_kmp, rc522) now have clean frontmatter `description:` fields upstream, so their committed text corrects on the next nightly catalog sync; this change is the durable guard so the leak cannot recur for a future figure first component.

- n/a (reported from the Add component dialog)

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

bdraco:
Thanks, both good catches.

Suggestion 1 (guard not pinned): addressed in 54e16d64. Added `test_inline_jsx_in_prose_line_is_skipped`, which feeds a prose line carrying an inline `<Figure …/>` so the line is accumulated (it doesn't start with `<`, so the prefix skip can't catch it) and the residual-markup guard's taken branch fires. Confirmed via a mutation check: with the guard the result is the clean follow-up paragraph, without it the inline-markup line leaks, so removing the guard now fails the test.

Suggestion 2 (unclosed / paired tags): leaving as is, intentionally. A paired tag's open line ends with `>`, so it hits the single-line prefix skip and its inner prose is still collected; only a genuinely unclosed self-closing tag would keep `in_jsx_block` set and swallow trailing prose, and that is malformed MDX we don't see in the docs. Even in that case the function returns an empty description (falls through to the schema source), never leaked markup; the residual-markup guard keeps that invariant regardless.

esphbot (approved):
No blocking issues found.

PR #1701, opened 2026-06-26
Union referenced bus hubs into component dependencies

bdraco:

Adding a component that rides a bus (for example an `atm90e32` button, or its `atm90e32` sensor hub) did not prompt to add the `spi` bus, so the config compiled to `Couldn't find any component that can be used for 'spi::SPIComponent'`. The catalog's `dependencies` is what drives the frontend's "add the bus" prompt, and most bus devices already list their bus there; but ESPHome doesn't always put the bus in a component's `DEPENDENCIES` (atm90e32, max6675, several i2c touchscreens rely on the device schema's `spi_id`/`i2c_id` ref plus a config-time check), so the schema index shipped about 30 of them bus-less.

The generator already unions a platform's own hub stem into `dependencies` when it's bound by a `use_id` ref instead of `DEPENDENCIES`; this applies the same idea to bus references. `_referenced_buses` collects any `references_component` whose category is `bus` (spi/i2c/uart/one_wire/modbus/canbus, via the existing `_CATEGORY_OVERRIDES` table) and unions it in. It is scoped to bus refs deliberately; optional sibling refs like `output_id` must not become hard dependencies.

The regenerated catalog is not in this PR; the nightly sync will pick it up. I verified locally that a regen adds only the bus dependency to the affected entries.

- n/a (reported in testing: adding an `atm90e32` button did not require `spi`)

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/script/sync_components.py
+++ b/script/sync_components.py
@@ -2113,6 +2113,15 @@ def build_component_entry(
     if domain and stem not in dependencies and _references_own_hub(config_entries, stem):
         dependencies.append(stem)
 
+    # A device that cross-references a bus hub (``spi_id`` / ``i2c_id`` /
+    # ``uart_id``) needs that bus block to exist; ESPHome enforces it at config
+    # time but doesn't always list it in ``DEPENDENCIES`` (atm90e32, max6675,
+    # i2c touchscreens), so the schema index ships them bus-less and the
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #1706, opened 2026-06-26
Materialise nested featured presets and flag preset fields

bdraco:

Two fixes for featured-component materialisation, both surfaced by the OLIMEX
ESP32-PoE-ISO "Onboard Ethernet" featured component.

1. **Nested presets reach their leaves.** A dict-shaped preset on a NESTED group
   (e.g. `clk: {pin, mode}`) was stamped onto the parent entry's `default_value`,
   but the frontend seeds NESTED groups by their leaf children, so `clk.pin` /
   `clk.mode` rendered empty. `_apply_preset_value` now recurses a dict preset into
   the group's leaves, propagating `locked`. A `pin`-typed entry keeps its dict
   value verbatim (the pin renderer reads it), so `power_pin` is unchanged.

2. **Preset fields are flagged with `from_preset`.** Value-only presets are
   pervasive across the bundle catalog and are indistinguishable from plain catalog
   defaults once materialised. The new `ConfigEntry.from_preset` flag marks every
   preset-bearing entry, so the frontend can seed presets while leaving plain
   catalog defaults unseeded (companion frontend PR makes the featured add emit only
   the preset fields, matching the create-time auto-add and re-enabling Add).

`from_preset` defaults to `False` and is set only at featured-materialise time, so
the static catalog JSON is unaffected.

- N/A

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#994

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/components/_resolve.py
+++ b/esphome_device_builder/controllers/components/_resolve.py
@@ -14,6 +14,7 @@ from ...models import (
     ComponentCatalogIndexEntry,
     ComponentCategory,
     ConfigEntry,
+    ConfigEntryType,
     FeaturedComponent,
     FieldPreset,
 )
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1714, opened 2026-06-27
Compute featured components' locked_pins from the PIN schema at sync

bdraco:

Adds an authoritative `locked_pins` to each featured component so the frontend
can hide a featured card when its fixed pins are already configured, without
guessing pin-ness from field-key names.

`FeaturedComponent.locked_pins: dict[str, int]` (e.g. `{"scl": 0, "sda": 1}`) is
computed at sync time in `sync_boards.py`: for each board's featured components,
load the underlying component body, and for every manifest field whose schema
entry is `type: pin` (the catalog already captures this, e.g. i2c `scl`/`sda` use
`pins.internal_gpio_pin_number`) and whose preset is `locked`, record the
canonical GPIO. It lands in both the per-board body JSON (`get_board` ships it)
and `featured_components.index.json`. Empty (omitted) for non-pin components.

The board catalog was regenerated; the only diff in the body files is the new
`locked_pins` maps (verified: 327 bodies changed, all `locked_pins`-only).

Companion frontend PR reads `locked_pins` to do the catalog hide and drops a
key-name heuristic (esphome/device-builder-frontend#1008).

- N/A (supports esphome/device-builder-frontend#1006)

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1008

## Checklist

- [x] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (board catalog regenerated via `script/sync_boards.py`).


esphbot (changes requested):
Blocking issues found.

bdraco:
Both addressed in bb631b1d:

- **Expander channels**: `_canonical_gpio` now returns `None` for a pin mapping that carries any key beyond the internal-GPIO sub-keys (`number`/`mode`/`inverted`/`ignore_strapping_warning`/`allow_other_uses`/`drive_strength`). A hub-referencing key like `pcf8574` means an expander-local channel, not a board GPIO, so it's skipped. Regenerated: kincony/relay expander pins drop out of `locked_pins` (e.g. `kincony_b16` `b16_input01` -> `{}`) while real board GPIOs on the same board stay (`binary_sensor_gpio_17` -> `{pin: 48}`). Added `test_featured_locked_pins_skips_io_expander_pins`.
- **Docs**: added a `locked_pins` bullet to `docs/API.md` next to `multi_conf`.

esphbot (approved):
No blocking issues found.

PR #1726, opened 2026-06-28
Synthesize an all-recommended full-setup bundle for every board

bdraco:

Boards imported from devices.esphome.io ship a `featured_components` list, and the importer derives `featured_bundles` so related components add together; but it only emits a bundle when one featured component references another via a `type:"id"` field. Boards whose recommended components are independent get no bundle at all, so they have no "full setup" option: the ESP32 Relay x4 has four standalone `switch.gpio` relays and no bundle, while the Arlec GLD112HA does because its light references two outputs.

This adds a catalog-build step, `_synthesize_full_setup_bundles`, that gives a board an `all_recommended` bundle covering all its recommended components when one doesn't already exist. It is synthesized in `sync_boards.build_catalog` (alongside `_stamp_featured_locked_pins`) and baked into the body JSON, so the hand-curated manifests stay untouched and re-running the devices importer is not required. Members are ordered dependencies first (existing bundle members, then standalone). Synthesis is skipped when a board has fewer than two recommended components or an existing bundle already covers them all.

Only complete onboard configs are eligible: synthesis is gated on a new `full_config` board attribute that defaults to whether the board is a devices.esphome.io import and is overridable per manifest. So a hand-curated starter kit's optional components aren't offered as one click, an onboard product can opt in, and an import with optional parts can opt out; the importer preserves the override across re-imports. `full_config` lives on the board body only, so the slim picker index is unchanged.

The existing frontend already renders bundle cards from `featured_bundles`, so the new bundle surfaces immediately; the companion frontend PR esphome/device-builder-frontend#1017 adds a one-shot add for it.

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist


esphbot (changes requested):
Blocking issues found.

- all_recommended bundle can aggregate pin-conflicting components → invalid config

bdraco:
Addressed the blocking review: `_synthesize_full_setup_bundles` now skips a board whose featured components exclusively claim the same board GPIO, so the bundle can't aggregate pin-conflicting components into an invalid config. Pins marked `allow_other_uses` and namespaced expander channels are shared on purpose and don't count. Verified across the catalog: zero remaining genuine conflicts; the one affected board (oil gauge, two switches on GPIO13) no longer gets a bundle, while the allow_other_uses case (tuya display) keeps its bundle. Added a regression test.

esphbot (approved):
No blocking issues found.

bdraco:
Thanks. Addressed all three non-blocking suggestions:

1. The pin-conflict skip now logs the board (`Skipping all_recommended for <id>: featured components share a board GPIO`), so a future import regression is visible in the sync output.
2. `_has_pin_conflict` now mirrors ESPHome's validator: a board GPIO used more than once is a conflict unless every usage sets `allow_other_uses` (the mixed marked/plain case is now caught). Output is unchanged across the catalog; added a regression test for the mixed case.
3. Documented the synthesized `all_recommended` bundle in `definitions/README.md` (synthesized at build time for `full_config` boards, baked into the body only, don't hand-add to manifests).

esphbot (approved):
No blocking issues found.

PR #1743, opened 2026-06-29
Skip the default Wi-Fi block on adopt when shared secrets are absent

### LINKED ISSUE #1742: [Bug] Take Control fails when Wi-Fi uses valid custom secrets / wifi.networks instead of wifi_ssid and wifi_password
### What happened?

When using the new ESPHome Device Builder dashboard, the Take Control / adoption flow fails for an [Athom RGBCCT bulb](https://www.athom.tech/blank-1/e14-color-bulb-for-esphome) when my Wi-Fi configuration uses a valid ESPHome `wifi.networks` structure with custom secret names.

I am using multiple Wi-Fi networks, so my configuration intentionally uses separate secret keys:

```
wifi:
  networks:
    - ssid: !secret wifi_0_ssid
      password: !secret wifi_0_key
    - ssid: !secret wifi_1_ssid
      password: !secret wifi_1_key
```

This is deliberate. It lets me keep multiple SSIDs and keys clearly named in secrets.yaml, instead of forcing everything through the single shared `wifi_ssid` / `wifi_password` convention.

However, when I click Take Control for the discovered Athom device, Device Builder appears to generate or validate an imported config that still assumes these standard shared secrets exist:

```
wifi:
  ssid: !secret wifi_ssid
  password: !secret wifi_password
```

Because my secrets.yaml does not contain those two standard keys, adoption fails with:

`invalid_args: Can't import — config doesn't validate: Secret 'wifi_ssid' not defined in "/config/esphome/athom-rgbct-light-9ade7c.yaml", line 16, column 9. Fix the errors in the editor and try again.`

This blocks the adoption flow before I can meaningfully fix the generated YAML in the editor.

### Reproduction steps
Use the new ESPHome Device Builder dashboard.
Have an adoptable Athom RGBCCT bulb discovered by the dashboard.

Configure secrets.yaml with custom Wi-Fi secret names, for example:

```
wifi_0_ssid: "..."
wifi_0_key: "..."
wifi_1_ssid: "..."
wifi_1_key: "..."
```

Do **not** define:

```
wifi_ssid: "..."
wifi_password: "..."
```
Try to use Take Control on the discovered device.
The import/adoption fails because `wifi_ssid` is missing.
### Minimal relevant YAML

This is the relevant shape of the config I am trying to use:

```
substitutions:
  name: "athom-rgbct-light"
  friendly_name: "Athom RGBCCT Bulb 2"
  ipv6_enable: true
  sntp_server_1: "ntp.time.nl"
  sntp_server_2: "ntp1.time.nl"
  sntp_server_3: "ntp2.time.nl"
  timezone: "Europe/Amsterdam"
  sntp_update_interval: "2h"

packages:
  athom_base: github://athom-tech/athom-configs/athom-rgbct-light.yaml

api:
  encryption:
    key: !secret encryption_key_athom-rgbct-light-9ade7c

wifi:
  networks:
    - ssid: !secret wifi_0_ssid
      password: !secret wifi_0_key
    - ssid: !secret wifi_1_ssid
      password: !secret wifi_1_key

web_server:
  port: 80
  version: 3
  local: true
  auth:
    username: !secret web_server_username
    password: !secret web_server_password

ota:
  - platform: esphome
  - platform: web_server
```
### Expected behavior

Device Builder should not hard-fail the adoption flow simply because the user does not use the exact secret names `wifi_ssid` and `wifi_password`.

Ideally, the adoption/import path should handle at least one of these cases:

If the imported or package-expanded config already contains a valid `wifi:` component, including `wifi.networks`, do not append or require the hard-coded default Wi-Fi block.
If the dashboard cannot safely determine that, still allow the generated YAML to be created/opened for editing instead of blocking adoption before the user can fix it.
If standard shared secrets are required for a specific path, the UI should explain that clearly and offer to create aliases, instead of failing with a validation error that looks like a broken config.
### Actual behavior

The Take Control flow fails with:

`invalid_args: Can't import — config doesn't validate: Secret 'wifi_ssid' not defined in "/config/esphome/athom-rgbct-light-9ade7c.yaml", line 16, column 9. Fix the errors in the editor and try again.`

The error is technically accurate, but from the user’s perspective it is confusing because their Wi-Fi secrets are defined, just not under the exact default names expected by the adoption helper.

### Dashboard version

2026.6.2

### How did you install it?

Home Assistant App

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell
invalid_args: Can't import — config doesn't validate: Secret 'wifi_ssid' not defined in "/config/esphome/athom-rgbct-light-9ade7c.yaml", line 16, column 9. Fix the errors in the editor and try again.
```

### Anything else?

I think this may be related to the same general design area as:

#1567 — Cannot create any device without Wi-Fi secrets
#1583 — Generate a no-network stub when no Wi-Fi secrets exist
#1590 — Collect Wi-Fi credentials in the create-device wizard, not first-run onboarding

Those issues/PRs seem to address the create-device / onboarding side of the problem. This report is specifically about the Take Control / adoption / import path.

The broader issue seems to be that Device Builder still treats `wifi_ssid` and `wifi_password` as the only shared Wi-Fi credential model in some places. That works for simple first-time setups, but it breaks down for valid ESPHome configurations using:

```
wifi:
  networks:
    - ssid: !secret wifi_0_ssid
      password: !secret wifi_0_key
    - ssid: !secret wifi_1_ssid
      password: !secret wifi_1_key
```

I was able to understand the failure because I am comfortable reading YAML and tracing the generated config. But this could be quite unfriendly for less technical users. A new user dipping their toes into ESPHome with a ready-made Athom or any other "[Made for ESPHome](https://devices.esphome.io/made-for-esphome/)" device may interpret this as “ESPHome is broken” or “I did something wrong,” even though their Wi-Fi setup is valid.

That is a shame, because the new Device Builder is otherwise a much more approachable entry point. Adoption should ideally be forgiving here: either detect existing valid Wi-Fi configuration, avoid appending the default `wifi_ssid` / `wifi_password` block, or let the user edit the generated config before validation failure blocks the workflow.

bdraco:

Take Control adoption appended a default `wifi:` block referencing `!secret wifi_ssid` / `wifi_password` whenever the device broadcast `network=wifi`, without checking that those keys exist in `secrets.yaml`. A user running `wifi.networks` with custom secret names (`wifi_0_ssid`, `wifi_0_key`, ...) therefore hit a hard validation failure, "Secret 'wifi_ssid' not defined", before they could open the editor to fix it.

This gates the default block on `wifi_secrets_defined`, the same check the create-device path already uses; when the shared secrets are absent the block is suppressed, so adoption lands a config the user completes in the editor with their own `wifi:`. When the secrets are present, behavior is unchanged.

Suppression rides the existing `network` argument to upstream `import_config` (a non-wifi value already skips the block, same path Ethernet adoption uses); an explicit upstream flag would be a cleaner end state and is noted as a follow-up.

- fixes https://github.com/esphome/device-builder/issues/1742

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/devices/importable.py
+++ b/esphome_device_builder/controllers/devices/importable.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import asyncio
 import logging
 from dataclasses import replace
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Final
 
 from esphome import const
 from esphome.storage_json import ignored_devices_storage_path
@@ -13,6 +13,7 @@ from esphome.storage_json import ignored_devices_storage_path
 from ...helpers.api import CommandError
 from ...helpers.json import JSONDecodeError, dumps_indent, loads
 from ...helpers.lazy_module import async_import_module
+from ...helpers.secrets_state import read_secrets_yaml, wifi_secrets_defined
 from ...models import (
     AdoptableDevice,
     DeviceState,
@@ -27,6 +28,11 @@ if TYPE_CHECKING:
     from .controller import DevicesController
 
 _LOGGER = logging.getLogger(__name__)
+
+# import_config only special-cases ``network == CONF_WIFI``, so any other
+# value suppresses its default `!secret wifi_ssid` block; "" reads as
+# "scaffold no network block". Revisit if upstream gains an explicit flag.
+_NO_DEFAULT_WIFI_BLOCK: Final = ""
 
 
 async def import_device(
@@ -59,6 +65,16 @@ async def import_device(
     )
     network = adoptable.network if adoptable and adoptable.network else const.CONF_WIFI
     loop = asyncio.get_running_loop()
+    if network == const.CONF_WIFI:
+        secrets = await loop.run_in_executor(
+            None, read_secrets_yaml, controller._db.settings.config_dir
+        )
+        if not wifi_secrets_defined(secrets):
+            # No wifi_ssid/wifi_password in secrets.yaml; import_config would
+            # otherwise append a `!secret wifi_ssid` block that fails validation
+            # (#1742). Skip it so adoption lands a config the user can complete
+            # in the editor (e.g. their own wifi.networks).
+            network = _NO_DEFAULT_WIFI_BLOCK
     # ``esphome.components.dashboard_import`` pulls in ~14 MB of
     # upstream code; load it through ``async_import_module`` so
     # the first adoption pays the cost on the dedicated import
--- a/tests/controllers/devices/test_import.py
+++ b/tests/controllers/devices/test_import.py
@@ -43,6 +43,20 @@ def _seed_import_state(controller: DevicesController) -> None:
     controller.state.import_result = {}
 
 
+def _write_wifi_secrets(config_dir: Path, *, standard: bool = True) -> None:
+    """Seed ``secrets.yaml``; *standard* writes ``wifi_ssid`` / ``wifi_password``.
+
+    When False, only a custom-named key is written so ``wifi_secrets_defined``
+    reports the shared secrets as absent (the #1742 scenario).
+    """
+    body = (
+        "wifi_ssid: my-network\nwifi_password: hunter2\n"
+        if standard
+        else "wifi_0_ssid: my-network\nwifi_0_key: hunter2\n"
+    )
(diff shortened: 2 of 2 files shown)


bdraco:
closing as I decided on an alternate solution which is much more involved

PR #1744, opened 2026-06-29
Pin CI esphome to the catalog version

bdraco:

CI installs esphome unpinned, so when a new esphome ships the lint job's `sync-boards` pre-commit hook re-stamps `boards.index.json` from the newer esphome and reds `main` plus every open PR until the catalog is re-synced by hand; this is what broke the 2026.6.3 catalog PR and would recur on every release.

Pin esphome for CI in `esphome-constraints.txt` and install the lint job with `-c`, so a release no longer reds CI; the catalog advances only when a sync deliberately bumps the pin. The component sync writes the pin in lockstep with the catalog it regenerates, and the device sync now follows the pin instead of tracking the latest prerelease, so both sync PRs stay self-consistent (pinned esphome equals the committed board stamp). While here, the lint backstop and the device-sync change gate are fixed to check `boards.index.json` instead of a `boards.json` that does not exist, and the component-sync version resolution is hardened against shell injection; the dispatch input is read via env and the resolved version is validated before any step installs, pins, or interpolates it.

Runtime test jobs (pytest, e2e) stay on the latest esphome on purpose, alongside the beta and dev channel jobs, so forward-compat breakage still surfaces; only the catalog-enforcing lint job is pinned.

- Surfaced by the 2026.6.3 catalog sync (PR #1740); no separate tracking issue.

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/.github/workflows/sync-component-catalog.yml
+++ b/.github/workflows/sync-component-catalog.yml
@@ -80,16 +80,28 @@ jobs:
         # 60 req/hr unauthenticated cap on shared runner IPs.
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Read the dispatch input via env, never inline ``${{ inputs.version }}``
+          # into the script: it's attacker-controllable and a textual splice
+          # would run as shell. ``$INPUT_VERSION`` is data, not code.
+          INPUT_VERSION: ${{ inputs.version }}
         run: |
           set -euo pipefail
-          if [ -n "${{ inputs.version }}" ]; then
-            echo "version=${{ inputs.version }}" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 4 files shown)


esphbot (changes requested):
Blocking issues found.

- Re-stamped board catalog is not validated before the PR opens

esphbot (changes requested):
Blocking issues found.

- Re-stamped board catalog is not validated before the PR opens

bdraco:
Addressed both in 0662d94b. The component sync now runs validate_definitions.py right after re-stamping boards, so an invalid board catalog can't open a PR; mirrors the device-catalog sync. The lint backstop now gates on git status --porcelain instead of git diff, so a newly generated untracked body file (a manifest added without regenerating) fails the check too.

esphbot (approved):
No blocking issues found.

Issue #1748, opened 2026-06-29
[Bug] Dead device stuck `online` after restart — stale mDNS PTR from a reflector marks device online with no reachability check


ReneNulschDE:
### What happened?

1. Restart container
2. Open Dashboard
3. See that a device XXX is offline (the device is broken since months)
4. wait one minute
5. The device XXX is reported online

### Dashboard version

2026.6.3

### How did you install it?

Home Assistant App (preview toggle)

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell
17:34:04 Device powerplug_02 (powerplug-02.yaml): unknown → offline (via ping)
17:35:04 Device powerplug_02 (powerplug-02.yaml): offline → online (via mdns)
```

### Anything else?

AI-claude analysis:

 device that has been physically dead for months reliably shows as **online** in the dashboard. On every add-on restart the device is first correctly reported `offline (via ping)`, then ~30–60 s later flips to `online (via mdns)` and stays there indefinitely — no `Removed`/offline event ever follows.

Root cause: `_on_esphomelib_service_state_change` claims `ONLINE` on a bare `_esphomelib._tcp.local.` PTR event **without any reachability cross-check**, and never walks that claim back if the service can't actually be resolved on the wire. On real home networks an mDNS **reflector/cache** (Home Assistant's own avahi reflector, a FritzBox, or another host) re-serves a stale PTR for a long-gone device, so the `Removed` (TTL-expiry) signal the design relies on never fires.

## Environment

- ESPHome Device Builder **2026.6.3** (HA OS add-on, host-network)
- Device: ESP8266 power plug, removed from the network months ago; still present as a config in the dashboard.

## Steps to reproduce

1. Have a device configured in the builder whose hostname is still cached/answered by an mDNS reflector on the LAN (e.g. a FritzBox with a static lease, or HA's own avahi reflector), but whose hardware is powered off.
2. Restart the add-on.
3. Observe the device status over the next ~60 s.

## Expected

The device stays `offline` — nothing is actually reachable, and no live ESPHome node answers `_esphomelib._tcp.local.`.

## Actual

```
17:34:04 Device powerplug_02 (powerplug-02.yaml): unknown → offline (via ping)
17:35:04 Device powerplug_02 (powerplug-02.yaml): offline → online (via mdns)
```

After 17:35 there are **no further transitions** — it is stuck `online`.

## Diagnostics

All run from inside the add-on container, device hardware confirmed dead:

[...]

PR #1750, opened 2026-06-29
Claim mDNS online only after the service resolves

bdraco:

The esphomelib mDNS browser callback marked a device online the moment a PTR
arrived, before resolving anything, and the cache-miss resolve task returned
silently when the SRV/A never came back. The eager claim was never walked back,
so a PTR with no resolvable address latched the device online forever with no IP;
the mDNS source then locked it out of the ICMP sweep, so nothing re-checked it.

This drops the eager claim. `_apply_service_info` is now the single online-claim
point, reached only on a cache hit or a successful wire resolve, so an announce
that never resolves claims nothing and falls through to the ICMP sweep, matching
the active-resolve path. It adds no new queries; the cache-miss resolve already
ran, we just stop claiming online before it succeeds.

Note on scope: this fixes the case where the SRV/A cannot be resolved (the one in
the report). A reflector that re-serves a full stale record (PTR plus SRV plus A)
still resolves and is indistinguishable from a live device without ICMP; that is
a misbehaving reflector and is out of scope.

- Related to https://github.com/esphome/device-builder/issues/1748

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1752, opened 2026-06-29
Lift platform-style bus dependencies (one_wire, canbus) into the catalog

bdraco:
# What does this implement/fix?

Importing a board whose featured leaf binds a **platform-style bus** produced a
config that would not compile. On the Shelly Pill, `sensor.dallas_temp` depends on
`one_wire`, but the generated board catalog had the sensor with no `one_wire`
entry and no `requires`, so the full-setup wizard landed a config that failed with
`Couldn't find any component that can be used for 'one_wire::OneWireBus'. Are you
missing a hub declaration?`.

The merged Lanbon L9 fix already lifts a bus a featured leaf depends on directly,
but only for **mapping-style** buses (those whose generated catalog category is
`bus`: i2c, spi, uart, modbus). ESPHome's other two buses, `one_wire` and
`canbus`, are `IS_PLATFORM_COMPONENT`; they are declared as `- platform: X` lists
and have no top-level component, so their generated category is the domain name
(`one_wire` / `canbus`) and the lift skipped them.

This generalizes the direct-bus lift to all six known bus types:

- `_BUS_CATEGORIES` + `_is_bus_dep` replace the hardcoded `category == "bus"` gate
  in `_collect_bus_dep_refs`, recognizing both bus shapes.
- `_materialize_bus` gains a platform-style branch: it resolves the
  `<domain>.<platform>` component (e.g. `one_wire.gpio`) from the source block,
  locks its hardware pins, and relaxes the mandatory id (a sole platform bus is
  auto-detected, so the source often omits one). Mapping-style i2c/spi/uart/modbus
  are unchanged.

Regenerated against ESPHome 2026.6.3. The catalog diff is four boards gaining
their one_wire bus, its `requires`, and the bundle reordering: shelly_the_pill,
hankerila_ea8, homemaster_openthermgateway (two one_wire buses, disambiguated by
`one_wire_id`), and sonoff_thr316.

**Related issue or feature (if applicable):**

- Reported on the Shelly Pill (`devices.esphome.io/devices/shelly-the-pill`); no
  separate issue filed.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`

## Frontend coordination

- [x] No frontend change needed

## Checklist

- [x] The code change is tested and works locally.
- [x] Pre-commit hooks pass (`ruff`, `codespell`, yaml/json/python checks).
- [x] Tests have been added or updated under `tests/` where applicable.
[...]

PR #1753, opened 2026-06-29
Lift bus dependencies that have no id in the source config

bdraco:

Follow-up to #1752 (now merged; this is rebased onto `main`). That PR taught the
importer to lift platform-style buses (one_wire/canbus); a full-catalog scan then
showed a second, broader cause of the same "device can't add properly" symptom: a
featured leaf binding a bus whose
source block has **no `id`**. ESPHome lets a sole bus omit its `id` (consumers
auto-detect it), and many boards do (Shelly HT Gen3's `i2c:` has `sda`/`scl` but
no `id`; m5stack/lilygo/lanbon displays; kincony i2c sensors). The old
`_materialize_bus` bailed whenever the bus block had no `id`, so none of these
buses were lifted and their full-setup config failed to compile.

This relaxes the id requirement uniformly: a sole, resolvable bus is lifted with
its pins locked even without an `id`, its local id falling back to
`<domain>_bus`. The two mapping/platform branches collapse into one resolver
since both now treat the id as optional.

The id check was also (accidentally) the only thing stopping `_ensure_buses` from
materializing a hub's non-bus dependency (e.g. `esp32`) as a bus, so this adds the
explicit bus-category guard the id check was standing in for. `_materialize_bus`
now always confirms the resolved component is a bus.

Regenerated against ESPHome 2026.6.3; 17 boards gain their previously-missing
i2c/spi/uart bus. Five kincony boards remain unsatisfied for a different reason
(the source has no bus block, or a `one_wire:` block with no `platform:`); those
are left for a separate edge-case fix.

- Follow-up to #1752.

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1755, opened 2026-06-30
Validate that a board's featured bus dependencies are satisfied

bdraco:

Follow-up to #1752 and #1753 (both merged; this is rebased onto `main`). Those PRs
fixed boards whose imported full-setup config
wouldn't compile because a featured leaf's bus dependency was never lifted into
the catalog. We only found those by hand, because nothing validated that a
board's featured components form a compilable set: `validate_definitions.py`
checks each piece structurally (schema, cross-references, pins) but never that a
featured leaf's bus dependency is actually present.

This adds that check. `_validate_featured_dependencies` flags any featured leaf
(on an imported board) whose catalog `dependencies` include a bus
(i2c/spi/uart/modbus/one_wire/canbus) that no featured component on the board
provides, with a message pointing at the sync-script lift. It is a cheap static
graph check, no `esphome config` compile and no `esphome.components` import, so it
fits the existing pre-commit / CI hook.

Five kincony boards remain knowingly unsatisfied for a different, source-level
reason (the source has no top-level bus block, or a `one_wire:` block with no
`platform:`); they are tracked in an explicit `_UNSATISFIED_BUS_ALLOW_LIST` with
the per-board reason rather than silently skipped, so removing one must be paired
with a fix. Run against the current catalog this passes; drop a board from the
allow list (or regress the sync script) and it fails with the offending leaf.

- Follow-up to #1752 and #1753.

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/constants.py
+++ b/esphome_device_builder/constants.py
@@ -99,3 +99,13 @@ BOARD_PIN_KEYS: frozenset[str] = frozenset(
 # ``source`` block. Shared so the importer (writer) and the loader (reader)
 # can't drift.
 DEVICE_IMPORT_SOURCE_TYPE = "esphome-devices"
+
+# Generated catalog categories for ESPHome's buses (the ``_CATEGORY_OVERRIDES``
+# bus entries in script/sync_components.py). Mapping-style buses (i2c/spi/uart/
+# modbus) collapse to ``"bus"``; platform-style buses (one_wire/canbus) keep
+# their domain name as the category because they are ``IS_PLATFORM_COMPONENT``
+# and have no top-level component, so the dep name itself equals the category.
+# Shared so the importer (script/sync_esphome_devices.py, which lifts buses) and
+# the validator (script/validate_definitions.py, which checks they were lifted)
+# can't drift. Stdlib-only home, so neither script pulls in ``esphome``.
+BUS_CATEGORIES: frozenset[str] = frozenset({"bus", "one_wire", "canbus"})
--- a/script/sync_esphome_devices.py
+++ b/script/sync_esphome_devices.py
@@ -48,6 +48,7 @@ sys.path.insert(0, str(_REPO_ROOT))
 
 from esphome_device_builder.constants import (  # noqa: E402
     BOARD_PIN_KEYS,
+    BUS_CATEGORIES,
     DEVICE_IMPORT_SOURCE_TYPE,
 )
 from esphome_device_builder.helpers.pin_gpio import parse_board_gpio  # noqa: E402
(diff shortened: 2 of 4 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Thanks @esphbot — all three applied in `1a38fcd8`:

1. **Per-board → per-`(board, bus)` allow list.** `_UNSATISFIED_BUS_ALLOW_LIST` is now a set of `(board_id, bus)` tuples, and the check skips only the matching dep instead of early-returning for the whole board. A *different* unsatisfied bus on one of the five kincony boards now still fails. Added `test_featured_dep_allowlist_is_per_bus_not_per_board` (an spi leaf on i2c-waived `kincony_mb` is flagged).

2. **`default_components` folded into the present set.** `_ref_ids()` now collects ids from both `featured_components` and `default_components`, so a bus provided via defaults counts as present. Added `test_featured_dep_satisfied_by_default_component`.

3. **`BUS_CATEGORIES` hoisted to `esphome_device_builder/constants.py`** (the stdlib-only module both scripts already import) and imported in both `sync_esphome_devices.py` and `validate_definitions.py`, so the importer and validator can't drift. No `esphome` import pulled in; catalog regen is byte-identical.

Full suite 5623 passed; `validate_definitions.py` stays clean on all 525 boards.

bdraco:
Thanks @copilot — addressed in `ae5b398e`:

**1. `TypeError` on a non-string `component_id`** — already fixed in the prior commit: the present-set is now built by a `_ref_ids()` helper that only adds values passing `isinstance(..., str)`, so a list/dict `component_id` is skipped, not hashed into a set. Added `test_featured_dep_non_string_component_id_does_not_crash` to pin it.

**2. `"." in cid` skips top-level components** — good catch on the gap, but I tested removing the guard and it flags **6 boards as false positives**, all `sn74hc595` (es32a08, esp32_relay_x16, kincony_kc868_a256, tuya_wifi_sprinkler_controller, usr_eg118). `sn74hc595` is **dual-mode**: it declares `dependencies: ["spi"]` but bit-bangs over GPIO (`data_pin`/`clock_pin`/`latch_pin`) on all six — no spi bus needed. The catalog `dependencies` over-declares the conditional bus. The dotted-leaf scoping is deliberate: platform leaves (sensors/displays/touchscreens) bind a bus *unconditionally* (it's their only connection), whereas bare top-level components are either buses themselves (no bus dep) or dual-mode hubs with conditional deps, and their bus is lifted by the importer's hub passes. Expanded the code comment to spell this out.

**3. Regression test for the top-level case** — added `test_featured_dep_dual_mode_top_level_hub_not_flagged`, which pins the *intended* behavior (an `sn74hc595` over GPIO is not flagged for spi) so the scoping isn't later 'fixed' into the false positives above.

`validate_definitions.py` stays clean on all 525 boards; 32 tests in `test_validate_featured.py`.

esphbot (approved):
No blocking issues found.

PR #1756, opened 2026-06-30
Infer one_wire platform gpio for a platform-less source block

bdraco:

Two imported boards (`kincony_kc868_aio`, `kincony_kc868_uair`) ship a
`dallas_temp` sensor whose `one_wire` bus was never lifted into the catalog, so
the full-setup config failed to compile. They were allow-listed by #1755 with the
note "one_wire: block omits platform:".

Root cause: their source (verified on the live devices.esphome.io pages) writes
the older `one_wire:\n  - pin: GPIO27` form with no `platform:` key. Modern
ESPHome requires `one_wire: - platform: gpio\n    pin: X`, so `_materialize_bus`
took the platform-style branch, found no `platform`, and bailed.

Fix: infer `platform: gpio` for a `one_wire` block that has a `pin` but no
`platform`. `gpio` is the only pin-driven one_wire platform (`ds2484` is
i2c-bridged and carries no raw `pin`), so the inference is unambiguous. Pulled the
platform resolution into a small `_block_platform` helper to keep
`_materialize_bus` under the complexity limit.

The two boards are removed from the guard's `_UNSATISFIED_BUS_ALLOW_LIST`; the
guard then verifies the bus is actually lifted (validation stays at 0 errors).
Regenerated against ESPHome 2026.6.3 — the diff is exactly those two boards gaining
a `one_wire.gpio` entry with its `pin` locked and the dallas sensor's `requires`.

- Follow-up to #1755 (fixes 2 of its 5 allow-listed boards).

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1757, opened 2026-06-30
Recover a consumer's bus_id from its source block

bdraco:

The last 3 boards allow-listed by #1755 — `kincony_ag8` (uart), `kincony_kc868_e8t`
(uart), `kincony_mb` (i2c) — shipped a full-setup config that wouldn't compile
because their bus was never lifted. The allow-list comments claimed "source has no
top-level block", but the live devices.esphome.io pages (and the cached source)
show the bus **is** there, with explicit ids, in the same code block as the leaf —
so that was a misdiagnosis. The real cause is an importer bug.

Each board has multiple same-platform leaves with **no `id:`** on the consumer
block (2 `switch.uart`, 2 `sensor.bl0939`, 16 `sensor.ina226`), each binding a bus
via `uart_id:`/`i2c_id:`. That ref is dropped from the finalized entry (the bus
isn't a featured component yet), so `_collect_bus_dep_refs` re-finds the source
block via `_find_consumer_block` — which filters by platform, gets N id-less
blocks, can't disambiguate, and returns `None`. With the ref lost and multiple
buses present, `_materialize_bus` can't pick one, so the bus never lifts.

Fix: `_Candidate` already retains the source block and each finalized entry's `id`
equals its unique `local_id`, so stash the source block on the entry at finalize,
read it back in `_collect_bus_dep_refs` (falling back to `_find_consumer_block`
for entries built outside the extractor, e.g. unit tests), and strip the transient
key in `_make_record` before serialization. Each consumer's `<bus>_id` is now
recovered directly: `kincony_mb`'s 16 ina226 split correctly across `bus_a`/`bus_b`,
ag8/e8t wire each leaf to its own uart.

This empties the guard's `_UNSATISFIED_BUS_ALLOW_LIST` — every imported board's
featured bus dependencies now lift, and `validate_definitions.py` proves it (0
errors). Regenerated against ESPHome 2026.6.3; the diff is exactly the 3 boards
gaining their uart/i2c entries + `requires`.

- Follow-up to #1755; finishes its 5 allow-listed boards (#1756 did the other 2).

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

bdraco:
Thanks @esphbot — applied your suggestion 1 (read-and-discard) in `b5fcf678`.

`_collect_bus_dep_refs` is the sole reader of `_source_block`, so it now `pop`s the key at the top of its loop (before any `continue`) and uses the popped value, and the separate strip loop in `_make_record` is gone. The transient key is removed at the moment it's consumed, for every entry, independent of how the entries are later serialized — closing the 'future serialization path leaks it' risk. Updated the `_finalize_entry` comment to point at the new consumption site.

Behavior-preserving: the regenerated catalog is byte-identical, no `_source_block` in any board body, validation stays at 0 errors, full suite green.

esphbot (approved):
No blocking issues found.

PR #1766, opened 2026-06-30
Serialise float and lambda YAML values in /json-config

### LINKED ISSUE #1765: [Bug] YAML configs with floats break /json-config endpoint and encryption key detection in Home Assistant
### What happened?

When adding a discovered ESPHome device in Home Assistant, the configuration flow attempts to fetch the encryption key from ESPHome through the `/json-config` endpoint. When a YAML config contains a float, the JSON encoder raises an exception, and Home Assistant falls back to prompting for an encryption key.
 
- Create a YAML config with a float, e.g. `test.yaml`:
  ```yaml
  esphome:
    name: test
  
  esp32:
    variant: esp32c3
  
  sensor:
    - platform: adc
      pin: GPIO01
      id: test
      filters:
        - delta: 0.1
  ```
- Start ESPHome Dashboard: `docker run -p 6052:6052 -v "$PWD":/config ghcr.io/esphome/esphome`
- Get the JSON representation through the `/json-config` endpoint: `curl 'http://localhost:6052/json-config?configuration=test.yaml'`
- Server responds with `500 Internal Server Error`
- ESPHome raises an exception:
  ```python
  ERROR (MainThread) [aiohttp.server] Error handling request from 192.168.65.1
  Traceback (most recent call last):
    File "/usr/local/lib/python3.13/site-packages/aiohttp/web_protocol.py", line 575, in _handle_request
      resp = await request_handler(request)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.13/site-packages/aiohttp/web_app.py", line 559, in _handle
      return await handler(request)
             ^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.13/site-packages/aiohttp/web_middlewares.py", line 117, in impl
      return await handler(request)
             ^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.13/site-packages/esphome_device_builder/helpers/json.py", line 98, in cors_middleware
      resp = web.Response() if request.method == "OPTIONS" else await handler(request)
                                                                ^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.13/site-packages/esphome_device_builder/helpers/auth.py", line 350, in auth_middleware
      return await handler(request)
             ^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.13/site-packages/esphome_device_builder/api/legacy.py", line 370, in legacy_json_config
      return web.json_response(config, dumps=dumps_str_non_str_keys)
             ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.13/site-packages/aiohttp/web_response.py", line 872, in json_response
      text = dumps(data)
    File "/usr/local/lib/python3.13/site-packages/esphome_device_builder/helpers/json.py", line 62, in dumps_str_non_str_keys
      return orjson.dumps(obj, option=orjson.OPT_NON_STR_KEYS).decode()
             ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  TypeError: Type is not JSON serializable: EFloat
  ```
- Comment out `delta: 0.1` in the YAML config and fetch the JSON again
- Server responds with the expected JSON:
  ```json
  {"esphome":{"name":"test"},"esp32":{"variant":"esp32c3"},"sensor":[{"platform":"adc","pin":"GPIO01","id":"test","filters":null}]}
  ```

### Dashboard version

esphome-device-builder 1.0.21 (esphome 2026.6.3)

### How did you install it?

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:
## What does this implement/fix?

`GET /json-config?configuration=<file>` returned **500** whenever the YAML
config contained a float (e.g. a `filters: - delta: 0.1`) or a `!lambda`, with
`TypeError: Type is not JSON serializable: EFloat` / `... Lambda`. Home
Assistant's ESPHome config flow uses this endpoint to auto-detect the API
encryption key, so the 500 dropped HA back to manually prompting the user.

**Root cause:** `api/legacy.py:legacy_json_config` parses with ESPHome's
`yaml_util.load_yaml`, which wraps every scalar in a source-position subclass
(`EInt`/`EFloat`/`EStr`/`EDict`/`EList`) and turns `!lambda` into a
`esphome.core.Lambda`. It then serialises via `dumps_str_non_str_keys` (orjson).
orjson serialises `str`/`int`/`dict`/`list` subclasses **natively but not
`float` subclasses, and not `Lambda`** — so `EStr` keys and `EInt` values pass,
but any `EFloat` value or `!lambda` raises.

The legacy Tornado dashboard never hit this: it shells out to
`esphome config --show-secrets`, re-parses the subprocess stdout with a *plain*
pyyaml loader (`SafeLoaderIgnoreUnknown`) — which yields vanilla `float`/`str`
(no E-types) and renders unknown tags like `!lambda` as plain strings via
`ignore_unknown` — then dumps with stdlib `json`. Nothing non-JSON-native ever
reaches the serialiser. Our lighter raw-load + orjson path (no subprocess) is
what surfaces these types; the raw load is intentional, so the fix stays on it.

**Fix:** a `_json_default` orjson hook that downcasts a `float` subclass to
plain `float` and renders `Lambda` as its source string, wired into
`dumps_str_non_str_keys` (the only caller feeding raw E-typed YAML). `default`
fires only for types orjson can't already handle, so `EInt`/`EStr`/`EDict`/
`EList` are untouched; genuinely unknown types still raise, keeping orjson
strict for every other call site. `esphome.core` is already in the eager import
graph, so the top-level `Lambda` import adds no cold-start cost.

## Types of changes

## Frontend coordination

## Checklist

Fixes #1765

### DIFF
--- a/esphome_device_builder/helpers/json.py
+++ b/esphome_device_builder/helpers/json.py
@@ -15,6 +15,7 @@ from typing import Any
 
 import orjson
 from aiohttp import web
+from esphome.core import Lambda
 
 from .origin import request_origin_allowed
 
@@ -38,6 +39,21 @@ def dumps(obj: Any) -> bytes:
     return orjson.dumps(obj)
 
 
+def _json_default(obj: Any) -> Any:
+    """Orjson fallback for ESPHome's ``yaml_util`` types orjson rejects.
+
+    orjson serialises ``str`` / ``int`` / ``dict`` / ``list`` subclasses
+    natively but rejects ``float`` subclasses (``EFloat``) and ``Lambda``,
+    so a config with a float or a ``!lambda`` would 500 ``/json-config``.
+    Downcast ``EFloat`` to ``float`` and render ``Lambda`` as its source.
+    """
+    if isinstance(obj, float):
+        return float(obj)
+    if isinstance(obj, Lambda):
+        return str(obj)
+    raise TypeError
+
+
 def dumps_str_non_str_keys(obj: Any) -> str:
     """
     Serialise *obj* allowing dict keys whose type isn't *exactly* ``str``.
@@ -48,7 +64,9 @@ def dumps_str_non_str_keys(obj: Any) -> str:
     Dict key must be str``. ESPHome's ``yaml_util`` returns dicts
     whose keys are ``EStr`` (a ``str`` subclass that carries
     source-position info), which is what the legacy
-    ``/json-config`` endpoint feeds in.
+    ``/json-config`` endpoint feeds in. The ``_json_default`` hook
+    additionally tolerates the ``yaml_util`` value types orjson rejects
+    — ``EFloat`` (a ``float`` subclass) and ``Lambda`` (#1765).
 
     Use this helper for that endpoint (and only there); the strict
     default of ``dumps`` still catches the more common bug shape —
@@ -59,7 +77,7 @@ def dumps_str_non_str_keys(obj: Any) -> str:
     ``web.json_response(dumps=...)`` (which expects a ``str``-
     returning callable, like ``dumps_str``).
     """
-    return orjson.dumps(obj, option=orjson.OPT_NON_STR_KEYS).decode()
+    return orjson.dumps(obj, option=orjson.OPT_NON_STR_KEYS, default=_json_default).decode()
 
 
 def dumps_str(obj: Any) -> str:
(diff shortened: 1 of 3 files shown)


bdraco:
this is going to be a bit of whack-a-mole so lets go back to the old slower way to do this

PR #1767, opened 2026-06-30
Resolve /json-config through esphome config

### LINKED ISSUE #1765: [Bug] YAML configs with floats break /json-config endpoint and encryption key detection in Home Assistant

bdraco:

The `/json-config` endpoint did a raw `yaml_util.load_yaml`. That left `${substitutions}` unexpanded, never merged `packages:`, and handed orjson ESPHome wrapper types it cannot serialize (EFloat, Lambda, IncludeFile), so a config containing a float returned a 500. Home Assistant reads this endpoint to detect the API encryption key, so the failure dropped HA back to prompting for the key (#1765).

It now resolves the config the same way the legacy dashboard and `devices/get_api_key` already do; it runs `esphome config --show-secrets` and serializes the plain validated output. Substitutions, packages, includes and secrets all resolve, and the three wrapper types are gone because the validated output carries none of them. The subprocess and parse are factored into a shared `run_esphome_config` helper, so the key fallback reuses the same code path. Status codes now match upstream; 404 for a missing file, 422 for an invalid config, 403 for a traversal.

- fixes https://github.com/esphome/device-builder/issues/1765

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -475,6 +475,6 @@ For Home Assistant ESPHome integration backward compat only.
 |----------|-------------|
 | `GET /devices` | List devices |
 | `GET /ping` | Online-status map `{<config>.yaml: true\|false\|null}` (third-party widgets, e.g. homepage) |
-| `GET /json-config?configuration=...` | Get parsed YAML as JSON |
+| `GET /json-config?configuration=...` | Fully-resolved config (substitutions/packages/includes/secrets) as JSON, via `esphome config`; 404 missing, 422 invalid, 503 unavailable, 403 traversal |
 | `GET /compile` (WebSocket) | Compile via spawn protocol |
 | `GET /upload` (WebSocket) | Upload via spawn protocol |
--- a/esphome_device_builder/api/legacy.py
+++ b/esphome_device_builder/api/legacy.py
@@ -362,12 +362,12 @@ def create_legacy_routes() -> web.RouteTableDef:
         if not esphome_cmd:
             return json_response({"error": "esphome unavailable"}, status=500)
 
-        # Generic 422 body on failure — ``esphome config`` stderr carries
-        # resolved secrets under ``--show-secrets`` and must not echo back.
-        _rc, config, _stderr = await run_esphome_config(esphome_cmd, config_path)
+        # Generic 422 body — an invalid config's error can carry resolved
+        # secrets (``--show-secrets``), so it never reaches the response.
+        config = await run_esphome_config(esphome_cmd, config_path)
         if config is None:
             return json_response({"error": "Configuration is invalid"}, status=422)
-        return web.json_response(config, dumps=dumps_str)
+        return json_response(config)
 
     @routes.get("/compile")
     async def legacy_compile(request: web.Request) -> web.WebSocketResponse:
--- a/esphome_device_builder/api/legacy.py
+++ b/esphome_device_builder/api/legacy.py
@@ -34,7 +34,7 @@ import aiohttp
 from aiohttp import web
 
 from ..helpers.api import CommandError
-from ..helpers.device_yaml import run_esphome_config
+from ..helpers.device_yaml import EsphomeConfigUnavailableError, run_esphome_config
 from ..helpers.event_bus import Event, StreamControls, stream_events
 from ..helpers.json import (
     JSONDecodeError,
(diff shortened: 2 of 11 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Thanks; addressed in d357dba.

Log levels: spawn failure and timeout now log at WARNING (genuine infrastructure faults, and those messages carry no secrets); a non-zero exit stays at DEBUG, since for this endpoint that is the expected invalid-config path the caller already surfaces as a 422, and warning on every mid-edit config would be noisy.

Non-mapping output: a zero-exit run whose output is not a mapping now logs at DEBUG, so it is observable rather than a silent None.

Non-scalar tags: the unknown-tag handler now returns the bare tag for a non-scalar node instead of stringifying its list of node pairs; a regression test covers it.

Unbounded subprocess spawning: this reuses the existing devices/get_api_key fallback already on main, the endpoint sits behind the REST auth gate, and HA calls it once per device add, so I left a concurrency bound out of this change; happy to add a shared semaphore across both spawn sites as a follow up if you want it.

esphbot (approved):
No blocking issues found.

bdraco:
Thanks; addressed in be83fe6.

Infra vs invalid config: run_esphome_config now raises EsphomeConfigUnavailableError on a spawn failure, a timeout, or a signal kill (negative return code), and returns None only when the config actually ran and failed validation. /json-config maps the exception to 503 (retryable) and a None to 422, so a transient backend fault no longer tells HA the config is permanently wrong. The non-zero exit case stays at DEBUG since that is the expected invalid-config path, while spawn/timeout/signal now log at WARNING.

get_api_key path: I left this returning "" for both an infra fault and a keyless config, on purpose. The "" is only a UI "open the editor and check" hint; it never drives a connection. The actual connection path is get_api_connection (in-process, raises on failure), and api_info.py already refuses a plaintext connect to a device whose config declares encryption but whose key did not resolve, so there is no plaintext-downgrade risk here.

docs/API.md: updated the /json-config row to describe the fully-resolved config and the 404/422/503/403 contract.

CancelledError vs timeout asymmetry: intentional, now commented. The timeout path reaps with proc.wait(); the cancel path deliberately does not, because awaiting wait() there would suppress or delay the cancellation propagation (the SIGKILL'd child is reaped by asyncio's watcher). This matches helpers.subprocess.run_subprocess_capture.

Also folded in from a type check: the handler moved to a module-level helper that guards a None devices controller (clean 500 instead of an AttributeError) and keeps create_legacy_routes under the complexity cap.

esphbot (approved):
No blocking issues found.

PR #1773, opened 2026-06-30
Drain interface monitor via gather so caller cancellation isn't swallowed

bdraco:

Followup to #1770. `MdnsSource.close_zeroconf` cancelled the interface-monitor task and awaited it inside a bare `try: await task; except asyncio.CancelledError: pass`. That swallows a `CancelledError` even when it is aimed at `close_zeroconf` itself, not the child finishing; a cancellation of the shutdown coroutine would be silently eaten, breaking cooperative cancellation.

Switch the drain to `asyncio.gather(task, return_exceptions=True)`, the same idiom `DeviceStateMonitor.stop` already uses. It captures the child's `CancelledError`, and any crash, into the result list without intercepting a cancellation directed at the current coroutine; a returned crash is still debug-logged so a dead monitor does not abort shutdown before zeroconf closes.

No test changes; the existing `test_close_zeroconf_swallows_crashed_interface_monitor` and `test_start_spawns_interface_monitor_and_stop_cancels_it` already exercise the crash and normal-cancel drain paths through the new code.

- follows up #1770

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/_device_state_monitor/mdns.py
+++ b/esphome_device_builder/controllers/_device_state_monitor/mdns.py
@@ -140,13 +140,13 @@ class MdnsSource:
         # Stop the interface monitor first so it can't reconcile a closing instance.
         if self._interface_monitor_task is not None:
             self._interface_monitor_task.cancel()
-            try:
-                await self._interface_monitor_task
-            except asyncio.CancelledError:
-                pass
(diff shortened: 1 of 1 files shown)


esphbot (approved):
No blocking issues found.

PR #1781, opened 2026-07-01
Fix recommended component add for onboard-ethernet boards

bdraco:

Adding a KinCony KC868 board (A64, A128, and the other onboard ethernet boards) and choosing "set up with all recommended components" popped a warning that one recommended component couldn't be added, yet nothing was actually missing. Two backend issues combined.

First, `add_component`'s required-field gate is stricter than ESPHome for type-conditional fields. The ethernet schema marks `clk` required for a LAN8720 (RMII) PHY, but the board preset sets the equivalent `clk_mode`, so the gate raised "Missing required field: clk" and the recommended add was skipped. `create` never runs that gate and ESPHome accepts `clk_mode`, which is why the device still built; the same error also blocked adding "Onboard Ethernet" from the featured catalog by hand. Now curated featured presets are trusted; the required-field gate applies only to user-driven form adds, and ESPHome validation at compile stays the real arbiter.

Second, ethernet is a singleton top level block, so re-adding it (create injects it as the network provider, then the recommended bundle re-lists it) would have appended a duplicate `ethernet:` key. `merge_component_yaml` now treats a singleton whose block already exists as a no op, and still appends when it is absent, so applying the bundle later to a device without ethernet keeps working. This also removes a latent duplicate-key bug for any singleton (ethernet, wifi, web_server).

Together the recommended ethernet add is now a clean no op and the warning no longer fires; verified end to end against a live KC868-A128. The warning text itself is emitted by the frontend and is being improved separately to name the component.

- reported for KinCony KC868-A64 / KC868-A128; affects 59 onboard ethernet boards, no separate issue filed

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1783, opened 2026-07-01
Fix featured recommended-add: infer requires and dedupe re-added components

bdraco:

Two fixes for the featured "recommended component" add flow, both surfaced by the Apollo ESK-1 starter kit.

First, a featured component that references a sibling component's id (the RTTTL player points at its buzzer output, a sensor points at its i2c bus) generated YAML with an undefined id when added on its own, so it failed to validate with "Couldn't find ID". The featured `requires` list already makes the frontend add the prerequisite first, but the manifests hand declared it in only 114 of 610 such cross references, so 166 boards were affected. This infers `requires` at sync time from those id references and flattens them transitively, then regenerates the catalog so every affected board picks up its missing prerequisites. A committed catalog test locks the invariant so a new manifest or import can't reintroduce the bug.

Second, adding a bundle that re-lists a component already in the config (the RGB + Buzzer module re-adds the buzzer output and rtttl player) spliced a duplicate entry, so ESPHome rejected it with "ID redefined". Only singletons were idempotent before; now the merge skips a block whose id is already defined, so a re-add is a no-op. Ids are config global, so the check spans domains.

- reported for the Apollo ESK-1 starter kit (RTTTL Player, RGB + Buzzer bundle); no filed issue

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1790, opened 2026-07-02
Reuse the drain_tasks helper for cancel-and-drain teardowns

bdraco:

Follow-up to the version-history PR, which moved `drain_tasks` into `helpers/async_.py` as a general cancel-and-drain helper. This reuses it at the hand-rolled sites that do the same thing, so the "cancel then gather, or cancel then suppress and await" idiom lives in one place.

Converted (six sites): teardowns that cancel task(s) then `gather(*tasks, return_exceptions=True)` (`DeviceBuilder.stop_network`'s background set, `WebSocketManager.cleanup`, the firmware queue's finally block, the peer-link client heartbeat), and single-task `stop()` paths that cancel then await under `contextlib.suppress(asyncio.CancelledError)` (`EditorController.stop`, `DeviceMqttMonitor.stop`, plus `DeviceBuilder`'s background probe task). All map onto `drain_tasks`, which snapshots the tasks, cancels them, and gathers with `return_exceptions=True`. The now-unused `contextlib` imports are removed.

The four `gather`-based sites are exact equivalents (default `log_exceptions=False`, same silent-swallow as the bare gather they replace). The three `suppress`-based sites aren't strictly bit-exact: the old `suppress(CancelledError)` only caught cancellation and let a non-cancel teardown exception propagate out of `stop()`, so those three pass `log_exceptions=True` to surface such an exception at WARNING rather than dropping it silently (a normal `CancelledError` is a `BaseException`, so it stays unlogged). These `stop()` paths aren't reached under their own cancellation the way the deliberately-skipped `finally` sites are, so the outer-cancel-propagation difference doesn't arise.

Left alone on purpose: the peer-link session and reachability-stream teardowns keep their `suppress(asyncio.CancelledError)` because it runs in a `finally` reached while the enclosing task is itself being cancelled and deliberately swallows that cancellation so the trailing cleanup (e.g. `unregister_peer_link_session`) still runs; `drain_tasks` gathers, which would propagate the cancel and skip that cleanup. Also skipped: sites that gather bare coroutines rather than cancel tasks (`close_all_websockets`, the ping fan-out), the monitor stop that wraps the gather in `wait_for` for a drain timeout, the mDNS close that inspects the gather result, the wake-worker and dashboard-advertise drains that log at ERROR / DEBUG, the sync-only `cancel()` with no await, and the `TimerHandle` cancels.

- follow-up to the `drain_tasks` helper added in #1785; no filed issue

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/api/ws.py
+++ b/esphome_device_builder/api/ws.py
@@ -17,7 +17,7 @@ from esphome.const import __version__ as esphome_version
 from ..constants import __version__
 from ..controllers.auth import AuthError
 from ..helpers.api import CommandError
-from ..helpers.async_ import create_eager_task
+from ..helpers.async_ import create_eager_task, drain_tasks
 from ..helpers.auth import extract_bearer_token
 from ..helpers.event_bus import StreamBackpressureError
 from ..helpers.json import JSONDecodeError, dumps_str, loads
@@ -193,10 +193,7 @@ class WebSocketClient:
 
     async def cleanup(self) -> None:
         """Cancel all pending tasks."""
-        for task in self._tasks:
-            task.cancel()
-        if self._tasks:
-            await asyncio.gather(*self._tasks, return_exceptions=True)
+        await drain_tasks(self._tasks)
 
     async def _handle_command(self, raw: dict[str, Any]) -> None:
         """Parse and dispatch a command."""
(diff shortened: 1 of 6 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Addressed in 8173cad2. You are right that the three suppress -> drain_tasks conversions were not bit-exact. Rather than only note it, I made them pass log_exceptions=True so a non-cancel teardown exception (which the old suppress(CancelledError) would have propagated) is surfaced at WARNING instead of being silently swallowed; a normal CancelledError is a BaseException so it stays unlogged. The four gather-based sites keep the default log_exceptions=False since they were already silent, so they remain exact. Updated the PR body to spell out the gather-exact vs suppress-logs-instead distinction.

esphbot (approved):
No blocking issues found.

PR #1791, opened 2026-07-02
Add run_in_executor helper and use it across the codebase

bdraco:

Follow-up to the `drain_tasks` extraction. Adds a small `run_in_executor(fn, *args)` helper to `helpers/async_.py` and uses it at the ~90 sites across 44 files that hand-roll the default-executor idiom `loop = asyncio.get_running_loop(); await loop.run_in_executor(None, fn, ...)`.

```python
async def run_in_executor[T](fn: Callable[..., T], *args: Any) -> T:
    return await get_running_loop().run_in_executor(None, fn, *args)
```

It mirrors Home Assistant's `async_add_executor_job` shape: callers stop re-deriving the loop for the common block-shift-to-a-thread case, and a `functools.partial` / lambda covers the keyword case. Where a scope only bound `loop` for the executor call, that binding is removed; where `loop` is still used (`call_later`, `create_task`, ...), it stays. `version_history`'s private `_in_executor` proof-of-concept is deleted and its callers point at the shared helper.

No behaviour change: the helper always uses the default executor (every converted site already passed `None`), and only the loop-derivation boilerplate is removed. The one `run_in_executor` that targets a *custom* executor (`helpers/lazy_module.py`'s dedicated import pool) is deliberately left alone. Now-dead `import asyncio` lines and a couple of orphaned `loop` parameters left by the conversion are cleaned up too.

- follow-up to the `drain_tasks` helper (#1790); no filed issue

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/_remote_build_lifecycle.py
+++ b/esphome_device_builder/_remote_build_lifecycle.py
@@ -16,6 +16,7 @@ from .controllers.config import (
     load_remote_build_settings,
 )
 from .controllers.remote_build.peer_link import PEER_LINK_PATH, make_peer_link_handler
+from .helpers.async_ import run_in_executor
 from .helpers.network_interfaces import (
     bind_available_port,
     ensure_single_host_for_ephemeral_port,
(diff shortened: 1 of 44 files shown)


esphbot (approved):
No blocking issues found.

PR #1792, opened 2026-07-02
Unify the remote-build store codecs behind a factory

bdraco:

Small follow-up from the duplication sweep. The two remote-build per-file stores each hand-rolled an encode/decode pair that was identical bar the model class and the log label: `encode_pairings`/`decode_pairings` (offloader pairings) and `encode_peers`/`decode_peers` (receiver peers). Both decoders soft-recover to an empty model on a corrupt blob.

This folds them into one local factory `_soft_recover_codec(model, label)` in `_storage_codecs.py`; the four exported names stay importable (they're now the factory's return values), the soft-recover posture and the exact log messages are unchanged. No behaviour change.

Scoped deliberately: the other model-backed stores are left bespoke because they differ in ways a shared factory would only obscure. `PreferencesStore` raises on a corrupt file (the caller preserves it for recovery) and encodes with `dumps_indent`; `DeviceMetadataStore` decodes to a filtered plain dict (not a model) and returns `{}` on corruption. Only the remote-build pair is genuinely the same shape, so the factory stays local to that module rather than becoming a cross-store helper.

- follow-up to the helper-extraction sweep (`drain_tasks` #1790, `run_in_executor` #1791); no filed issue

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/remote_build/_storage_codecs.py
+++ b/esphome_device_builder/controllers/remote_build/_storage_codecs.py
@@ -15,10 +15,11 @@ entirely) is load-bearing enough to live in its own seam.
 from __future__ import annotations
 
 import logging
+from collections.abc import Callable
 
 from ...helpers.json import dumps as json_dumps
 from ...helpers.json import loads as json_loads
-from ...models import OffloaderRemoteBuildSettings, ReceiverPeers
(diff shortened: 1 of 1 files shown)


esphbot (approved):
No blocking issues found.

PR #1795, opened 2026-07-02
Emit wifi for boards with an esp32_hosted radio provider

bdraco:

Groundwork for the first boards whose Wi-Fi comes from an ESP-Hosted co-processor. The generator now emits the wifi block only when the chip actually has a radio, either natively or because a component in the new WIFI_RADIO_PROVIDER_COMPONENT_IDS set (esp32_hosted) arrives through defaults; a wifi claim without the radio falls back to the no network stub instead of a config ESPHome rejects on P4/H2. The manifest validator enforces the matching invariant, a no native wifi variant claiming wifi must carry a radio provider in default_components. Also emits enable_idf_experimental_features for 32MB flash boards since ESPHome refuses ota without it, and the slow E2E board test now resolves each board's default_components from disk so CI validates the same YAML a real create produces.

- none

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

PR #1800, opened 2026-07-02
Lock the apollo-esk-1 accessory power pin so the editor renders it

bdraco:

On the Apollo ESK-1 the auto added accessory power switch showed an empty Pin field in the structured editor. The manifest preset carried pin value 4 without locked: true, GPIO4 is reserved in the board pin map, and the editor disables reserved pin options unless the board's locked_pins rescues them, so the selected option was disabled and the select rendered blank. This locks the preset so sync stamps locked_pins and the field renders as GPIO4 with its ACCESSORY PWR label, verified in the live dashboard. validate_definitions now rejects any pin preset whose value lands on an available: false pin without locked: true; an audit of every manifest found this was the only offender, so the guard lands clean. A companion frontend PR keeps a field's current value selectable even when it sits on a reserved pin the board did not lock to that section, covering hand edited YAML.

- none

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1056

## Checklist


esphbot (approved):
No blocking issues found.

bdraco:
Moved the reserved-lock check below the is_imported skip in 86eeb565, so it now polices hand-authored pin maps only; the imported path synthesizes every pin as available: false and locks every pin preset, so the check carried no signal there and depended on the cross-script invariant staying true. Added a test pinning the exemption (test_field_preset_imported_skips_reserved_lock_check).

esphbot (approved):
No blocking issues found.

PR #1804, opened 2026-07-03
Mark spi as multi_conf so boards can add a second SPI bus

bdraco:

ESPHome's `spi:` accepts a **list** of buses (`cv.ensure_list` in its `CONFIG_SCHEMA`),
but — unlike `i2c` and `uart` — it does **not** set the component-registry `MULTI_CONF`
flag. Our catalog derives `multi_conf` straight from that flag, so `spi` landed as
`multi_conf: false`. The YAML merge (`helpers/yaml/component.py`) then treats a second
top-level `spi:` as a **no-op** and silently drops the new bus.

That breaks any board needing two SPI buses — surfaced by the ESP32-2432S028R "Cheap
Yellow Display", whose ILI9341 display (CLK14/MOSI13) and XPT2046 touch (CLK25/MOSI32/
MISO39) sit on separate buses: the touchscreen ended up referencing a `touch_spi` bus
that never got emitted.

Fix: stamp `spi` `multi_conf: true` at the introspection source in
`script/sync_components.py` via a small curated set (`_LIST_SCHEMA_MULTI_CONF`) for
list-schema components ESPHome doesn't flag, so `spi` behaves exactly like `i2c`/`uart`
through the whole pipeline (including the reference-advanced pass). The merge already
renders `multi_conf` components as a YAML list and normalises an existing mapping-form
block, so the second bus now splices as a list item — no merge changes needed. Regenerated
`components/spi.json` + the `spi` index entry accordingly.

Existing single-SPI boards now emit list-form (`spi:\n  - clk_pin: …`), which ESPHome
accepts identically; verified a two-bus display+touch config passes `esphome config`.

- n/a (prerequisite for the CYD board catalog entry)

## Types of changes

## Frontend coordination

## Checklist

- [x] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (regenerated via `script/sync_components.py`, scoped to `spi`).


esphbot (approved):
No blocking issues found.

PR #1815, opened 2026-07-03
Add LILYGO T4-S3 curated board

bdraco:

Adds a curated board manifest for the LILYGO T4-S3, an ESP32-S3 board with a 2.41" 450x600 RM690B0 AMOLED panel on a quad-SPI bus. It features the quad SPI bus, the `mipi_spi` display (model `T4-S3`), and the BOOT button, with a one-click Display bundle.

The panel is self-emissive, so there is no backlight output; GPIO9 enables the panel power rail and is supplied by the model. The board's CST226SE touch panel and PMU share an I2C bus (GPIO6/7), left free to assign. `enable_pin` is intentionally not locked in the manifest because it is an `ensure_list` field, so the frontend seeds it as a list and a scalar lock would reject the add; the model fills it.

Pins were verified against three sources: the esphome `mipi_spi` model, the LILYGO AMOLED-Series library board struct, and the dobodu RM690B0 T4-S3 micropython driver; all agree the QSPI bus is clk 15, data 14/10/16/12, with cs 11, reset 13, enable 9, and no DC pin. A generated Display bundle config passes `esphome config`, and `validate_definitions` reports 552 boards OK.

- N/A

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
No blocking issues found.

Issue #1812, opened 2026-07-03
'Rename Hostname' does not use remote build server


codersaur:
### The problem

I've configured my ESPHome instance to use a remote build server. However, when I've used the 'Rename hostname' option it has built locally and not sent to the remote build server.

### Which version of ESPHome has the issue?

2026.6.2

### What type of installation are you using?

Docker

### What platform are you using?

ESP32

### YAML Config

```yaml

```

### Anything in the logs that might be useful for us?

```txt

```

bdraco:
Right now this calls `esphome rename`. We would need to lift all that functionality into the builder so this isn't a bug, its an epic feature request.

bdraco:
Confirmed. The rename ran as a single fused `esphome rename` subprocess, which does the compile internally, and the remote build pool only dispatches compile jobs, so a rename structurally could not use a build server. #1819 decomposes the rename into a compile of the renamed YAML (remote eligible, same scheduling as an install) plus a dependent flash of the old device address that swaps the files on success; esphome/device-builder-frontend#1064 is the companion so the dialog follows the compile into the flash.

PR #1817, opened 2026-07-03
Extract shared rename rewrite and address-resolution helpers

bdraco:

Behavior free prep for decomposing the rename flow so it can use a remote build server (#1812). The rename job today is a single fused esphome rename subprocess, which is why it always compiles locally; splitting it into a compile plus a dependent flash needs a few helpers to live where both the devices and firmware controllers can reach them. This extracts the esphome.name retargetability check and rewrite from _config_only_rename into helpers.yaml.rewrite_rename_content, moves default_mdns_address to helpers.hostname so the firmware package can use it without importing the devices package, adds an exclude_job_ids parameter to check_rename_lock so a future two job chain can be checked without self rejecting, and adds controllers/firmware/rename_flow.py with the old device OTA address resolution (StorageJSON address first, then the scanner's hostname or IP, then the mDNS default). No behavior changes; the config only rename path produces the same output and errors as before.

- prep for https://github.com/esphome/device-builder/issues/1812

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/devices/mutations_create.py
+++ b/esphome_device_builder/controllers/devices/mutations_create.py
@@ -11,6 +11,7 @@ from esphome.storage_json import StorageJSON
 from ...helpers.api import CommandError
 from ...helpers.async_ import run_in_executor
 from ...helpers.device_yaml import parse_platform_from_yaml
+from ...helpers.hostname import default_mdns_address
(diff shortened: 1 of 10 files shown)


esphbot (approved):
No blocking issues found.

PR #1819, opened 2026-07-03
Decompose rename into a remote-eligible compile plus flash-and-swap tail

### LINKED ISSUE #1812: 'Rename Hostname' does not use remote build server

bdraco:

Rename hostname always compiled locally because it was one fused esphome rename subprocess, and the remote dispatch pool only routes COMPILE jobs. This decomposes the rename into the same chain shape installs use: the dashboard rewrites esphome.name itself, writes the new YAML up-front, and queues a COMPILE of it (remote eligible via the build scheduler, so a paired build server picks it up) plus a dependent RENAME tail that runs esphome upload against the old device address on the upload lane and swaps the files on success. Any other tail terminal reverts the chain by deleting the new YAML; the revert hook covers all three finalisation sites, cancelling the tail cascades up to its compile, and a new rename_fs_lock orders a retry's supersede and write against the superseded chain's revert. The old address resolves at enqueue (StorageJSON address, then the scanner's hostname or IP, then the mDNS default) and persists on the tail's port so it survives restarts. A persisted RENAME from before this change still runs the fused CLI on the compile lane, mirroring how fused INSTALL was retained.

Behavior notes: the dashboard rewriter refuses names it cannot retarget in place (from packages, an !include, or an embedded substitution) with an actionable error on every path; the fused CLI either failed mid job on those or silently broke the rename (an !include supplied name renamed the file without changing the hostname), and it lossily flattened embedded substitutions to literals. The devices/rename response now also carries tail_job so the frontend can follow the chain.

Depends on the helpers extracted in #1817 (merged).

- fixes https://github.com/esphome/device-builder/issues/1812

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1064

## Checklist


esphbot (changes requested):
Blocking issues found.

bdraco:
Addressed the three findings in c81161e: the supersede now runs after the rename lock check (a rejected rename no longer cancels the device's in-flight build, pinned by a test), the successful swap wipes the old device's build dir and idedata cache via the shared wipe_device_build_dir (moved to helpers.build_artifacts), and the revert cleans the new name's StorageJSON and build tree as well. Same commit also folds the chain commit sequence into a public factories.commit_chain shared with the install chain and hoists the rewrite so offline and online renames go through one rewrite and refusal point.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

bdraco:
Both addressed in 60950c6: finalize_rename_swap now catches Exception (with a noqa note naming the wrong-direction-revert hazard) and the failure test is parametrized over OSError and RuntimeError; the is_rename_tail docstring is trimmed to the contract (tail vs fused CLI job) without the decomposition history. The JobType.RENAME enum comment keeps its legacy-retention note deliberately — it mirrors the INSTALL member's comment directly above it, which is the established shape for enum members retained for persisted-job compatibility.

esphbot (approved):
No blocking issues found.

PR #1840, opened 2026-07-05
Make the esphome interpreter selectable per firmware job

bdraco:

Groundwork for a recurring support problem; an operator's remote build server is running an older esphome than the dashboard, so every remote compile silently produces firmware built with that old version. The user reinstalls again and again, keeps getting the same old version, and files a bug because nothing they do seems to change the result. They never realize the receiver is the one that is out of date. The planned fix has the receiver provision a venv with the dashboard's esphome version and build with that, so a stale receiver stops handing back old firmware.

This PR is only the first step and it is behaviour free on its own. It makes the esphome interpreter selectable per job instead of a single value shared by the whole controller. It adds an overridable `_resolve_esphome_cmd(job)` that returns the installed invocation today, and threads its result into `_build_command` through a new optional `esphome_cmd` argument. With no override the argv is identical to before, so there is no functional change yet; the follow up PRs add the version aware provisioning that hangs off this seam.

- groundwork for remote build version auto provisioning; no issue closed by this PR

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/firmware/controller.py
+++ b/esphome_device_builder/controllers/firmware/controller.py
@@ -633,7 +633,7 @@ class FirmwareController:  # noqa: PLR0904 (grandfathered; new public methods ne
         )
 
     async def _resolve_esphome_cmd(self, job: FirmwareJob) -> list[str]:
-        """Return the esphome CLI invocation to compile *job* with."""
+        """Return the esphome CLI invocation to run *job* with."""
         return self.state.esphome_cmd
 
     def _build_cache_args(self, job: FirmwareJob) -> list[str]:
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
On the remote_runner.py local-upload path not going through `_resolve_esphome_cmd`: that is intentional. That subprocess is the offloader flashing a device locally with a binary the receiver already compiled at the correct version, so the upload should use the local installed esphome, not a provisioned interpreter; version matching only applies to the compile the seam feeds. Leaving it untouched here keeps this PR a behavior-free seam; the provisioning that consumes the seam lands in a follow up.

esphbot (approved):
No blocking issues found.

PR #1844, opened 2026-07-05
Advertise and persist receiver auto-provision capability

bdraco:

Groundwork for the same recurring problem as the interpreter seam (#1840); a remote build server on an older esphome silently hands back firmware built with that old version, and the user reinstalls over and over never realizing the receiver is the stale one. The planned fix has the receiver provision a venv with the offloader's esphome version; before the offloader can route a version-mismatched compile to a receiver, it has to know that receiver can actually provision.

This adds that capability negotiation. The receiver advertises an `auto_provision_supported` flag on every peer-link session-open, right alongside the `esphome_version` it already sends; the offloader lifts it off the post-handshake response and persists it on `StoredPairing`, next to the stored version. The flag is sourced from a single seam function that returns `False` today, so nothing advertises the capability yet and no dispatch behaviour changes; the follow up that adds the provisioner flips the seam to the real check and reads the persisted flag when deciding whether to route a mismatched compile. Older receivers that never send the field, and older sidecars without it, read back as `False`, so a mixed version fleet stays safe.

It also validates the receiver's advertised version as a PEP 440 version string at the same wire seam; a malformed or injected value now falls back to unknown rather than reaching storage, and later a pip install argument once the provisioner lands. The check is a small regex so it stays off the cold import path, no `packaging` import.

- groundwork for remote build version auto provisioning; follows #1840; no issue closed by this PR

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/remote_build/bus_handlers.py
+++ b/esphome_device_builder/controllers/remote_build/bus_handlers.py
@@ -104,13 +104,11 @@ def on_offloader_peer_link_opened(
     controller: OffloaderController, event: Event[OffloaderPeerLinkOpenedData]
 ) -> None:
     """
-    Add ``pin_sha256`` to ``_open_peer_links`` and refresh the receiver version.
+    Add ``pin_sha256`` to ``_open_peer_links``; refresh version + capability.
 
-    Receiver's ``esphome_version`` rides on every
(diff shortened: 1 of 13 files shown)


esphbot (approved):
No blocking issues found.

PR #1861, opened 2026-07-05
Add --remote-build-only headless build-server mode with one-shot pairing

jesserockz:

Adds `--remote-build-only`: run the backend as a headless remote-build server (no HTTP dashboard) that an operator can pair from their main builder.

One line on the build server (a pairing-source choice is required — see below):

```sh
esphome-device-builder --remote-build-only --allow-any-pairing-source /var/lib/esphome-builder
```

On first run (zero approved peers) the process opens the standard 5-minute pairing window with a one-shot auto-approve armed, and prints the fingerprint to the console — the pin rendered as the Matrix-SAS 7-emoji sequence (exact port of the frontend's `src/util/pin-emoji.ts`, so the banner matches the pair dialog glyph-for-glyph), the emoji names, and the formatted hex:

```
======================================================================
 REMOTE BUILD PAIRING — window open for 5 minutes
 On your main Device Builder open Settings -> Send builds and
 pair this server. Verify the fingerprint shown there matches:

   🏆 🔥 ⏰ 🌙 🦁 🔒 🔑
   (Trophy, Fire, Clock, Moon, Lion, Lock, Key)
   dd 69 d4 0a eb f9 6f cc ad ea 3c dd aa 7a a8 24 ...
======================================================================
```

The operator pairs from the main builder's UI and eyeballs the fingerprint; the first `pair_request` inside the window is approved without the inbox dance (the row is flushed to `.receiver_peers.json` before the wire response — it is the mode's single trust anchor, so the write is not debounced) and the window closes behind it. Exactly one pairing is allowed: the auto-approve disarms on use and only fires with zero APPROVED rows, and with no UI to open another window, later pair requests get `no_pairing_window`. If nothing pairs before the window lapses the process exits 1. Subsequent runs with an approved peer skip the bootstrap and just serve.

## Pairing-source safety

Trust-on-first-use is never a silent default. `--remote-build-only` requires an explicit pairing-source choice:

- `--allow-pairing-source <IP>` (comma-separated, IPv4/IPv6) — auto-approve **only** a `pair_request` from a known address; any other LAN peer is refused as a closed window (leaking nothing) *without* disarming, so the intended builder can still pair. Closes the "anyone on the LAN who times the window wins the pairing" vector.
- `--allow-any-pairing-source` — the explicit opt-in to accept the first pairing from any source (the banner says so in as many words).

The parser rejects passing both, passing neither, and passing either without `--remote-build-only`. The accepted residual risk of the any-source mode is written up in `docs/THREAT_MODEL.md`. A stronger PSK-based pairing (UI-generated one-time key) is planned as a follow-up.

## Details

- No HTTP site is bound in this mode — `DeviceBuilder.run` skips `web.run_app` and drives the lifecycle through the new `_remote_build_only` runner; the peer-link Noise listener + mDNS advertise are the only network surface.
- The listener is force-enabled over a persisted `RemoteBuildSettings.enabled=false` (there is no UI to flip the toggle back on). Not combinable with `--ha-addon` (parser rejects the pair).
- The config-dir positional is **required** in this mode (no `./configs` fallback): it holds the server's identity, pairing, and build state, and a cwd-relative default would mint a fresh identity (new fingerprint, re-pair needed) whenever the service started from a different directory.
- `helpers/pin_emoji.py` is an exact port of the frontend's Matrix-SAS implementation, validated with a 50-pin node cross-check against the real TS. A drift test reads the SAS table out of the installed `esphome-device-builder-frontend` wheel (the exact bundle the browser runs — no network) and compares it against the backend port; if the frontend changes its rendering, the backend suite goes red so the port is re-synced before the next release (skips only when the frontend package isn't installed).

Verified end-to-end against a live instance: real Noise XX preview + pair from a second process (approved in one round-trip, emoji matching the frontend's rendering for the same pin), a non-allowlisted source refused as a closed window, second pair refused, SIGTERM clean exit, restart takes the already-paired path.

## Types of changes

## Frontend coordination

## Checklist


esphome_device_builder/controllers/remote_build/_receiver_state.py
```diff
@@ -29,6 +29,12 @@ class ReceiverState:
     pairing_window_clients: dict[Hashable, float] = field(default_factory=dict)
     pairing_window_handle: asyncio.TimerHandle | None = None
 
+    # Armed by ``--remote-build-only`` first-pair bootstrap: while
+    # True (and zero peers are APPROVED), the next ``pair_request``
+    # inside the open window is approved without the inbox dance.
+    # One-shot — ``record_pair_request`` disarms it on use.
+    auto_approve_first_pair: bool = False
```
bdraco, line 44: This opens the builder to impersonation attacks. Needs to be a documented / accepted risk

esphome_device_builder/helpers/pin_emoji.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def pin_emoji(pin_sha256: str) -> str:
+    """Render the 7-emoji SAS sequence, space-separated for console display."""
+    return " ".join(emoji for emoji, _ in pin_emoji_slots(pin_sha256))
+
+
+def pin_emoji_names(pin_sha256: str) -> str:
+    """Render the matching SAS emoji names, for terminals with poor emoji fonts."""
+    return ", ".join(name for _, name in pin_emoji_slots(pin_sha256))
```
bdraco: nit: could cache and have a helper to eliminate the duplication

esphome_device_builder/controllers/remote_build/_receiver_state.py
jesserockz: Would there be a way to pre-determine where the request would come from and add something to the cli to only allow a connection from that builder/ip/hostname?

bdraco: You could start with --allow-pairing-source <IP> .. but you'd have to know it in advance... 

That said the risk is minimal and bounded but since its easy to know when one comes online its a potential attack vector for anyone with LAN access

bdraco: Maybe generate a key and allow pairing by PSK

esphbot: Documented in ARCHITECTURE.md "Headless build server" — Bluetooth-style trust. Risk bounded: single-use window, OOB fingerprint compare. Gap: window trusts first requester, no source check. Suggest gating auto-approve on `--allow-pairing-source <IP>` or PSK before merge. Then risk is mitigated, not only disclosed.

esphbot: Yes. Add `--allow-pairing-source <IP/hostname>`. Gate `_auto_approve_pair_request` on `peer_ip` matching. Reject mismatched sources. Turns "first requester wins" into "only declared builder wins". Operator already knows the main builder's address.

esphbot: PSK is the strongest option. Print secret in the banner. Offloader supplies it in `pair_request`. Receiver rejects mismatch inside `_auto_approve_pair_request`. Racing the window alone becomes insufficient. Compose with source-IP gate for defense in depth.

esphbot (changes requested):
Blocking issues found.

esphome_device_builder/controllers/remote_build/_receiver_state.py
jesserockz: I have added some cli flags to allow a set of source ips, or enforce a flag `--allow-any-pairing-source` to make sure the user chooses that path themselves.

jesserockz: I think PSK is a good idea, but that can be done on top of this work in another PR and requires frontend changes.

esphome_device_builder/helpers/pin_emoji.py
esphbot: Slots computed once per pairing banner. No hot path, no per-request cost. `pin_emoji_slots` already the shared indexing core; `pin_emoji`/`pin_emoji_names` differ only in which tuple field they join. Caching adds process-lifetime state for one call. Leaving as-is.

bdraco (approved):

esphbot (approved):
No blocking issues found.

PR #1862, opened 2026-07-05
Add --socket option to listen on a UNIX socket

vexofp:

Add a command line option (`--socket`) that makes the primary web server listen on a specified UNIX socket path (instead of TCP). This restores a feature that was previously available in the legacy dashboard.

- fixes https://github.com/orgs/esphome/discussions/3754

## Types of changes

## Frontend coordination

## Checklist


vexofp:
The option name (`--socket`) and behavior (overriding `--host`/`--port`) were chosen to mirror that of the legacy dashboard. The host/port options _do_ still affect the mDNS advertisement, which seems useful if you point them at where a reverse proxy ultimately listens.

The current changes do not affect the HA ingress or remote build sites, since UNIX socket listening is mainly used with reverse proxies and seems less useful/applicable for those. If requested in the future, `--ingress-socket`/`--remote-build-socket` options could extend the pattern to those sites.

I did not find any existing tests that seemed related to the socket binding options, so I held off adding anything for now. Happy to add some if I can get a bit of direction as to the preferred location/approach.

bdraco:
Definitely needs some tests for this

bdraco:
@esphbot review

esphbot (changes requested):
Blocking issues found.

- No test coverage for the --socket path (maintainer-requested)
- New --socket flag is undocumented, but the docs checkbox is ticked

vexofp:
Added additional documentation and UNIX socket case coverage to the relevant `test_*_binds_public_*` tests.

Did not try to tackle messing with the socket file permissions. (Which _is_ a divergence from the legacy dashboard.) For one, it is poorly-supported / rather cumbersome under `aiohttp` (see https://github.com/aio-libs/aiohttp/issues/4155). Also, the legacy behavior of making the socket world-writable was problematic, since that undermines one of the key security advantages of UNIX sockets and any work-around that later restricts the socket file suffers from a vulnerable race window. There are other ways of safely _adding_ access for socket files (e.g. default ACLs). If there are requests for something like a `--socket-mode` option in the future, that seems like it could fit its own PR.

bdraco:
If you want to do an `aiohttp` PR, I can take a look

esphbot (changes requested):
Blocking issues found.

- Operator precedence makes the TCP-path host assertion vacuous

esphbot (approved):
No blocking issues found.

bdraco (commented):
Please update THREAT_MODEL.md to explain the default permissions are not a bug and not something we defend

bdraco:
@esphbot review

bdraco:
@esphbot review

bdraco:
@esphbot review

bdraco (approved):

vexofp:
> If you want to do an `aiohttp` PR, I can take a look

I'm not sure I have an implementation I'd be comfortable putting in to a general-purpose library...

Fundamentally, the kernel `bind()` interface lacks a mechanism by which to specify initial permissions. (See "Pathname socket ownership and permissions" in [`man UNIX(7)`](https://man7.org/linux/man-pages/man7/unix.7.html)). It follows the process `umask()`, equivalent to as if `open()` were called with all `mode` bits set.

Changing the `umask()` while binding the socket works, but cannot be made thread-safe without cooperation from the whole program. `chmod()`ing the file after creation is safe only if the requested permissions are a superset of those initially set by the mask. Even just portably _checking_ if the requested permissions could be safely added after-the-fact requires a subprocess (because there is no thread-safe/atomic API to query the `umask()`)...

Maybe you could spawn a clean Python subprocess (since `fork()` is disallowed in multithreaded processes on macOS) that inherits the socket, sets the `umask()`, and then `bind()`s there? Feels fragile and much heavier than callers would likely expect...

esphbot (approved):
No blocking issues found.

bdraco:
> > If you want to do an `aiohttp` PR, I can take a look
> 
> I'm not sure I have an implementation I'd be comfortable putting in to a general-purpose library...
> 
If you ever do come up with a solution for `aiohttp`, feel free to ping me for a review over there

PR #1887, opened 2026-07-07
Target id-less multi-entity sub-sensors via synthetic ids

### LINKED ISSUE #1886: [Bug] Invalid YAML in the automation editor when sensors are disabled
### What happened?

1. Create a new project, "ESPHome Starter Kit By Apollo"
2. Add component "AHT20 Temperature & Humidity Module"
3. Add I²C Bus as suggested
4. Do **NOT** enable Temperature and Humidity Sensor
5. Click Add

The structured editor is a bit confused now: It offers to add an automation to the "AHT20 Temperature & Humidity Module"

Image

You can even add a trigger:

Image

Which creates an invalid YAML:

Image

Even after manually enabling the temp and hum sensors, the automation editor is still confused because no ids are set:

Image

Image

Image

After adding an id to temp and hum sensory manually, the automation editor works correctly.

### Dashboard version

1.2.0

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Adding the AHT20 module without enabling its Temperature/Humidity sub-sensors left the automation editor offering entity triggers on the platform item itself; picking one spliced `on_value:` directly under `- platform: aht10`, which is schema-invalid. With the sub-sensors enabled but id-less, they could not be targeted at all until the user hand-added `id:` lines.

Both came from the same gap: `iter_subentities` only yielded sub-blocks with a declared `id:`, so `_scope_component_instances` saw `subs == []` and surfaced the multi-entity platform item as a plain leaf target (the compromise pinned in #1302 as "a separate pre-existing gap").

This extends the existing synthetic-id convention to sub-entities:

- An id-less sub-block now keys on `<parent_instance_id>_<sub_key>` (`aht20_temperature`, or `sensor_0_temperature` under an id-less parent), composing with the `<domain>_<idx>` scheme from #1302. The writer already locates sub-entities by parent + sub-key (`upsert_subentity_handler`), so upsert, parse, and delete round-trip with no writer changes; existing hand-written handlers under id-less sub-blocks become visible and editable.
- Container-ness now comes from the catalog definition (`platform_subentity_keys`), not from which sub-blocks happen to be configured, so a multi-entity platform is never surfaced as a leaf and entity triggers can no longer land on the platform item. With no sub-blocks configured the instance is an empty flagged container; the frontend skips it and explains, instead of producing invalid YAML.

Verified against a live dashboard on the reporter's config shape: get_available surfaces the flagged container and synthetic subs, upsert lands the handler inside the `temperature:` block, the parse round-trips, and the written YAML passes `esphome config`.

Not covered here (follow-up): adding a second `on_value_range` to a sub-entity overwrites the first, because the sub-entity write path predates list-shaped trigger entries; fix in a separate PR.

- fixes https://github.com/esphome/device-builder/issues/1886

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1129

The wire shape is unchanged; the companion teaches the frontend's fallback parser the same synthetic sub-entity ids and adds an explanatory empty state when a container prefill has no configured sub-entities. The frontend degrades gracefully without it (empty picker, submit already blocked).

## Checklist


bdraco:
@esphbot review

bdraco:
Review-pass follow-up: the controller now reuses the parser's `instance_id` composition (promoted to a public `parsing` export) instead of hand-rolling the declared-or-positional string, since the synthetic sub ids make that agreement load-bearing; the leftover `list(...)` materialization is gone with it.

@esphbot review

esphbot (changes requested):
Blocking issues found.

- get_available docs still say only "ided" sub-entities surface

bdraco:
Fixed the stale `get_available` row in docs/API.md: it now describes configured sub-entities surfacing on their declared id or the synthetic `<parent_instance_id>_<sub_key>` fallback, and the container being flagged `is_entity_container` even with no sub-blocks configured. Also ticked the docs box in the description.

@esphbot review

esphbot (changes requested):
Blocking issues found.

- get_available docs still describe only "ided" sub-entities and omit the empty-container case

esphbot (approved):
No blocking issues found.

PR #1890, opened 2026-07-07
Log the remote-build pairing banner from a local key

bdraco:

Fixes a flaky failure in `test_bootstrap_first_pair_success` (seen on PR #1888's CI, unrelated to that change); the headless `--remote-build-only` pairing banner sometimes printed no key.

`_bootstrap_first_pair` logged the banner reading `receiver.state.bootstrap_pairing_key`, but that read happens after an `await` (the identity load). When a pair request lands during that await it runs the auto-approve path, which clears the field to `None`, so the banner logs `None` and the test's `assert key in banner` fails. In production the operator has not seen the banner yet so cannot pair that early, but the state field is still the wrong thing to read. Hold the generated key in a local and log from that, so the banner always shows the key regardless of concurrent clearing.

Verified by widening the identity-load await to force the race; the banner logged `None` before the change and the real key after.

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1930, opened 2026-07-09
Platform-scope ethernet type options and esp32-only SPI fields

bdraco:

Adding the Ethernet component on an RP2040 board offers all 17 chipset types, though only ENC28J60, W5100, W5500, W6100 and W6300 validate there; the rest are ESP32 PHYs. Longstanding bug; the schema bundle flattens away ethernet's entire platform split (the per type `cv.only_on` wrappers, `RP2_ETHERNET_TYPES`, and the `cv.only_on_esp32` gates on the SPI sub fields), so the catalog ships a flat union and no `supported_platforms` at all.

This recovers the split from the live ethernet module at sync time, the same way logger's `hardware_uart` gets its per platform UART choices. The `type` entry gains `platform_options` derived from `ETHERNET_TYPES`, `RP2_ETHERNET_TYPES` and `SPI_ETHERNET_TYPES` (the controller already resolves those to plain `options` per target, so no frontend change), and the component's own `supported_platforms` becomes the option map keys, so ethernet stops being offered on esp8266 and libretiny boards where no type validates. The esp32 only SPI sub fields (`clock_speed`, `interface`, `polling_interval`) already carry `supported_platforms` from the existing `_apply_platform_constraints` introspection, so nothing to do there. On esphome builds predating `RP2_ETHERNET_TYPES` the split is skipped and the flat union ships as before.

Out of scope, noted for later: esp32 variant fidelity (RMII types only validate on classic ESP32 and P4, GENERIC/YT8531 only on S31); those rules live in upstream validator code rather than exported constants, and `platform_options` supports variant keys when we want to layer them on.

- found testing RP2040 boards after #1926

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

Issue #1928, opened 2026-07-09
[Bug] Firmware tasks - compile twice for out of network device


Masterz69:
### What happened?

May be ut is intended behaviour, but for me looks like a bug.

Started update for all my devices. 
For devices currently in offline: firmware compiled, upload failed, both tasks appeared in Queue again.
After cycle is done those tasks started again: firmware compiled from scratch, not using previous compilation.
Then upload failed second time and now marked as failed.

Probably second attemp to upload is good idea, but why need to fully recompile firmware ?

### Dashboard version

Device Builder 1.1.0 ESPHome 2026.6.5

### How did you install it?

Home Assistant App (preview toggle)

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

Masterz69:
Even more I notice when tried to "Install"  already compiled firmware (just several minutes ago) - compilation starting again from scratch...
From all components etc.

```
Tool Manager: Installing platformio/tool-esptoolpy @ ~1.30000.0
Unpacking  [####################################]  100%
Tool Manager: tool-esptoolpy@1.30000.201119 has been installed!
HARDWARE: ESP8266 80MHz, 80KB RAM, 1MB Flash
ESPHome: Excluding Updater.cpp from build (using native OTA backend)
ESPHome: Excluding core_esp8266_waveform_phase.cpp from build (waveform not required)
ESPHome: Excluding core_esp8266_waveform_pwm.cpp from build (waveform not required)
Dependency Graph
|-- ESP8266WiFi @ 1.0
|-- ESP8266mDNS @ 1.2
|-- noise-c @ 0.1.11
Compiling .pioenvs/garland-hall/src/esphome/components/api/api_buffer.cpp.o
Compiling .pioenvs/garland-hall/src/esphome/components/api/api_connection.cpp.o
Compiling .pioenvs/garland-hall/src/esphome/components/api/api_frame_helper.cpp.o
```
^^ firmware just compiled recently, installed and no changes made in config or ESPHome version.

bdraco:
Its missing in the bulk api.  The recompile should be a quick noop before upload to make sure nothing has changed.  I'll post more details once I decide on a fix

bdraco:
[...]

PR #1932, opened 2026-07-09
Share one per-item core between single and bulk command handlers

bdraco:

Follow up to #1929, which fixed an offline deferral that existed in the single install but not in install_bulk. This makes the underlying convention structural for the two remaining drift prone pairs: every single/bulk command pair now shares exactly one per item core function, so a policy added to one entry point cannot silently miss the other.

`firmware/compile` and `compile_bulk` now both call a new `factories.enqueue_compile` instead of duplicating the resolve, create, enqueue body by hand. `devices/archive` and `archive_bulk` now both call a new `archive.archive_single_checked` (validate plus archive) instead of the bulk closure re declaring the wrapper steps inline. The bulk handlers keep only their bulk scoped concerns: batched boundary validation, stale first ordering, and the per item error policy. CLAUDE.md documents the convention with the canonical examples.

No behavior change; the full test suite passes with no test edits. Follow ups will unify the delete/archive not found error surface (a real drift bug the audit found) and add single/bulk parity guard tests.

- related to https://github.com/esphome/device-builder/issues/1928

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -291,6 +291,18 @@ against legacy behaviour before assuming the simpler version suffices.
   `devices/list_archived` (cold archive listing, read-once). `labels/list`
   is the snapshot-fetch-then-events holdover — new code lands through
   `initial_state`, not by copying it.
+- **Single/bulk command pairs share one per-item core function.** Every
+  `X` / `X_bulk` verb keeps its per-item behavior in a single callable
+  both handlers invoke (`factories.enqueue_install_or_defer` /
+  `enqueue_compile` for firmware, `archive.archive_single_checked` /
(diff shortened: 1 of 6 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
Applied the cleanup pass in 6dcf24ab; the archive validation now lives inside the three archive.py cores instead of a new checked wrapper (no new names, both handlers keep calling _archive_single), the two firmware bulk loops share one _run_bulk error policy helper, and the CLAUDE.md bullet no longer references the parity test file that lands in a follow up.

@esphbot review

esphbot (approved):
No blocking issues found.

PR #1936, opened 2026-07-09
Add single/bulk parity guards for every command pair

bdraco:

Third and final part of the anti drift work from the #1928 audit. Two guard layers keep single and bulk WS commands from diverging again.

An inventory guard walks every `@api_command` under `controllers/` and asserts each `*_bulk` command has a registered single twin. Per verb parity scenarios then run the same situation through the single command and a one element bulk call and assert corresponding outcomes: install chains match for online devices, offline devices defer to the same compile only job from both entry points, compile threads force_local identically, delete and archive of a missing YAML produce the same not_found message as the bulk error row, set_labels of an unknown device likewise, and the two intentional differences are pinned explicitly (firmware batch validation is atomic while devices bulk isolates bad rows).

The firmware conftest `StubDevices` gains an optional device catalog with a configuration keyed O(1) index, replacing the private stub in test_bulk_order. CLAUDE.md's single/bulk convention bullet now points at the parity files.

Stacked on #1933.

- related to https://github.com/esphome/device-builder/issues/1928

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -303,7 +303,10 @@ against legacy behaviour before assuming the simpler version suffices.
   and re-raises fleet-wide `NO_COMPATIBLE_PEER`; devices'
   `run_bulk_per_device` collects `{configuration, success, error}`
   rows). Never inline the per-item body in a bulk loop — that's how
-  `install_bulk` missed the offline deferral (#1928).
+  `install_bulk` missed the offline deferral (#1928). New pairs ship
+  with parity coverage: `tests/test_single_bulk_parity.py` (every
+  `*_bulk` command has a single twin) plus per-verb scenarios in
+  `tests/controllers/{firmware,devices}/test_single_bulk_parity.py`.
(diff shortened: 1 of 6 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
Applied the cleanup pass in 0e7753a9; the inventory guard now reuses collect_api_commands (inheriting production's underscore skip) and drops the dead module level scan, a second inventory guard pins the keyword surface of every bulk twin against an intentional gap allowlist, the force_local parity test now wires a paired receiver so it actually discriminates (mutation verified), the redundant delete happy path test is gone, and the devices tests share one controller where the single call cannot mutate state.

@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1940, opened 2026-07-10
Carry active_source through device rebuilds

### LINKED ISSUE #1939: Update Available Marker Does not Return if  queued upgrade is canceled
### The problem

If multiple devices are chosen to have their version update and then some are cancelled from the Firmware Tasks window before the upgrade runs, the blue dot indicator does not show for those devices that were cancelled.

### Which version of ESPHome has the issue?

2026.6.5

### What type of installation are you using?

Home Assistant Add-on

### What platform are you using?

ESP32

### Component causing the issue

_No response_

### YAML Config

```yaml
Not relevant
```

### Anything in the logs that might be useful for us?

```txt
No
```

### Additional information

_No response_

bdraco:

Fixes the update available dot vanishing for devices whose queued bulk update was cancelled. `Device.active_source` is runtime only and was missing from `load_device_from_storage`'s carry forward, so any rebuild reset it to UNKNOWN; the post compile `scanner.reload` in a bulk update did exactly that for every device, and the frontend gates the update and modified dots on `active_source == "mdns"`. Flashed devices reboot and get re-claimed on their announce, cancelled ones sit dotless until the next mDNS TTL refresh, which is what the reporter saw. The monitor only re-fires `on_source_change` on a real transition, so nothing repaired the field in the meantime.

The fix carries `active_source` from the previous in-memory device, the same shape as `state`, `ip_addresses` and the `api_encryption_active` carry that fixed this class of wipe before. Reproduced end to end with fake mDNS announces plus a headless browser: before the fix the dot disappeared for the two devices whose compiles completed, after it all three keep the dot and `active_source` stays mdns through enqueue, cancel and upload.

- fixes https://github.com/esphome/device-builder/issues/1939

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/helpers/device_yaml/_loading.py
+++ b/esphome_device_builder/helpers/device_yaml/_loading.py
@@ -11,7 +11,7 @@ from esphome.const import CONF_PACKAGES
 from esphome.core import EsphomeError
 from esphome.storage_json import StorageJSON
 
-from ...models import Device, DeviceState
+from ...models import Device, DeviceState, ReachabilitySource
 from ...models.boards import normalize_platform
 from ..mac_addresses import derive_interface_macs
 from ..storage_path import resolve_compiled_config_path, resolve_storage_path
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1950, opened 2026-07-10
Variant-gate esp32 sram1_as_iram to the classic ESP32

bdraco:

Gates the esp32 `framework.advanced.sram1_as_iram` field to the classic ESP32 variant in the Visual Editor. #1948 surfaced the field for every esp32 variant, but esphome rejects it on any other variant at validate time:

```python
if config[CONF_VARIANT] != VARIANT_ESP32 and advanced[CONF_SRAM1_AS_IRAM]:
    raise cv.Invalid(f"'{CONF_SRAM1_AS_IRAM}' is only supported on {VARIANT_ESP32}", ...)
```

So on `variant: ESP32C2` / `ESP32S3` / … the toggle was an offered-but-invalid option. This stamps `depends_on: variant` + `depends_on_value_any` on the field so the frontend only shows it on the classic ESP32 (both cases — the board-written `esp32` and a user-typed `ESP32` — are covered). The variant list isn't hand-maintained: `_esp32_variant_gate` derives it from esphome by enabling the field per variant and running esphome's `final_validate`, keeping the variants it doesn't reject (`sram1_as_iram` → `(esp32, ESP32)`; a field valid on every variant derives to `None`, no gate). So an upstream change to which chips accept the field flows through on the next sync without a code edit. `esp32.json` regenerates byte-identically to the hand-listed version.

`sram1_as_iram` is nested under `framework.advanced` while `variant` is a top-level esp32 field, so this needs the frontend to resolve a nested `depends_on` against the component root — that's the companion PR below. Verified together in the editor: `variant: esp32` shows the Advanced group + toggle; `variant: ESP32C2` hides both.

Follow-up to #1948; surfaced while testing it.

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1170

## Checklist


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
Addressed the empty-tuple suggestion by removing the hand-maintained variant list entirely (9641b2bc): `_esp32_variant_gate` now derives the restriction from esphome — it enables the field per variant and runs esphome's `final_validate`, keeping the variants that aren't rejected (`sram1_as_iram` → `(esp32, ESP32)`, matching the old hardcode; regenerates `esp32.json` byte-identically). The "valid on all variants" case is now `None` (no gate) derived naturally rather than an unexercised sentinel, and it's pinned by a test (`disable_fatfs` → None). This also answers the 'can we derive it without hardcoding' question.

@esphbot review

bdraco:
Pushed two follow-ups: dropped the `except ImportError: return None` from `_esp32_variant_gate` and its sibling `_esp32_default_framework` — the sync requires esphome (it asserts the version at startup), so the import should fail loud rather than silently ungate the field; `None` now means only "valid on every variant."

@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
Addressed all three in a49e3231, in the fail-loud spirit — the guard used to collapse several outcomes into a permissive no-gate; now they're distinguished:

- **Suggestion #1 / silent-failure #1** (rejected-everywhere and no-testable-variants → silent no-gate): both now `raise RuntimeError`. `tested == 0` means no variant validated a base config (upstream schema shift); `not accepted` means a curated field is valid on no variant. Either is a sync-config error, so it crashes the sync rather than surfacing an ungated always-invalid field. `len(accepted) == tested` stays the one legitimate no-gate case.
- **Silent-failure #2** (fragile error-path inspection): the reject check now matches the field key in the error's `path` **or** its message (`field_key in (e.path or []) or field_key in str(e)`), so it survives esphome moving where it reports the rejection.

Added a test that a bogus key raises (`test_variant_gate_fails_loud_when_underivable`). Catalog output unchanged.

@esphbot review

esphbot (approved):
No blocking issues found.

PR #1957, opened 2026-07-11
Surface web_server auth on the main form

bdraco:

The Web Server panel shows the security banner "The web server is enabled without authentication" while the `auth:` group it points at only renders after flipping "Show advanced settings (16)". Every web_server field classifies advanced (upstream marks none of them with `cv.Visibility`, and `auth` is `cv.Optional`, so the optional-means-advanced default applies), leaving the main form empty except the banner.

`auth` joins `_IMPORTANT_KEY_ORDER` in the credentials group, so `_classify_advanced` keeps the group on the main form and the visibility cascade no longer forces its required `username`/`password` children under the disclosure. This matches the other two security-banner targets, which already render on the main form: `api.encryption` (via `_FIELD_OVERRIDES`) and `ota.esphome.password` (`password` is an important key). `auth` is the only key with that name in the whole catalog, so the regenerated diff is limited to `web_server.json`.

No frontend change: rendering honors the per-entry `advanced` flag, the advanced count is derived (drops to 15), and the banner's "Enable authentication" flow is untouched.

- reported directly with a screenshot of the web_server panel; no issue filed

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1958, opened 2026-07-11
Gate web_server sorting_groups on version 3

bdraco:

The Web Server panel renders the Sorting Groups list on every version even though esphome accepts `sorting_groups` only with `version: 3` (the description even says so). With `version: 2` set, or with no version at all (the default is 2), the field still shows and invites configuration that fails validation with "'sorting_groups' is only supported in 'web_server' version 3".

The version restriction lives in upstream's custom `validate_sorting_groups` validator, which the schema bundle can't express, so this uses the documented `_FIELD_OVERRIDES` hook to add the structured gate: `depends_on: version` with `depends_on_value_any: [3, "3"]` (the YAML scalar arrives as an int or a quoted string, same both-shapes precedent as the esp32 `sram1_as_iram` variant gate). The frontend already evaluates these gates; an unset `version` resolves to no match, so the field also hides in the default-version case.

No frontend change needed.

- reported directly with a screenshot of the web_server panel on version 2; no issue filed

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1966, opened 2026-07-11
Surface esp32 minimum_chip_revision on the framework form

bdraco:

The device log doc links already point users at `minimum_chip_revision` (the "Set minimum_chip_revision" hint), but the field was still `advanced: true, hidden: true` in the generated esp32 catalog, so a user following the hint had no way to set it in the visual editor. `sram1_as_iram` got lifted out of the hidden advanced group for exactly this reason; this does the same for `minimum_chip_revision`.

The curated `_ESP32_ADVANCED_VISIBLE` allowlist becomes a mapping from field key to the value used when `_esp32_variant_gate` probes each variant; the old boolean `True` probe fails `minimum_chip_revision`'s `one_of` schema on every variant, which would trip the fail-loud guard. The gate still derives the classic-ESP32 restriction from esphome's FINAL_VALIDATE, so the field ships with the same `depends_on: variant` gate as `sram1_as_iram`.

Verified against a live dev dashboard: the ESP32 Platform form's Advanced card shows the Minimum Chip Revision select (0.0 through 3.1) beside Sram1 As Iram on a classic ESP32, and both disappear when the variant switches to ESP32C3.

- fixes the editor gap behind the `Set minimum_chip_revision` log doc link

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1983, opened 2026-07-13
Regenerate the imported board catalog with the repaired sync

bdraco:
# What does this implement/fix?

Regenerates the imported board catalog with the repaired sync from #1979 and #1982, and closes the validation gap that let broken recommended setups ship.

The regeneration touches 224 imported manifests plus their generated bodies and indexes. The headline changes:

- The three st7701s boards from the issue (`guition_esp32_s3_4848s040`, `esp32_s3_4_0_inch_smart_display`, `tuya_t3e_esp32_s3`) get their complete display presets (`dimensions`, `data_pins`, `init_sequence` locked) plus the display data-pin GPIOs in the pins block.
- 19 imported boards gain an `onboard_psram` featured entry lifted from their page, stamped as a `requires` prerequisite of PSRAM-allocating leaves.
- Dependency hubs and their bus chains now ride along: `tuya`, `time`, `ads1115`, `modbus_controller`, DAC hubs, `i2s_audio`, plus `modbus` riding `uart`. Multi-bus consumers carry their upstream `<bus>_id` locked. Named pin aliases (`TX1`) and previously dropped nested presets come through, and 6 `board:`-only manifests gain their authoritative esp32 `variant`.
- 8 boards are removed because their upstream pages fail real ESPHome validation beyond repair (templated ids, platform-mismatched components, invalid pin modes): `genvex_nibe_alphainnotec_heat_recovery_ventilation`, `govee_h6086_e27_5w_500lm_rgbcct_bulb`, `iot_devices_ggreg20_v3`, `ledvance_smart_table_frame_multicolor`, `loratap_sc411wsc_curtainswitch`, `makerfabs_ac_dimmer`, `martin_jerry_rp_fc01`, `moes_ir_thermostat_s16pro`. Each was shipping a recommended setup ESPHome rejects; they return automatically once the pages are fixed, and each has an investigation issue: #1984, #1985, #1986, #1987, #1988, #1989, #1990, #1991. Other page fixes flow the opposite way too: `aquaping` and `mirabella_door_window_sensor`, dropped by the stricter guard in #1979, import again now that named pins carry through.

[...]

PR #1992, opened 2026-07-13
Reuse featured buses in lifts and prune dropped entries' pins

bdraco:

Two importer defects found while regenerating the catalog (#1983), split out per review so the code lands on its own.

**Duplicate bus lifts.** The hub lift passes start with an empty bus map, so a hub whose bus was already featured (lifted by an earlier pass or extracted directly) materialized a second copy: on `hankerila_ea8` and the kincony boards the pcf8574/ads1115 hubs ended up with `i2c_id: bus_a` locked while `requires` pointed at a freshly minted `bus_a_2` duplicate of the same physical bus. CI caught exactly one of them (`test_committed_catalog_declares_every_reference` resolves references through a fields-id map, and the duplicate won the dict slot on every board except hankerila). `_seed_bus_local` now maps every already-featured bus into each pass's `_LiftState`, keyed the same way `_ensure_buses` resolves deps (`(domain, upstream id)` plus the sole-bus `(domain, None)` fallback), so lifts reuse instead of duplicating. Verified against the upstream KinCony pages: one physical `bus_a`, and the regenerated manifests now mirror it 1:1.

**Stale pins after gate drops.** When the validation gate drops an entry, its `pins[].occupied_by` labels survived, so GPIOs rendered as occupied by components the board no longer defines (esphbot's suggestion on #1983). `_apply_drops` now prunes pins owned by dropped entries and relabels shared GPIOs a surviving entry still locks. The reconstruction matches how extraction produced the data in the first place: the drop-match set covers every label `_occupancy_label` can emit (display name, upstream id, local id, component id fallback), and survivor pin fields are detected by the catalog's `type: "pin"` (so `sda`/`scl`/`clk` count) plus the pin-group name pattern, with expander channels excluded because extraction never records them as board GPIOs.

The catalog regeneration that exercises both fixes ships in #1983, which is based on this branch.

- follow-up to #1982; the regenerated catalog in #1983 depends on it

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/script/_full_setup_gate.py
+++ b/script/_full_setup_gate.py
@@ -228,7 +228,10 @@ def _entry_for_path(
 
 
 def _apply_drops(record: dict[str, Any], drop_ids: set[str]) -> None:
-    """Remove *drop_ids* from the record's featured entries, bundles, and requires."""
+    """Remove *drop_ids* from the record's featured entries, bundles, requires, and pins."""
+    dropped = [
+        entry for entry in record.get("featured_components") or [] if entry["id"] in drop_ids
+    ]
     record["featured_components"] = [
         entry for entry in record.get("featured_components") or [] if entry["id"] not in drop_ids
     ]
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1995, opened 2026-07-13
Carry AUTO_LOADed dependencies into the catalog so IR hubs lift

### LINKED ISSUE #1991: Investigate moes_ir_thermostat_s16pro import: remote_transmitter hub not lifted for climate_ir
The full-setup validation gate (#1982) refuses this board, so the catalog regeneration in #1983 removes it. Filing to investigate.

Skip reason: `full setup fails validation: Component climate_ir requires component remote_transmitter @ data['climate_ir']`.

The upstream page does define a `remote_transmitter:` block, so unlike the other refused boards this one may be an importer gap: the dependency lift did not materialize the hub (worth checking why; the block may be list shaped, carry an unrepresentable field, or the `climate_ir` dependency may resolve oddly since `climate.*` platforms declare the dep through the `climate_ir` base). If the lift can be fixed, this board comes back on the next sync.

Upstream page: https://devices.esphome.io/devices/Moes-IR-Thermostat-S16Pro/

bdraco:

Fixes #1991. The Moes IR thermostat's page defines a `remote_transmitter:` block, but the import gate refused the board with `Component climate_ir requires component remote_transmitter`. The dependency-driven hub lift never fired because the catalog ships every `climate_ir` platform dependency-less: ESPHome declares the requirement through `AUTO_LOAD` (`climate_ir_lg` auto-loads `climate_ir`, whose `DEPENDENCIES` carries `remote_transmitter`), and the schema bundle only surfaces the direct list.

`sync_components.py` now walks each component's `AUTO_LOAD` closure at catalog build and folds the auto-loaded components' `DEPENDENCIES` into the entry's catalog deps (auto-loaded names themselves are subtracted since they are always present, and the additions land before the transport-implicit filter so nothing implicit leaks). The dependency assembly moves into `_assemble_dependencies` to keep `build_component_entry` readable. On the live corpus this enriches 48 components: 23 IR climate platforms gain `remote_transmitter`, audio and BLE components gain `esp32`, `online_image` gains `display`, and so on — all validation-enforced deps the frontend's add dialog and the board importer previously couldn't see.

The enriched deps exposed one more duplicate-bus shape in the importer: a hub whose chained bus dep (`modbus` riding `uart`) also appears as a direct dep now looked the same bus up under two keys (the explicit `uart_id` and the ref-less sole-bus form) and lifted it twice. `_register_bus_keys` memoizes a lifted bus under its canonical block id, and `_lookup_bus` resolves a ref-less dep through the sole page block's id before materializing, so both paths land on the one physical bus.

With both changes the Moes board imports with its IR transmitter hub (pin 7 locked, carrier 50%) and its full setup passes the validation gate; `earu_ea777` and `kincony_m30` keep exactly one uart. The catalog regeneration (components plus the returning board) ships in a stacked follow-up PR.

- fixes https://github.com/esphome/device-builder/issues/1991

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1997, opened 2026-07-13
Resolve page substitutions at import time so templated values extract

### LINKED ISSUE #1984: Investigate genvex_nibe_alphainnotec import: substitution-built modbus_controller id fails validation
The full-setup validation gate (#1982) refuses this board, so the catalog regeneration in #1983 removes it. Filing to investigate whether the import can be repaired or the upstream page needs a fix.

Skip reason: `full setup still fails validation after repairs`.

Gate log:

```
dropping modbus_ctrl_id: IDs must only consist of upper/lowercase characters, the underscore character and numbers. The character '$' cannot be used @ data['modbus_controller'][0]['id']
dropping binary_sensor_modbus_controller_1..3: Component binary_sensor.modbus_controller requires component modbus_controller
```

The upstream page builds its `modbus_controller` id from a `${...}` substitution, and the import carries neither the substitutions block nor resolved values, so the lifted hub ships a literal `$` id and every modbus consumer cascades. Possible directions: resolve the page's `substitutions:` at import time, or treat a substitution-valued `id` like other unrepresentable values and lift the hub with a generated id instead.

Upstream page: https://devices.esphome.io/devices/Genvex-Nibe-AlphaInnotec-heat-recovery-ventilation/

### LINKED ISSUE #1988: Investigate loratap_sc411wsc import: substitution-valued pins leave nothing extractable
The repaired sync (#1979, #1982) now skips this previously imported board entirely (`no extractable featured components`), so the catalog regeneration in #1983 removes it. Filing to investigate.

Every featured candidate's required `pin` is a `${...}` substitution (`pin: ${open_switch}`), which the required-field guard from #1979 correctly refuses to ship half-imported. The page defines the values in its `substitutions:` block, so resolving page substitutions at import time would recover this board fully. Same underlying theme as the Genvex board.

Upstream page: https://devices.esphome.io/devices/LoraTap-SC411WSC-CurtainSwitch/

bdraco:

Fixes #1988 and #1984. The LoraTap SC411WSC page defines every featured candidate's pin as a `${open_switch}`-style reference to its own `substitutions:` block, so the required-field guard correctly refused each half-import and the board vanished with `no extractable featured components`. The Genvex page is the same theme with substitution-valued ids (`id: ${modbus_ctrl_id}`).

The importer now runs ESPHome's real substitution pass (`esphome.components.substitutions.do_substitution_pass`) over the parsed config fence before extraction, at the tail of `_first_config_yaml` so every consumer of the tree sees resolved values. Whole-scalar references keep their type, `$var` and `${var}` forms both resolve, inter-variable references chain, and jinja expressions like `${1/86400}` evaluate. The pass is non-strict: an undefined reference stays literal and the existing `$` rejection in `_is_simple_scalar` / `_TEMPLATE_VAR_RE` stripping remain the safety net for it, as they do when the pass fails or esphome is unavailable (the board then imports exactly as before). The resolved tree is rebuilt as plain builtins since `substitute()` emits OrderedDicts that neither the manifest SafeDumper nor the gate's `generate_device_yaml` path can represent.

355 of 772 upstream pages define a substitutions block, so the catalog regeneration (140 changed manifests plus the two returning boards, both passing the validation gate with all their featured entries) ships in a stacked follow-up PR. No new gate refusals; the three remaining refused boards (govee_h6086, iot_devices_ggreg20_v3, martin_jerry_rp_fc01) are the genuinely broken pages already tracked in their own issues.

- fixes https://github.com/esphome/device-builder/issues/1988
- fixes https://github.com/esphome/device-builder/issues/1984

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/script/sync_esphome_devices.py
+++ b/script/sync_esphome_devices.py
@@ -225,10 +225,10 @@ _PLACEHOLDER_PATTERNS: list[re.Pattern[str]] = [
     re.compile(r"\byour[\s_-]+(key|address|token|id)\b", re.IGNORECASE),
 ]
 
-# Template substitutions like ``${friendly_name}`` that upstream pages
-# resolve at runtime via ``substitutions:``. We don't carry the
-# substitutions block forward, so anything still containing one is
-# unsafe to surface as a preset value or as an ``occupied_by`` label.
+# Template substitutions like ``${friendly_name}``. The page's own
+# ``substitutions:`` block is resolved in ``_resolve_page_substitutions``,
+# so anything still containing one is an undefined reference (or the pass
+# was skipped) — unsafe to surface as a preset value or ``occupied_by`` label.
 _TEMPLATE_VAR_RE = re.compile(r"\$\{[^}]*\}")
 
 # Deprecated flat ``clk_mode: GPIO<n>_(IN|OUT)`` encodes the RMII clock
@@ -513,6 +513,8 @@ def _first_config_yaml(body: str, device_dir: Path) -> tuple[dict[str, Any], str
     """
     First parseable ``yaml`` config fence as ``(parsed, raw_text)``, following ``file=``.
 
+    *parsed* has the page's ``substitutions:`` block resolved over the tree.
+
     A device page often splits optional snippets across separate fences. The
     onboard ``ethernet:`` block (real hardware we lift) is sometimes a standalone
     fence after the base config, so fold just that one into the primary fence when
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #2002, opened 2026-07-13
Add Waveshare ESP32-C3-Zero

DeepCoreSystem:

Add Waveshare ESP32-C3-Zero

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

Maintainer edit: rebased onto current main and reduced the diff to just this board — the previous push regenerated the whole board catalog with a local esphome install, reverting a dozen unrelated board bodies (which is what was failing `Lint + smoke checks`). The catalog JSON is now regenerated single-board via `script/update_board.py waveshare_esp32_c3_zero` against the pinned esphome (2026.7.0b2), and the PR description checkboxes are fixed so the label applier can parse them.

bdraco:
@esphbot review

Verified the board data against the canonical sources — the [Waveshare wiki](https://www.waveshare.com/wiki/ESP32-C3-Zero) and the official [arduino-esp32 `waveshare_esp32_c3_zero` variant](https://github.com/espressif/arduino-esp32/blob/master/variants/waveshare_esp32_c3_zero/pins_arduino.h): RGB LED on GPIO10 (WS2812, RGB order), BOOT button on GPIO9, UART0 TX=21/RX=20, USB D-/D+ on 18/19, ADC on GPIO0–5, strapping 2/8/9, JTAG 4–7, GPIO12–17 consumed by flash, 4MB flash, 160MHz, no USB-UART bridge — all match. `esp32-c3-devkitm-1` is the right base board (platformio has no dedicated C3-Zero definition; same fallback as esp32-c3-supermini and generic-esp32c3).

One value adjusted in the last push: `ram_size` 393216 → 327680 to match the platformio `maximum_ram_size` convention every other C3/S3 board in the catalog uses.

esphbot (changes requested):
Blocking issues found.

- Branch is stale — reverts the apollo-esk-1 docs_url fix (#2004) and conflicts on merge

bdraco:
@esphbot review

Rebased onto `760fa880` (#2004's apollo-esk-1 docs_url fix is now in the base, not reverted) and re-ran `script/update_board.py waveshare_esp32_c3_zero`; the drift check against a full `sync_boards.py` is clean.

esphbot (approved):
No blocking issues found.

bdraco:
Both suggestions verified — no change needed:

1. **`rgb_order: RGB` is correct for this board.** The official arduino-esp32 variant for this exact board ([`variants/waveshare_esp32_c3_zero/pins_arduino.h`](https://github.com/espressif/arduino-esp32/blob/master/variants/waveshare_esp32_c3_zero/pins_arduino.h)) explicitly declares `RGB_BUILTIN_LED_COLOR_ORDER LED_COLOR_ORDER_RGB` — that macro exists solely to override the GRB default, so it's a deliberate vendor statement, not an omission. The S3-Zero sibling variant declares the same; Waveshare Zero boards ship RGB-ordered LEDs.

2. **`available: true` + `occupied_by` is the right shape here.** `available: false` means reserved: the picker disables the pin, and `validate_definitions.py` rejects a featured preset on a reserved pin unless the preset is `locked` — the `onboard_led` preset targets GPIO10 unlocked, so flipping it to `false` would fail validation. Both pins are genuinely usable (GPIO10 drives the LED via the featured preset; GPIO9's BOOT button is commonly reused as a user `binary_sensor`), and `occupied_by` already surfaces the 'occupied by …' hint in the picker. The guition boards use `available: false` for pins hard-wired to non-optional peripherals (display bus), which is a different case.

bdraco (approved):

bdraco:
Thanks @DeepCoreSystem

PR #2019, opened 2026-07-13
Adopt the per-line layout for the components and automations indexes

### LINKED ISSUE #2012: Adopt the per-line index layout for components.index.json and automations.index.json
#2011 switched `boards.index.json` and `featured_components.index.json` to a one-entry-per-line layout so concurrent PRs stop conflicting textually and diffs are per-entry reviewable. The two remaining single-line catalog indexes should adopt the same layout:

- `components.index.json` — 936 entries on one 459KB line. Can switch today: its envelope has a single `components` list, so `sync_components.py`'s `swap_split_catalog_in` call just needs `entries_per_line=True` (+ `index_entries_key="components"`) and a regen.
- `automations.index.json` — ~1084 entries across **five** lists (`triggers` / `actions` / `conditions` / `light_effects` / `filters`) on one 354KB line. Needs a multi-key variant of `dumps_envelope_entries_per_line` (accept a set of entry keys instead of one) before it can switch.

Why it matters: these two churn on every ESPHome release via the nightly catalog-sync PRs, and the repo policy is that a human reviews every catalog PR — a one-line 400KB blob defeats that review (bots skip it for size too). Same rationale as #2011, arguably stronger.

Suggested shape: generalize `dumps_envelope_entries_per_line(payload, entries_key)` to accept `entries_keys: Collection[str]`, keep `swap_split_catalog_in`'s opt-in kwarg, wire both `sync_components.py` writers, ship the one-time regenerated indexes, and extend `tests/test_catalog_split_layout.py`. Once all callers are per-line, the `entries_per_line` kwarg can become always-on and disappear.

bdraco:

Adopts the per-line index layout from #2011 for the two remaining single-line catalog indexes, `components.index.json` and `automations.index.json`, so the nightly catalog sync PRs get per-entry reviewable diffs instead of a 400KB one-line blob.

`dumps_envelope_entries_per_line` now takes a collection of entry keys instead of one, which lets the automations envelope expand all five of its lists (`triggers`, `actions`, `conditions`, `light_effects`, `filters`). With that, every `swap_split_catalog_in` caller is per-line, so the `entries_per_line` opt-in and the now dead `sort_keys` kwarg are gone and per-line is the only index path.

The two shipped indexes were reformatted mechanically through the same writer helper, not by a full catalog sync, so the diff is pure layout; the reformat asserted `orjson.loads` equality against the old bytes. Runtime readers parse whole files with orjson and are layout agnostic. New tests pin the multi-key helper behaviour and the committed line shape of both files, mirroring the boards pins.

- fixes #2012

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #2038, opened 2026-07-14
Give the mDNS clear op its own seam instead of the empty-ip sentinel

### LINKED ISSUE #2036: Give the mDNS clear op its own seam instead of the empty-string apply_ip sentinel
`apply_ip(name, "")` multiplexes two distinct operations over one wire: a real single-IP observation, and "the device left mDNS, clear the resolved set". Since #2034 the clear op has different semantics (drop `runtime_state.ip_addresses`, keep the last-known `Device.ip`), and the empty-string sentinel is decoded in three places: `_dispatch_ip`'s dedupe predicate, `on_ip_change`'s skip/mutate predicates, and the `RecordingMonitorCallbacks.on_ip_change` test double, which needed its own `if ip:` guard to stay faithful to production.

The cleaner shape follows the precedent `on_persisted_ip_invalidated` already set (a dedicated callback instead of a sentinel on `on_ip_change`): a `monitor.clear_resolved_addresses(name)` method owning the set-only dedupe, paired with an `on_resolved_addresses_cleared: Callable[[str], None]` callback. That deletes the `if not ip:` branch from `apply_ip`, the sentinel arm from the shared dedupe, the conditional in the conftest recorder, and the sentinel prose on `IPChangeCallback`; the single caller (`mdns._verify_removed`) becomes self-documenting. `apply_ip_addresses([])` is a second spelling of the clear with zero production callers and would fold into the same seam.

Deliberately not done in #2034 to keep the bugfix diff minimal; the monitor constructor already takes many callbacks, so this fits naturally into the public-surface refactor tracked in #2021.

Found during the #2034 /simplify altitude review.

bdraco:

Since #2034 the empty-string form of `apply_ip` multiplexed two different operations over one wire: a real single-IP observation, and "the device left mDNS, clear the resolved set but keep the last-known `Device.ip`". That sentinel had to be decoded in three places (the monitor's dedupe, the controller callback, and the conftest recorder, which needed its own `if ip:` guard to stay faithful).

This gives the clear op its own seam, following the precedent `on_persisted_ip_invalidated` set:

- `monitor.clear_resolved_addresses(name)` owns the set-only dedupe and fires a dedicated `on_resolved_addresses_cleared` callback; `mdns._verify_removed` calls it directly and reads as what it does.
- `on_ip_change` is back to a single unconditional loop with no sentinel arm; `state_callbacks.on_resolved_addresses_cleared` handles the clear (skip already-clear siblings, keep `device.ip` and disk, fire DEVICE_UPDATED).
- `apply_ip("")` and `apply_ip_addresses([])` now raise `ValueError` pointing at the new seam instead of silently misbehaving; no production caller passes either (MQTT gates on truthiness, every `apply_ip_addresses` caller guards non-empty).
- The conftest recorder drops its production-mirroring conditional and records the clear as its own tuple.

Behavior-free for every production path; the only observable change is that programmer misuse of the empty forms fails loud.

- fixes https://github.com/esphome/device-builder/issues/2036

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/_device_state_monitor/controller.py
+++ b/esphome_device_builder/controllers/_device_state_monitor/controller.py
@@ -76,9 +76,17 @@ SourceChangeCallback = Callable[[str, ReachabilitySource], None]
 # mDNS IP resolution callback. ``primary`` is the IPv4 we lock onto
 # for ICMP / OTA cache args (or the first scoped IPv6 when no V4 is
 # present); ``addresses`` is the announced set in zeroconf's
-# ``parsed_scoped_addresses`` order. Empty primary + empty list clears
-# only the resolved set; ``device.ip`` keeps the last-known primary.
+# ``parsed_scoped_addresses`` order. Both are always non-empty — a
+# confirmed loss of resolution flows through
+# ``ResolvedAddressesClearedCallback`` instead.
 IPChangeCallback = Callable[[str, str, list[str]], None]
+
+# Confirmed loss of mDNS resolution for *name*: drop the resolved
+# ``ip_addresses`` set. ``device.ip`` keeps the last-known primary
+# (RAM mirrors the metadata sidecar) so the api_reviver can repair a
+# mid-process mDNS death; only its identity-verified invalidation
+# clears that value.
+ResolvedAddressesClearedCallback = Callable[[str], None]
 
 # mDNS ``version`` TXT change.
 VersionChangeCallback = Callable[[str, str], None]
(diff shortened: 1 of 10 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

Issue #2044, opened 2026-07-14
Give the monitor a confirmed-offline seam instead of mdns composing ledger teardown


bdraco:
`MdnsSource._verify_removed`'s confirmed-miss branch hand-composes the monitor's "confirmed gone" teardown from ledger internals — four mutations in sequence: `apply(name, OFFLINE, "mdns")`, the resolved-address clear, `forget(name)`, and `state.reachability.clear(name)`. Which per-name ledgers exist is monitor knowledge, not mdns knowledge, and a sibling composition already exists in `controllers/devices/scan_change.py` (`reachability.clear` + `forget` on YAML deletion).

Cost: adding a new per-name ledger to `MonitorState` requires remembering to update the mDNS source's OFFLINE branch, and a future confirmed-gone source (e.g. MQTT LWT) would copy the quartet.

Follow-up shape: a monitor-level method owning the composition (e.g. `confirmed_offline(name, source)`), with `_verify_removed`'s branch collapsing to one call. The OFFLINE-semantics asymmetry (verify-before-demote, #1748/#1776/#1993) stays exactly where it is — only the mutation bundle moves down a layer.

Flagged during the mdns.py /simplify altitude review as out of scope for the module-cleanup PR (needs monitor API design + cross-module reroute).

Issue #2043, opened 2026-07-14
Cut the O(N-squared) per-sweep terms in the mDNS hot paths


bdraco:
A /simplify efficiency pass over `_device_state_monitor/mdns.py` found two O(N²)-per-sweep terms plus two cheap tidy-ups, all measurable only at large fleet scale (~1000 devices), so deferred out of the module-cleanup PR:

1. **`has_live_ptr` via `should_ping` makes the sweep O(N²).** zeroconf's `current_entry_with_name_and_alias` materialises and scans every PTR under the type-domain key (one per ESPHome node on the LAN); `_select_ping_targets` calls `should_ping` per device per sweep, and in a healthy fleet every device hits the `has_live_ptr` branch. 1000 devices ≈ 1M record touches + 1000 kilo-entry list allocations per sweep, on the event loop. Fix shape: one `get_all_by_details(_ESPHOME_SERVICE_TYPE, _TYPE_PTR, _CLASS_IN)` pass per sweep building a live-alias set, threaded through the sweep (or memoised on `MdnsSource` behind a sweep token).

2. **Full reachability snapshot built per positive observation with zero drawer subscribers.** `monitor.apply` → `reachability.observe` → `on_observation` unconditionally builds the whole `MdnsCacheInfo` snapshot (five cache walks incl. the O(fleet) PTR scan from item 1) and fires `DEVICE_REACHABILITY` into a listener set that is empty whenever no drawer is open — the common state. Fix shape: an `EventBus.has_listeners(event_type)` accessor gating snapshot construction; the drawer subscription already re-sends a snapshot on subscribe, so nothing is lost.

3. **`resolve_api_mdns_targets` uses full `get_mdns_cache_info(...)` as an existence check** (`shared.py`). A `has_cached_trace(name)` returning on the first non-empty record bucket (A → AAAA → SRV → TXT, PTR last) skips the decode/allocation. Only measurable in a large fleet stuck on the `ping` source (mDNS-dark restart).

4. **`get_cached_addresses` constructs a fresh `AddressResolver` per call**, up to twice per `.local` device per sweep. A `dict[str, AddressResolver]` keyed on resolver name is safe to reuse (`load_from_cache` re-reads the live cache each call); prune on `forget` to avoid growth on renames.

Items 1 and 2 are the priority; 3 and 4 ride along.

PR #2045, opened 2026-07-14
Tidy MdnsSource: shared seams, owned lifecycle, and truthful docs

bdraco:

A /simplify pass over the whole `_device_state_monitor/mdns.py` module (four review angles: reuse, simplification, efficiency, altitude). Applies the behaviour-preserving cleanups; the fleet-scale efficiency findings and a monitor-API change are deferred to #2043 / #2044.

- **`cache_apply_or_resolve` goes public and absorbs the last duplicate.** `ImportableDiscovery.on_http_service_state_change` still carried a verbatim copy of the cache-first tail #2042 unified — the unification had stopped at the module boundary. The helper is now a public seam (beside `resolve_then`) and the importable HTTP path calls it.
- **`importable.setup()` moves to `DeviceStateMonitor.start()`.** MdnsSource was constructing a sibling source's discovery object inside its own `start()`; the monitor now owns all source lifecycle. (Nuance: setup now also runs when zeroconf fails to start — previously the early return skipped it. Benign: with no browser, `import_state` stays empty and every importable read already returns empty.)
- **The `_dispatch` closure becomes the bound method `_on_browser_event`** — `start()` sheds 13 inline lines; handler registration is unchanged.
- **`close_zeroconf` reuses `helpers.async_.drain_tasks(log_exceptions=True)`** instead of inlining the cancel/gather/log idiom (its comment even admitted "mirroring the drain in `DeviceStateMonitor.stop`"). Only delta: a crashed interface monitor now logs at WARNING instead of DEBUG during shutdown, which is arguably the right level.
- **`has_live_ptr` drops a dead expiry re-check.** zeroconf's `current_entry_with_name_and_alias` filters expired entries itself (verified against zeroconf 0.150's `_cache.py`), so the `is_expired` re-check could never fire; `_cached_ptr`'s "expired or not" docstring said the opposite of what the API does and is corrected, and `get_mdns_cache_info`'s don't-filter-expired comment now notes the PTR row is live-only. The pinning test dropped its unreachable expired-PTR branch.
- **`_cached_txt_properties` self-guards on `self._zeroconf`** like its two sibling cache walkers, instead of threading a `zc` parameter its only caller walrused out.
- **`_MDNS_REFRESH_PADDING_SECONDS` moves next to the invariant it protects** (`refresh_mdns`'s cache-short-circuit contract in mdns.py) instead of living in controller.py, which never used it; the package re-export path is unchanged for the `devices/reachability.py` consumer.
- Comment/docstring drift: the esphomelib handler no longer restates the helper's contract two hops away; the `_apply_txt_properties` tri-state comment drops its editing-history framing (the four load-bearing cases stay); `resolve_then`'s docstring drops the stale two-caller inventory (it now has four).

Deliberately **not** applied, per the reviews' own analysis: folding `resolve_and_claim` into the helper (documented decision from #2042 — awaited variant with a different timeout), dropping `resolve_then`'s `zeroconf` parameter (not provably equivalent across the `close_zeroconf` shutdown race), and removing `get_cached_addresses`' `normalize_hostname` hop (relies on zeroconf cache-key internals). Deferred as issues: #2043 (the two O(N²)-per-sweep terms plus two hot-path tidy-ups), #2044 (a monitor-level confirmed-offline seam replacing `_verify_removed`'s hand-composed ledger teardown).

- follow-up to #2042; defers #2043 and #2044

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/_device_state_monitor/__init__.py
+++ b/esphome_device_builder/controllers/_device_state_monitor/__init__.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 # ``from ..controllers._device_state_monitor import X`` callers
 # (including tests that reach for private constants and module
 # helpers) keep working unchanged across the split arc.
-from .controller import _MDNS_REFRESH_PADDING_SECONDS as _MDNS_REFRESH_PADDING_SECONDS
 from .controller import DeviceStateMonitor as DeviceStateMonitor
 from .helpers import _decode_txt_bytes_to_sorted_pairs as _decode_txt_bytes_to_sorted_pairs
 from .helpers import device_name_from_service as device_name_from_service
+from .mdns import _MDNS_REFRESH_PADDING_SECONDS as _MDNS_REFRESH_PADDING_SECONDS
--- a/esphome_device_builder/controllers/_device_state_monitor/controller.py
+++ b/esphome_device_builder/controllers/_device_state_monitor/controller.py
@@ -54,11 +54,6 @@ from .shared import _SOURCE_PRIORITY, should_ping
 _LOGGER = logging.getLogger(__name__)
 # Cap on draining the ping / API-info / resolve tasks at shutdown.
 _STOP_DRAIN_TIMEOUT = 2.0
-# Padding added to the cached A record's TTL when the drawer's
-# refresh loop schedules its next probe. Sleeping ``ttl + this``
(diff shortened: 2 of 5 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #2059, opened 2026-07-14
Refresh stale docs caches in sync_components and cover them in --clean

### LINKED ISSUE #2053: sync_components docs caches never invalidate, so local regens use stale image/description sources
Found while regenerating the catalog for the web_server auth follow-up (esphbot review on #2051).

`load_image_map` (and the MDX docs-repo fetch) cache under `.cache/` with a bare `if not cache_file.exists()` guard — nothing ever refreshes them. A checkout whose cache dates from May produced a regen that silently *dropped* `image_url` for components whose docs-index rows were added since (imu.svg, pcm5122.svg, speaker.svg, xdb401.jpg), while CI (no cache) kept them. The `--clean` flag only wipes `esphome-schema-*` dirs, so there's no supported way to refresh the docs-derived caches short of knowing to `rm` them by hand.

Suggest either including the docs caches in `--clean`, or giving them a max age (the schema bundles are version-keyed so immutable caching is right for them; the docs index and MDX repo are moving targets).

bdraco:

The docs `components/index.mdx` cache under `.cache/` was guarded by a bare exists() check, so a local checkout never refreshed it; a stale copy silently dropped `image_url` for components added to the docs index since the cache was written, while CI (no cache) kept them.

The cache now refreshes once it is older than 24 hours; if the refresh fetch fails the stale copy is still used, and the empty map (which trips the sentinel abort) only happens when there is no cached copy at all. `--clean` now also wipes the docs-derived caches (the index file and the esphome.io clone), not just the schema bundles. The version-keyed schema bundles keep their immutable caching.

- fixes #2053

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/script/sync_components.py
+++ b/script/sync_components.py
@@ -45,6 +45,7 @@ import re
 import shutil
 import sys
 import textwrap
+import time
 import unicodedata
 import urllib.request
 import zipfile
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #2064, opened 2026-07-14
Model boolean enum unions as full option lists in the catalog

### LINKED ISSUE #2057: zigbee wipe_on_boot never renders in the visual editor: default_with gate hides the field, and its enum is half-modelled
Repro: a config with `zigbee: {wipe_on_boot: once}` on any non-nRF52 board — the structured editor never shows a **Wipe On Boot** field, even with advanced settings on. Two stacked causes:

1. **`default_with` gates leak into field visibility.** Upstream validates the field with `cv.OnlyWith(..., "nrf52")` *for its default only*, but the sync maps that to `depends_on_component: "nrf52"`, and the frontend's visibility predicate (`config-validation.ts` — entry hidden unless `depends_on_component` is in `presentComponents`) hides the whole field on every board without an `nrf52` component. "This default only applies with X" and "this field only exists with X" are different contracts; the model has one knob for both. Every `default_with`-carrying field on a component that can appear without the gating component has this problem.

2. **The boolean|enum union is half-modelled.** Upstream accepts `true`/`false` *or* `once`; the schema bundle emits `type: enum` with a single `once` value plus the `default_with`, and the catalog lands as the contradictory `type: boolean` **with** a one-entry `options` list. Even if the visibility gate were fixed, the dropdown would offer only `once` with no way to express the boolean states.

Pre-existing on `main` (not from #2055 — that PR only moved the option's sentence docs from label to description). Found while browser-testing #2055 with a zigbee test config.

bdraco:

Upstream validates `zigbee.wipe_on_boot` with `cv.Any(cv.boolean, cv.one_of("once"))`; the schema bundle only carries the `once` enum half, and the live introspection pass then retyped the entry to `boolean` while leaving the one entry options list in place, so the editor rendered a dropdown offering only `once` with no way to express the boolean states. `mqtt.discovery` has the same shape with its `CLEAN` value.

The sync now folds the boolean half into the options instead of retyping the entry: the field stays `string` and gains `true` and `false` options, and a bool default is stringified so it matches an option value. `check_catalog.py` gains a guard that fails when any entry is typed `boolean` while carrying options; the catalog is regenerated for `zigbee` and `mqtt`.

The issue's other half, the `default_with` gate hiding the field off nRF52, is intended behavior; upstream also validates the field with `cv.requires_component("nrf52")`, so a value is invalid on any other board and hiding it there is correct. Verified in the live editor: the field renders as a true/false/once dropdown with `false` marked default, and picking a value round trips through save.

- fixes https://github.com/esphome/device-builder/issues/2057

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #2094, opened 2026-07-16
Run the remote build-environment reset as a tracked full clean-all job

bdraco:

Lets a paired offloader reset a build server's whole build environment as a tracked job.

This closes a real recovery gap for the headless build servers added this month: `--remote-build-only` serves no HTTP dashboard at all, only the peer-link listener, so such a server has no UI from which to run `firmware/reset_build_env`. When its toolchain wedges (corrupt PlatformIO cache, broken venv) the only recovery today is shell access to the machine, and reset-build-env is exactly the last resort meant to save you from dropping to a shell. With this PR the operator of any paired dashboard can trigger the server's full local reset, `esphome clean-all` plus every cached esphome venv, from their own UI.

This is the rework of the closed #2088. That design was a synchronous request/ack doing a scoped per-offloader wipe; review showed the scoped wipe could not recover a wedged server (the shared PlatformIO toolchain and venvs were out of reach) and the full clean takes 2 to 15 minutes, far past any sane ack window. The new design runs it through the job API instead:

- `remote_build/reset_peer_build_env {pin_sha256}` enqueues a REMOTE-source RESET_BUILD_ENV mirror job on the offloader and returns it, so the frontend follows it like any build.
- The runner sends one bundle-less `reset_build_env {job_id}` frame; the receiver enqueues its own local reset job tagged `remote_peer` + `remote_job_id`, and the existing JobFanout streams the clean-all output and terminal state back into the mirror job. Cancel rides the existing `cancel_job` correlation.
- The receiver refuses `busy` while any job is active or a bundle is mid-upload. A full reset wipes the shared toolchain every tenant compiles from, and a peer must never force-cancel other dashboards' work, so this deliberately does not reuse the local command's cancel-all sweep.
- Capability-gated: the receiver advertises `reset_build_env_supported` on the handshake; old receivers never send it and old offloaders never send the frame.
- A peer-link loss mid-reset fails the mirror job with "session lost" while the receiver's wipe runs to completion; the reset is idempotent so a retry just finds less to delete.
- `upload_blocked` now exempts REMOTE-source resets; they wipe the receiver's tree, not any local artifact a flash reads.

Note for reviewers: because the reset wipes the shared toolchain and every cached esphome version, it affects every dashboard paired to that server. The frontend confirm dialog states this plainly.

- fixes n/a (field report; supersedes closed #2088)

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1256

## Checklist


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
Agreed on the duplicate-reset guard naming — it follows the pre-existing submit_job/download_artifacts convention, so renaming all three guards together is its own small concern rather than this PR's. Filed #2096 to introduce a properly named sibling (e.g. DuplicateRequestError) with the same WS error-code mapping.

PR #2097, opened 2026-07-16
DRY the peer-link-client request registration and session-loss drain

### LINKED ISSUE #2092: DRY the peer-link-client request/ack future machinery
The offloader-side `PeerLinkClient` now runs three independent request/ack flows over the peer link — `submit_job`, `download_artifacts`, and (as of #2088) `reset_build_env` — and each re-implements the same future-based machinery:

- a per-request `dict[str, Future]` on the client (`_submit_job_acks`, `_artifacts_downloads`, `_reset_env_acks`)
- register-future-then-send, `asyncio.wait_for` with a timeout
- a session-loss drain block in `_run_session_loops`'s `finally` that force-completes the pending futures with `SubmitJobSessionLostError` (three near-identical blocks at `peer_link_client/client.py:677/687/697`)

Each verb rolls its own copy. A small "pending-ack registry" abstraction (register / resolve-by-id / drain-on-close) would collapse the three into one, and the session-loss `finally` would drain a single structure instead of three.

Not urgent — the duplication is the established per-verb idiom and each copy is correct. Surfaced during a `/simplify` pass on #2088.

Behaviour-free refactor.

bdraco:

Finishes DRYing the peer-link-client request/ack future machinery. #2094's simplify pass already consolidated the two ack verbs (`submit_job`, `reset_build_env`) onto shared register/await/drain helpers; this folds the third flow, `download_artifacts`, into the same shape.

`_register_ack_future` becomes `_register_pending[EntryT]` taking an entry factory, so it also serves the download map whose entries carry assembler state alongside the future. `_drain_pending_acks` becomes `_drain_pending` taking `(job_id, future)` pairs, so the hand-rolled download drain block in `_run_session_loops`'s `finally` collapses into the same call as the two ack drains. Session-lost drain messages are byte identical for all three verbs; the duplicate-registration message is now one template ("request already registered"). Dropping the download drain block also took `_run_session_loops` back under the complexity threshold, so its `noqa: C901` is gone.

Deliberately unchanged: `download_artifacts` still has no ack timeout (streaming a large tarball can legitimately exceed the 60s ack window; the heartbeat plus session-loss drain bound it), so `_await_ack` stays with its two callers. The three maps stay per-verb and precisely typed rather than merging into one registry class, which would trade away the `Future[SubmitJobAckFrameData]` / `Future[ResetBuildEnvAckFrameData]` / `_DownloadArtifactsState` typing for casts.

Behaviour free; no wire or API change.

- fixes https://github.com/esphome/device-builder/issues/2092

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/remote_build/peer_link_client/_submit.py
+++ b/esphome_device_builder/controllers/remote_build/peer_link_client/_submit.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+from collections.abc import Callable
 from typing import TYPE_CHECKING, Any, Literal, cast
 
 from ....helpers.peer_link_bundle import (
@@ -60,7 +61,12 @@ async def submit_job(
     already.
     """
     channel = _require_open_channel(client, label="submit_job")
-    ack_fut = _register_ack_future(client._submit_job_acks, job_id, label="submit_job")
+    ack_fut = _register_pending(
+        client._submit_job_acks,
+        job_id,
+        asyncio.get_running_loop().create_future,
+        label="submit_job",
+    )
     try:
         await _send_submit_job_frames(
             client,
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #2126, opened 2026-07-16
Fold the remaining toolchain literal into the shared constants

### LINKED ISSUE #2123: Fold the remaining toolchain literal in artifacts_tarball into the shared constants
`controllers/remote_build/artifacts_tarball.py:209` compares the build toolchain against a bare string literal:

```python
if getattr(storage, "toolchain", None) != "esp-idf":
```

Two things to fold in once #2105 lands:

- [ ] Use the shared constant. #2105 adds `TOOLCHAIN_ESP_IDF` / `TOOLCHAIN_SDK_NRF` to `constants.py`, matched as plain strings rather than through `esphome.const.Toolchain` members, because the floor has no ceiling and that enum keeps gaining members; reading one a supported esphome lacks would raise for every caller, not just the platform it names. This site encodes the same "is this a native IDF build" decision and should read from the same place, so a future toolchain is one edit rather than three.

- [ ] Drop the `getattr` default. `StorageJSON.toolchain` landed in esphome 2026.5.1, which is now the declared floor (#2120), so the attribute is always present and the `None` default is unreachable. Same class as the shims in #2122, but this one is the DRY half rather than a dead branch, so it is tracked separately.

Blocked on #2105, which introduces the constants. Behaviour is unchanged either way; a sidecar recorded as `esp-idf` compares equal to `TOOLCHAIN_ESP_IDF` because the constant is that same string.

bdraco:

`artifacts_tarball.py` compared the build toolchain against a bare `"esp-idf"` literal; that is the same native-IDF decision the backtrace decoder already reads from `TOOLCHAIN_ESP_IDF`, so it now reads the constant too. The unpack side in `remote_artifacts_materialise.py` carried the identical literal, so it gets the same treatment, and a future toolchain is one edit rather than several.

The `getattr(storage, "toolchain", None)` default goes away on the pack side. `StorageJSON.toolchain` landed in esphome 2026.5.1, under the declared 2026.6.0 floor, so the attribute is always present and the default was unreachable. The unpack side keeps its `.get` default on purpose; it reads a wire dict shipped by the receiver, where the key can legitimately be absent.

That floor also made `HAS_NATIVE_IDF_TOOLCHAIN` in `tests/conftest.py` always true, so it is gone and the three native-IDF tests it guarded now always run.

No behaviour change either way; `Toolchain.ESP_IDF.value` is exactly `"esp-idf"`, so the constant compares equal to what a sidecar records.

#2105 introduced the constants and has since landed, so this is rebased onto main and stands alone.

- fixes https://github.com/esphome/device-builder/issues/2123

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/constants.py
+++ b/esphome_device_builder/constants.py
@@ -128,9 +128,11 @@ FEATURED_EXCLUDED_CATEGORIES: frozenset[str] = frozenset({"core", "ota", "update
 # stores. Matched as strings rather than through ``esphome.const.Toolchain``
 # to keep this module a stdlib-only leaf; the wire values don't change.
 # Shared because the dashboard's spawn gate (``controllers/devices/
-# backtrace.py``) and the helper child's idedata decision (``helper_cli.py``)
-# encode one contract and must agree; stdlib-only home, so the child pays
-# nothing to import it.
+# backtrace.py``), the helper child's idedata decision (``helper_cli.py``) and
+# the offload pack / unpack pair (``controllers/remote_build/
+# artifacts_tarball.py``, ``helpers/remote_artifacts_materialise.py``) encode
+# one contract and must agree; stdlib-only home, so the child pays nothing to
+# import it.
 TOOLCHAIN_ESP_IDF = "esp-idf"
 TOOLCHAIN_SDK_NRF = "sdk-nrf"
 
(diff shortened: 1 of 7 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

Issue #2122, opened 2026-07-16
Remove the esphome compatibility shims the 2026.5.1 floor makes dead


bdraco:
#2120 raised the declared floor to `esphome>=2026.5.1`. Several compatibility shims exist only to support esphome older than that, so they are now unreachable. Three of them carry a TODO or a comment naming exactly this moment as the trigger to remove them.

Verified against the `2026.5.1` tag: every symbol each shim falls back for is present at the floor, so no fallback branch can execute.

### The shims

- [ ] **`controllers/firmware/_upload_targets_fallback.py`** (whole module) plus the `try/except ModuleNotFoundError` around it in `controllers/firmware/helpers.py:50-55`, and `tests/controllers/firmware/test_upload_targets_fallback.py`. The module's own header says `TODO: remove after the esphome dep floor is past the release that ships esphome.upload_targets`; `esphome/upload_targets.py` landed in 2026.5.0, and the test file says `TODO: delete with _upload_targets_fallback`.

- [ ] **`helpers/device_yaml/_loading.py:37-63`**, the `resolve_packages` guard and the `do_packages_pass` / `merge_packages` two step fallback under it. `resolve_packages` (esphome/esphome#16235) is present at 2026.5.1. The comment says "Once the dashboard's dep floor moves past the release that shipped `resolve_packages`, the entire fallback path can be deleted in one commit". Note the second `try/except` there is forward compat rather than backward compat, so read the comment before cutting; the plan it describes drops both and calls `resolve_packages` directly.

- [ ] **`controllers/devices/helpers.py:10-17`**, the `friendly_name_slugify` fallback to `esphome.dashboard.util.text`. It lives in `esphome.helpers` from esphome/esphome#16206 and is present at 2026.5.1. The comment says "once those drop the hard dependency goes away".

- [ ] **`controllers/devices/helpers.py:19-22`**, the `FRIENDLY_NAME_MAX_LEN = 120` fallback. `esphome.core.config.FRIENDLY_NAME_MAX_LEN` is present at 2026.5.1.

- [ ] **`helpers/device_yaml/_loading.py`**, the `getattr(storage, "core_platform", None)` default and its test `tests/test_device_yaml.py::test_load_device_handles_storage_without_core_platform_attr`. `StorageJSON.core_platform` (esphome/esphome#9028) is present at 2026.5.1. Flagged as a follow up in #2120, which left the shim in place and only corrected the comment that justified it by citing the old floor.

### Not in scope

[...]

PR #2131, opened 2026-07-17
Raise the esphome floor to 2026.6.0

bdraco:

Raises the `esphome` floor from 2026.5.1 to 2026.6.0.

#2120 moved the floor off 2024.1.0 and settled on 2026.5.1 because that is where `StorageJSON.toolchain` lands, which was the oldest version the dashboard could actually read. That answered "what is the earliest version that works" rather than "what do we support", and the two are not the same question for a project that has not shipped and does not publish old versions.

The concrete thing that prompted this: on `esphome == 2026.5.1` exactly, esp32's `_decode_pc` calls `toolchain.get_idedata(config)` unconditionally, with no branch for the native ESP-IDF toolchain. 2026.5.2 added the `CORE.using_toolchain_esp_idf` branch that resolves addr2line off the IDF toolchain instead. Verified against both sources rather than inferred:

- 2026.5.1 — `_decode_pc` calls `get_idedata` whatever the toolchain.
- 2026.5.2 — branches on `using_toolchain_esp_idf` first, never touching idedata.

So the backtrace decoding in #2105 cannot work on a native-ESP-IDF build at the current floor: it degrades to `decode_failed` after spawning a helper child that was never going to succeed. Supporting one superseded patch release would mean either wearing a futile subprocess or carrying a version-conditional gate, and a shim for a single old patch is the kind of thing #2122 exists to delete.

2026.6.0 rather than 2026.5.2 because there is no reason to hold the floor at the oldest version that technically works. The catalog is already generated against the 2026.7.0 schema bundle, so this narrows the gap between the floor and what the catalog describes.

- n/a

## Types of changes

- [x] Dependencies bump — `dependencies`

## Frontend coordination

## Checklist

## Verification

Full suite on this branch: 6695 passed, 11 skipped. The one failure, `test_libretiny_bk7231n_compile_download_round_trip`, reproduces on a clean `main` without this change and is unrelated.

Nothing else in the tree pins 2026.5.1; the remaining hits are unrelated version fixtures in `tests/test_version_compat.py` and the sync-emit tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ version = "0.0.0" # Set by GH action on release
 
 [project.optional-dependencies]
 esphome = [
-  "esphome>=2026.5.1",
+  "esphome>=2026.6.0",
 ]
 test = [
   "blockbuster>=1.5.5,<1.6",


bdraco:
@esphbot review

Floor bump to 2026.6.0. Rationale is in the description; the short version is that 2026.5.1 was chosen in #2120 as the oldest version that technically works (`StorageJSON.toolchain` lands there), which is a different question from what we support.

The concrete trigger, verified against both sources rather than inferred: at `esphome == 2026.5.1` exactly, esp32's `_decode_pc` calls `get_idedata` unconditionally; 2026.5.2 added the `using_toolchain_esp_idf` branch. So #2105's decoding cannot work on a native-ESP-IDF build at the current floor. Worth noting this corrects a claim in #2105's own `_pin_idedata` docstring, which said that branch arrived in 2026.6 — it did not, and you repeated that figure back to me from the docstring in your review there. Both are fixed now.

Went to 2026.6.0 rather than the minimum 2026.5.2 because holding the floor at the oldest version that technically works has no upside here, and the catalog is already generated against the 2026.7.0 schema bundle.

Worth a look at whether this makes anything in the tree floor-dead that should go to #2122. One I already know of, once #2105 lands: `_pin_idedata` pins an `IDEData({})` sentinel even for toolchains that should never consult idedata, and its justification is specifically the 2026.5.1 behaviour this bump drops. It stays as cheap insurance, but the docstring reasoning will want revisiting.

Suite on this branch: 6695 passed, 11 skipped. The single failure, `test_libretiny_bk7231n_compile_download_round_trip`, reproduces on a clean `main` and is unrelated.

PR #2132, opened 2026-07-17
Raise the esphome floor to 2026.7.0

bdraco:

Raises the `esphome` floor from 2026.6.0 to 2026.7.0, following #2131.

#2131 went to 2026.6.0 on the reasoning that there was no point holding the floor at the oldest version that technically works. The same reasoning goes one further: 2026.7.x has shipped, so 2026.6.x is no longer receiving patches, and a floor that admits an unpatched series is a floor we would not actually support.

This also makes the three versions that should agree actually agree. The catalog is generated against the 2026.7.0 schema bundle, and `boards.index.json` is stamped `esphome_version: 2026.7.0` — which `sync_boards.py` already refuses to run against a mismatched install. The floor was the odd one out.

| | before | after |
|---|---|---|
| `pyproject.toml` floor | 2026.6.0 | 2026.7.0 |
| `boards.index.json` stamp | 2026.7.0 | 2026.7.0 |
| `components.index.json` schema | 2026.7.0 | 2026.7.0 |

- follows #2131

## Types of changes

- [x] Dependencies bump — `dependencies`

## Frontend coordination

## Checklist

## Verification

Run against esphome **2026.7.0** rather than the 2026.6.2 that was installed here, so the suite reflects the floor this declares instead of a version it no longer admits: 6710 passed, 10 skipped.

The one failure, `test_libretiny_bk7231n_compile_download_round_trip`, reproduces on a clean `main` without this change and is unrelated to the floor.

Nothing else in the tree pins the floor; the remaining `2026.6` / `2026.5` hits are unrelated version fixtures in `tests/test_version_compat.py` and the sync-emit tests.


bdraco:
@esphbot review

Follows #2131, which the maintainer has already merged at 2026.6.0. 2026.7.x has shipped, so 2026.6.x is no longer getting patches, and a floor that admits an unpatched series is not a floor we would actually support.

This also makes the three versions that should agree actually agree. `boards.index.json` is stamped `esphome_version: 2026.7.0` and `components.index.json` carries `esphome_schema_version: 2026.7.0`; `sync_boards.py` already refuses to run unless the installed esphome matches that stamp. The floor was the odd one out at 2026.6.0.

On verification, worth being explicit because it changes what the numbers mean: this venv had esphome **2026.6.2**, which this PR makes unsupported. Reporting a green suite from it would have been measuring a version the floor no longer admits, so I upgraded to 2026.7.0 first and ran there. 6710 passed, 10 skipped. The single failure, `test_libretiny_bk7231n_compile_download_round_trip`, reproduces on a clean `main` and is unrelated.

I also re-ran #2105 against 2026.7.0 for the same reason, since its decode paths were originally verified on 2026.6.2: 71 passed, mypy clean, and `_decode_pc` still carries the `using_toolchain_esp_idf` branch at this version.

Once both this and #2105 land, `_pin_idedata`'s `IDEData({})` sentinel is worth a look for #2122: its justification is specifically the 2026.5.1 behaviour, which is now three floors below us. It stays as cheap insurance, but the docstring reasoning is dead.

esphbot (approved):
> [!TIP]

PR #2133, opened 2026-07-17
Run CI pytest with uv run --no-sync

bdraco:
# What does this implement/fix?

Adds `--no-sync` to every `uv run pytest` invocation in CI.

Each pytest job builds its venv explicitly (`uv venv` + `uv pip install -e '.[esphome,test]'`, plus the channel-specific esphome install where the matrix asks for one). A bare `uv run` then re-locks the project universally, resolving the `esphome` extra for every Python and platform `requires-python` admits, not just the environment the job runs in. That re-lock is pure overhead on green runs and a false failure otherwise: esphome 2026.7.0 pins `cryptography==49.0.0` while its pinned `esptool==5.3.1` caps `cryptography<49.0.0` on Intel macOS, so any universal resolve of `esphome>=2026.7.0` is unsatisfiable and every pytest job dies before collecting a test (surfaced on #2132; we have no Intel macOS runner in the matrix).

`--no-sync` runs pytest in the prepared venv as-is. Verified locally: with the floor at 2026.7.0, `uv lock` reproduces the CI failure verbatim and `uv run --no-sync` executes cleanly.

**Related issue or feature (if applicable):**

- unblocks #2132

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [x] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Frontend coordination

- [x] No frontend change needed
- [ ] Companion frontend PR: esphome/device-builder-frontend#<number>

## Checklist

- [x] The code change is tested and works locally.
- [x] Pre-commit hooks pass (`ruff`, `codespell`, yaml/json/python checks).
- [ ] Tests have been added or updated under `tests/` where applicable.
- [x] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (regenerate via `script/sync_components.py` if a sync is needed).
- [x] Architecture-level changes are reflected in `docs/ARCHITECTURE.md` and/or `docs/API.md`.

PR #2135, opened 2026-07-17
Restore the Windows build-path fix for the native ESP-IDF toolchain

bdraco:

esphome 2026.7.0 made the native ESP-IDF toolchain the default for esp32 `esp-idf` configs (2026.6.0 defaulted to PlatformIO). That toolchain installs into `platformdirs.user_cache_dir("esphome")/idf`, which on Windows is `C:\Users\<user>\AppData\Local\esphome\Cache\idf`; the install nests ~209 chars of paths below that, esphome only warns when the projection exceeds MAX_PATH, and ESP-IDF does not support spaced install paths (`First Last` profiles). Our Windows relocation covered `ESPHOME_DATA_DIR` and `PLATFORMIO_CORE_DIR` but not the IDF prefix, so the just-works guarantee had a hole for the new default toolchain.

`windows_short_build_paths` now also points `ESPHOME_ESP_IDF_PREFIX` at `C:\esphb\<id8>\idf`, with the same rules as the pio dir: a user-set var is respected, an existing default cache is migrated in once, and `esphome clean-all` removes the relocated dir (it resolves the toolchain path through the same env var).

This is also what broke the windows-real-compile job on #2133 (first run since 2026.7.0 shipped; the Monday scheduled run on main would have failed identically): the e2e pinned only the PlatformIO layout (`.pioenvs`, `PLATFORMIO_CORE_DIR`). Since both toolchains stay supported (`toolchain: platformio` is an esp32 opt-in, and pioarduino remains the path for other platforms), `test_windows_short_paths` now parametrizes its compile fixture over both: the native default and the PlatformIO opt-in each get the deep + spaced compile, `clean`, and `clean-all` assertions against their own layout. The workflow fans the suites out one per runner (native, platformio, migration) since each is a cold toolchain download plus a real compile and this is the longest CI job; wall clock stays at single-compile time. The esp8266 `test_windows_migration` keeps covering the `~/.platformio` migration path.

- unblocks the windows-real-compile job on #2133

## Types of changes

## Frontend coordination

## Checklist

## Verification

Unit coverage in `tests/test_windows_build_paths.py` drives the new branch cross-platform (sets and restores the var, respects a user-set prefix, migrates the default cache, relocates with no migration source). The real proof is this PR's own windows-real-compile run, which the path filter triggers: a deep + spaced native ESP-IDF compile, then `clean` and `clean-all`, against the relocated dirs.

### DIFF
--- a/.github/workflows/windows-real-compile.yml
+++ b/.github/workflows/windows-real-compile.yml
@@ -1,11 +1,11 @@
 name: Windows real-compile (MAX_PATH)
 
-# tests/e2e/slow/windows: a real ESP-IDF compile on a windows-latest runner that proves the
-# Windows build-data relocation (short, space-free root) builds and keeps building — covering
-# both MAX_PATH and spaced profile dirs. A cold ESP-IDF toolchain download makes this a
-# ~20-40 min job, so it is gated to changes that touch the mechanism plus a weekly drift check
-# rather than running on every PR. The default test matrix ignores tests/e2e, so this workflow
-# is the only place this e2e runs.
+# tests/e2e/slow/windows: real compiles on windows-latest runners that prove the Windows
+# build-data relocation (short, space-free root) builds and keeps building — covering both
+# MAX_PATH and spaced profile dirs. Each suite is a cold multi-minute toolchain download plus a
+# full compile, so they fan out one-per-runner (this is the longest CI job) and are gated to
+# changes that touch the mechanism plus a weekly drift check rather than running on every PR.
+# The default test matrix ignores tests/e2e, so this workflow is the only place these e2e run.
 #
 # YAML anchors (``&trigger-paths`` / ``*trigger-paths``) keep the push and pull_request path
 # filters in lockstep.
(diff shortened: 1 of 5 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All three suggestions taken in af4fcffc: empty/whitespace ESPHOME_ESP_IDF_PREFIX now counts as unset (mirroring esphome's own semantics, with a unit test pinning it), the per-dashboard IDF copy tradeoff is noted in the module docstring, and the IDF failure path has its own test keying the interrupted move on cache_idf.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Docstring reflowed to the file's fill width.

@esphbot review

bdraco:
Both silent-failure findings addressed in 2dd530ef.

1. A present-but-empty ESPHOME_ESP_IDF_PREFIX is now restored verbatim on exit instead of popped; the empty-prefix test asserts the post-exit value.
2. A failed move is now classified: when the copy verifiably completed (every file still under the source has an identical-size counterpart at the destination) and only the source delete died, the destination is adopted instead of falling back to the half-deleted source. An empty source stays ambiguous and fails closed, so the existing interrupted-move contracts are unchanged. New test simulates the delete-phase death with a completed copytree.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

bdraco:
Simplified per review: the copy-completed classifier is gone. It defended a rare double failure (cross-volume move plus delete-phase death) that both toolchain managers already self-heal (pio verifies packages, idf_tools reinstalls missing tools) and that the existing discard-and-retry semantics converge on the next run anyway. The failed-move path is back to plain fail-closed. The partial-dst discard now uses esphome.helpers.rmtree, which clears the read-only flags toolchain trees carry on Windows and which this repo already uses in env_provisioner.

@esphbot review

bdraco:
(The simplification commit is 51609c7c; the previous comment raced the push.)

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

bdraco:
All three taken in eadb95fc: the stale classifier comment is rewritten to what the mock actually pins (and why the partial dst stays: it feeds the next run's rmtree discard), a failed non-user-set IDF relocation now warns with the concrete fallback cache path so the eventual compile failure is diagnosable, and the platformdirs ImportError logs at debug so a skipped migration is distinguishable from an absent cache.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Fixed in the latest commit: the discard is now try/except with the OSError included in the warning, and the discard-failure test mocks the realistic raise (esphome's rmtree succeeds or raises; the old silent-no-op mock was an artifact of the previous ignore_errors call).

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Taken: the cache path is computed once and the warning substitutes a literal when platformdirs is unavailable, so the message never reads "will use None".

@esphbot review

esphbot (approved):
> [!TIP]

PR #2190, opened 2026-07-19
Adopt the atomic write helpers at the remaining unsafe write sites

### LINKED ISSUE #2189: Adopt the atomic write helpers at the three remaining unsafe write sites
Follow-up to #2183 / #2188. With `atomic_write_exclusive` landed, an audit of every direct write in `esphome_device_builder/` found three sites that should adopt the atomic helpers.

**To fix**

- `controllers/devices/importable.py:96`: the adoption flow writes its device YAML with a bare `open(path, "x")`, the exact shape #2188 removed from create and clone; a crash mid-write leaves a partial YAML that then blocks the retry with already-exists. Swap to `atomic_write_exclusive` (the surrounding `FileExistsError` contract is already in place).
- `helpers/auth.py:187` (`_persist`): hand-rolled tmp-write, chmod 0o600, replace; a re-implementation of `atomic_write(path, data, mode=0o600)` that also misses its Windows transient `PermissionError` retry, so a session-store save can fail spuriously under an AV or indexer hold. Swap to `atomic_write`.
- `controllers/version_history/git_repo.py:298`: the one-time `.gitignore` seed uses `write_text`; it is only written when missing, so a torn write persists and gets committed to history. Swap to `atomic_write` (small, but one line).

**Audited, no change needed**

- `controllers/remote_build/submit_job.py:812` and `controllers/devices/mutations_import_bundle.py:143`: transient bundle bytes; a torn tarball fails extraction cleanly and the next attempt overwrites.
- `helpers/remote_artifacts_materialise.py:303,340` and `controllers/devices/importable.py:257`: StorageJSON / idedata build artifacts, recoverable on the next compile or materialise (the documented criterion).
- `helpers/windows_build_paths.py:214`: two-byte marker file, recreated on next start.

`script/` was left out of scope; sync scripts are re-runnable dev tooling.

**Tests**

Each swapped site keeps its existing suites; the adoption site gains a race test mirroring create/clone's (patch the staged writer), and the auth swap drops the private `_persist` staging in favour of the shared helper's existing coverage.

bdraco:

The post-#2188 audit found three direct writes that should ride the atomic helpers; this adopts them.

- The adoption flow's device YAML now writes through `atomic_write_exclusive` instead of a bare `open("x")`, so a crash mid-adopt can't leave a partial YAML that blocks the retry; the `FileExistsError` contract is unchanged and the existing lose-the-race test now exercises the real exclusive publish.
- `SessionStore._persist` drops its hand-rolled tmp-write, chmod, replace for `atomic_write(path, data, mode=0o600)`, picking up the shared Windows transient-`PermissionError` retry it was missing; the existing test pins 0o600 on the persisted store.
- The one-time `.gitignore` seed uses `atomic_write` with an explicit `mode=0o644` (the staging tempfile is born 0o600, which would otherwise silently tighten a user-visible config-dir file).

The audited-and-fine sites are listed in the issue. No behaviour change beyond the crash-safety and the Windows retry.

- fixes https://github.com/esphome/device-builder/issues/2189

## Types of changes

## Frontend coordination

## Checklist

- [ ] Tests have been added or updated under `tests/` where applicable (the existing suites pin each site: the adopt lose-the-race test, the 0o600 session-store assertion, and the gitignore seed tests).

### DIFF
--- a/esphome_device_builder/controllers/devices/importable.py
+++ b/esphome_device_builder/controllers/devices/importable.py
@@ -11,6 +11,7 @@ from esphome.storage_json import ignored_devices_storage_path
 
 from ...helpers.api import CommandError
 from ...helpers.async_ import run_in_executor
+from ...helpers.atomic_io import atomic_write_exclusive
 from ...helpers.device_yaml import generate_adoption_yaml
 from ...helpers.json import JSONDecodeError, dumps_indent, loads
 from ...helpers.lazy_module import async_import_module
@@ -91,10 +92,9 @@ async def import_device(
             )
 
             def _write_exclusive() -> None:
-                # Exclusive-create, matching ``import_config``'s
+                # Staged exclusive-create, matching ``import_config``'s
                 # FileExistsError contract for a concurrent writer.
-                with path.open("x", encoding="utf-8") as f:
-                    f.write(content)
+                atomic_write_exclusive(path, content.encode("utf-8"))
 
             await run_in_executor(_write_exclusive)
     except FileExistsError as exc:
--- a/esphome_device_builder/controllers/version_history/git_repo.py
+++ b/esphome_device_builder/controllers/version_history/git_repo.py
@@ -34,6 +34,7 @@ from pathlib import Path
 
 import esphome_device_builder
 from esphome_device_builder.constants import is_secrets_file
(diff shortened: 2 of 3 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

Issue #2208, opened 2026-07-19
[Bug] logger.log requires Logger ID in some situations


breti:
### What happened?

1. Use this configuration:
(Generic ESP32-S3 Board configuration with `logger.hardware_uart` and `button.template` added)

```
# Board: Generic ESP32-S3 Board (Generic)
# Definition: definitions/boards/generic-esp32s3/manifest.yaml

esphome:
  name: testdev
  friendly_name: Testdev

esp32:
  variant: esp32s3
  framework:
    type: esp-idf

logger:
  hardware_uart: UART0
  baud_rate: 115200

api:
  encryption:
    key: "YHClfBJFLYDOoaUzYD+z3j3zq3wlIEQQlCyAMYvHyEM="

ota:
  - platform: esphome

wifi:
  ssid: !secret wifi_ssid
  password: !secret wifi_password
  ap:
    ssid: Testdev Fallback Hotspot
    password: "QT9vxr1SHa7T"

captive_portal:

button:
  - platform: template
    name: Testbutton
    id: button_template_1
```

2. Click on `- platform: template`
3. `+ Add Automation`
4. Which trigger? `Button => On Press`
5. `+ Add Action`
6. `Logger => Log`
7. The structured editor asks for a Logger ID and creates invalid YAML

<img width="1063" height="799" alt="Image" src="https://github.com/user-attachments/assets/b32fe5e9-0c05-42b5-a56c-40b342e3b37d" />

### Dashboard version

1.6.7

### How did you install it?

Desktop app

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

PR #2209, opened 2026-07-20
Flag which available component instances declare a YAML id

bdraco:

`automations/get_available` surfaces every configured instance with an `id`, but when the YAML declares none that id is synthesized (`logger` for a singleton, `switch_1` positional). The synthetic form is the round trip identity the parser and writer key on, so it has to stay, but the frontend can't tell it apart from a real id and the action picker's By target flow writes it into reference params, producing `logger_id: logger` and a validation error (#2208).

Each `AvailableComponentInstance` now carries `has_explicit_id`, true only when the YAML actually declares `id:`, so the frontend can skip the pre fill and let ESPHome auto resolve the instance. Companion frontend PR consumes the flag and adds a revert to auto option so configs already carrying the dangling id can be repaired from the visual editor.

- part of https://github.com/esphome/device-builder/issues/2208 (closed by the companion frontend PR)

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1317

## Checklist


bdraco:
@esphbot review

bdraco:
@esphbot review

bdraco:
Both suggestions triaged. Suggestion 1: the instance builder docstring now names the flag, 700ceb5d. Suggestion 2 is a false alarm, the numbers reference different things on purpose: #2208 is the tracking issue (cited in docs/API.md and closed by the companion frontend PR), while esphome/device-builder-frontend#1317 mentions #2209 only to point at this PR as the source of the wire flag. Both links resolve as intended.

@esphbot review

Issue #2220, opened 2026-07-20
M5Stack bluetooth proxies still ship empty pin tables, need pins_from


bdraco:
### What happened?

Follow up to #2219. The automatic donor rule gives pin tables to the six pin free boards with a unique same chip generic donor, but the two m5stack bluetooth proxies stay empty: the atom module's products expose different pin sets (`m5stack-atom-echo` curates 6 pins, `m5stack-atom-lite` 34, plus a bare `m5stack-atom` entry), so no single donor is safe to pick automatically. Pin fields on those two proxies still degrade to free text.

### Suggested improvement

Add a `pins_from: <board-id>` manifest key as the manual escape hatch: register it in `board.schema.json` (top level has `additionalProperties: false`), read it in the manifest loader or the sync's donor pass, and make it durable across proxy manifest regeneration — `sync_bluetooth_proxies.py`'s `emit_manifest` full overwrites manifests and today only preserves `full_config`, so either extend that preservation block or emit the key from the generator's identity table. Wire `m5stack-atom-lite-bluetooth-proxy` to `m5stack-atom-lite` and `m5stack-atom-s3-bluetooth-proxy` to `m5stack-atoms3`, then regenerate the two bodies.

### Dashboard version

dev (2026.7.0)

### How did you install it?

Dev checkout

### Where is it running?

macOS

PR #2221, opened 2026-07-20
Inherit pin tables from the same-chip generic board

### LINKED ISSUE #2219: Pin fields lose the GPIO dropdown on package-adopted devices with no local board
### What happened?

On a package adopted device (the stock bluetooth proxy config, `packages: esphome.bluetooth-proxy: github://...`), every pin typed field in the structured editor renders as a plain text input instead of the GPIO dropdown. Reproduced with the UART Bus editor's RX Pin.

The chain: the device's `esp32:` block lives inside the remote package, so the local YAML carries no platform or board; the backend's `_derive_board_id_from_yaml` returns an empty `board_id`; the frontend never fetches a board; and the pin renderer falls back to free text whenever it has no board pin table. The catalog itself is fine (`uart.rx_pin` is `type: pin`).

### Suggested improvement

This is a generic ESP32, so the pin range is known for sure. Two complementary fixes:

1. After the first compile the backend already knows the platform: `StorageJSON.target_platform` is read by `controllers/config/controller.py` and could seed the device's `board_id` to the platform's generic catalog board (`generic-esp32` and friends) whenever the YAML derivation comes up empty. That lights up the real pin picker with the generic pin table.
2. Never compiled devices could get the same treatment in the frontend: when `board_id` is empty but the platform is known (or becomes known), fall back to the generic board's pins rather than free text.

Workaround today: the device overview's Change board flow pins a board manually (`board_id_user_set`), which restores the dropdowns.

### Dashboard version

dev (2026.7.0)

### How did you install it?

Dev checkout

### Where is it running?

macOS

bdraco:

Eight board bodies shipped empty pin tables, so every pin typed field in the structured editor fell back to a free text input for devices on those boards (the wizard's Generic ESP32 Bluetooth Proxy being the reported case). They are all plain devkits whose chip already has a curated table in the catalog; esphome's board pin alias maps carry nothing for the plain devkits, so the platform augments left them empty.

`sync_boards.py` now runs a last resort donor pass after the platform augments: a board with no pins inherits the pin table of the unique same chip (`platform`/`board`/`variant`) `is_generic` donor. That fills six of the eight, including all four bluetooth proxy generics; the regenerated bodies are included. The two m5stack proxies stay empty on purpose, the atom module's products expose anywhere from 6 to 34 pins so no single donor is safe to pick automatically; #2220 tracks a manual `pins_from` escape hatch for them.

- fixes https://github.com/esphome/device-builder/issues/2219

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both findings addressed in 74399efc. The docstrings are trimmed to their contracts, no issue refs or motivation (matching the house style I should have followed the first time). Kōan's second point was the sharper one: the atom test's docstring now states the real protection, donor absence rather than the ambiguity guard, and warns that a future generic atom board must not silently flip them; the ambiguity guard itself gained a direct unit test with two synthetic same-chip generics so its safety is no longer incidental.

@esphbot review

esphbot (approved):
> [!TIP]

PR #2223, opened 2026-07-20
Add pins_from so the m5stack proxies name their pin donors

### LINKED ISSUE #2220: M5Stack bluetooth proxies still ship empty pin tables, need pins_from

bdraco:

Follow up to #2221. The two m5stack bluetooth proxies stayed pin free because their donors are curated product boards, not generics, so the automatic donor rule can never pick them (the atom module's products expose anywhere from 6 to 34 pins).

Manifests can now name the donor explicitly with `pins_from: <board-id>`. The sync applies explicit donors before the automatic generic rule; `validate_definitions` is the loud gate (the target must exist, carry a pin table, and share the chip, and a manifest cannot carry both `pins_from` and its own `pins`). The key is emitted from `sync_bluetooth_proxies`' identity table so a proxy regeneration re-emits it instead of clobbering the hand edit. Wired for `m5stack-atom-lite-bluetooth-proxy` (inherits `m5stack-atom-lite`, 34 pins) and `m5stack-atom-s3-bluetooth-proxy` (`m5stack-atoms3`, 49 pins); regenerated bodies included.

- fixes https://github.com/esphome/device-builder/issues/2220

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

bdraco:
Fixed in c01595d1. The FastestSafeLoader import chain pulls in esphome, and the validate-definitions hook runs in envs without it (that is exactly why only sync-boards is skipped on pre-commit.ci). The manifest reader now carries its own four line libyaml or fallback alias with a comment explaining why it must not import from the helpers package, and I verified the validator imports cleanly with esphome imports poisoned. The 6x loader speedup stays.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both landed in 0ac4772c. The mutual exclusivity check now flags the presence of the pins key rather than its truthiness, so `pins: []` beside `pins_from` is rejected too, with a test pinning that the key is the conflict. The scan indexes by `data.get("id") or manifest.parent.name`, so a manifest missing its id (the schema requires it, but the sync stays lenient by design) degrades to the folder name instead of a KeyError.

Also for the record, pre-commit.ci is green as of the previous head; the earlier failure was the esphome import chain fixed in c01595d1.

@esphbot review

bdraco:
Both findings handled. Suggestion 1 had already landed in 0ac4772c (the review raced the push): the scan keys by `data.get("id") or manifest.parent.name`, so an id-less manifest degrades to the folder name rather than a KeyError. The silent failure point is fixed in afd55e82: the scan now logs a warning naming the manifest and the parse error before skipping, with a test that builds a good and a broken manifest under tmp_path and pins both the surviving mapping and the warning.

@esphbot review

esphbot (approved):
> [!TIP]

PR #2229, opened 2026-07-20
Adopt the uv-managed Python composite for every CI job but CodSpeed

bdraco:

Finishes the migration to uv managed Python in CI, the same shape home-assistant core uses. The lint job, the release build, the release notes generator, and the five auxiliary workflows still installed the interpreter with actions/setup-python and layered a standalone setup-uv on top; they now all go through the existing setup-uv-python composite (setup-uv with cache-python, astral's PGO/LTO/BOLT build) that the test matrix and e2e jobs already use. Managed Python is not a system interpreter, so each converted job creates a venv and puts its bin dir on PATH; every later bare python, pre-commit, pytest, and shell:python step resolves through it unchanged. The pre-commit cache is re-keyed on the composite's reported version since env.pythonLocation came from setup-python.

This drops one action and its post steps per job, removes the dependence on the runner tool cache (a fresh 3.13.x patch before the image update costs setup-python a 10-25s download), and runs the scripts on the faster astral interpreter. The CodSpeed benchmarks job intentionally stays on actions/setup-python so callgrind instruction counts remain comparable. The composite's setup-uv pin is bumped to the v8.3.2 SHA the deleted standalone call sites were already on.

Verified on this PR's CI plus a check-board-images dispatch from the branch; the two catalog sync workflows and regen-board-catalog are the same pattern reviewed by eye, and the release path gets exercised at the next release.

- follow up to #2226

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/.github/actions/generate-release-notes/action.yml
+++ b/.github/actions/generate-release-notes/action.yml
@@ -59,16 +59,13 @@ outputs:
 runs:
   using: composite
   steps:
-    - name: Set up Python
+    - name: Set up uv and Python 3.13
       if: inputs.initial-release != 'true'
-      uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
+      # Resolves through the caller's workspace checkout — both call
+      # sites check the repo out at the workspace root.
+      uses: ./.github/actions/setup-uv-python
       with:
         python-version: "3.13"
-
-    - name: Install Python dependencies
-      if: inputs.initial-release != 'true'
-      shell: bash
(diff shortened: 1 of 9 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot confirmed on the lint run for faa2cc52: the restore key rendered as pre-commit|py3.13|eb1d9a37..., so the composite output populated; one expected miss on the new key shape, then saved for subsequent runs.

PR #2235, opened 2026-07-20
Enable ccache for the native ESP-IDF e2e job

bdraco:

Speeds up the E2E native ESP-IDF job with ccache, the device-builder side of what esphome/esphome#17722, #17728, and #17735 do for host and platformio builds. Those PRs target other toolchains; for native IDF the support already landed on esphome dev: espidf/framework.py enables ccache whenever the binary is on PATH, with CCACHE_BASEDIR rewriting the per-run build paths so repeat compiles hit across runs, and it stores the cache under the IDF tools path.

That tools path lives inside ESPHOME_ESP_IDF_PREFIX, which this job already persists with a rolling run_id cache key, so the objects ride the existing toolchain cache with no new cache entry. The only missing ingredient was the ccache binary (not preinstalled on ubuntu runners, so esphome silently disabled itself); the job now apt installs it and prints ccache -s after the run, mirroring esphome's own CI, so hit rates are visible in the log. First run after merge seeds the cache; runs after that should skip most of the ninja compile.

- companion to esphome/esphome#17735

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -419,9 +419,7 @@ jobs:
         # installed IDF version is pinned by ``esphome@dev`` (reinstalled
         # each run), not by the test file -- so use a rolling ``run_id`` key
         # that saves a fresh cache every run (picking up dev IDF bumps) and
-        # ``restore-keys`` to warm from the most recent prior cache. The
-        # ccache dir lives under the same prefix, so compile objects ride
-        # this cache across runs too.
+        # ``restore-keys`` to warm from the most recent prior cache.
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
         with:
           path: ~/.cache/esphome-idf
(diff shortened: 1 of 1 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

Reworked: the ccache objects now live in their own explicitly restored cache entry (CCACHE_DIR moved outside the IDF prefix; rolling run_id key with restore-keys, same shape as the toolchain cache) instead of implicitly riding the multi-GB toolchain save.

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

Head is now c7b983b5. The stats-step fragility from the prior review is gone by construction: CCACHE_DIR is set explicitly by the workflow to ~/.cache/esphome-ccache (a dedicated cache entry), so the stats step reads our own path rather than re-deriving esphome's internal layout.

esphbot (approved):
> [!TIP]

PR #2273, opened 2026-07-22
Type int-keyed catalog enums as integer via live introspection

### LINKED ISSUE #2272: Catalog sync types all-integer enums as string, causing quoted YAML churn on select picks
While auditing esphome/device-builder-frontend#1372, a sweep of the generated component catalog found ~76 config vars whose `options` values are all integer literals but whose `type` is `string`; examples: `sensor.ina226` `adc_averaging` (1/4/16/64/128), `sensor.ads1115` oversampling values, `sensor.mlx90393`, the `wk21xx` UART channel fields, `bk72xx`/`ln882x` fields. Only four such fields are typed `integer` today (`uart.baud_rate`, `logger.baud_rate`, `speaker.resampler.bits_per_sample`, `esp32_camera.jpeg_quality`).

The frontend coerces select emits by the entry's declared type (esphome/device-builder-frontend#1372), so these string typed fields keep emitting strings; the serializer must then quote them (`oversampling: "16"`) and an authored bare `16` re serializes quoted after any pick. ESPHome accepts both forms (`cv.enum(..., int=True)` coerces), so it is round trip churn rather than breakage, but it is the same churn class the recent frontend sweep closed.

Fix belongs in `script/sync_components.py` type refinement: when a config var's enum values are all integer literals, emit `type: integer` (keeping the options list), the same way the four already typed fields ended up integer. The frontend then coerces picks on those fields with no further change.

bdraco:

The catalog typed ~60 all integer enums as `string` (ina226 `adc_averaging`, the wk21xx UART channel fields, `web_server.version`, ...), so every select pick re serialized the value quoted (`adc_averaging: "16"`) even when the user had authored a bare `16`. The issue proposed a static rule (all integer looking options become `integer`), but that turns out to be unsafe: cc1101's `wait_time`, max7219digit's `rotate_chip`, ads1115's `sample_rate`, atm90e32's `current_phases` and fs3000's `model` are string keyed enums upstream and reject a bare int outright, so the static rule would have broken them.

Instead the sync now reads the live enum's key types through the existing refined types introspection: a `cv.one_of` / `cv.enum` whose values are all real ints refines `string` to `integer` (options kept), while string keyed enums are left alone. To reach the wk21xx fields the schema walker gains an opt in `descend_list_items` mode that follows `cv.ensure_list` item schemas.

The walker's revisit dedupe also changes from bare `id` to `(id, path)`, and that part is deliberately unconditional, so it affects every `_walk_schema_keys` collector (refined types, field ranges, platform defaults and constraints), not just the enum path. Previously a sub schema shared by two fields was only visited at the first path, so collector output on the later siblings was silently shadowed; in HEAD, one haier binary_sensor's `web_server.sorting_weight` is `float` while its five identical siblings are `string`. The regenerated catalog therefore also picks up those corrections: 60 string to integer flips, ~1460 string to float/lambda/mac_address fixes and 123 recovered `range` bounds on previously shadowed fields, and 4 recovered `unit_options`. The three `remote_receiver` protocol fields stay string (registry built schemas the walker cannot see); their validators accept both forms, so nothing breaks there.

On the frontend side no change is needed: all select emit paths coerce picks by the entry's declared type since device-builder-frontend#1375 (fixing device-builder-frontend#1372), with a test pinning exactly this shape, an INTEGER entry whose option values are strings emits a number. That fix is on frontend main but not yet in the released bundle (0.1.268 predates it), so the user visible unquoting completes with the next routine frontend bump; until then a pick keeps serializing quoted exactly as today, which every affected validator accepts.

- fixes #2272

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

PR #2283, opened 2026-07-23
Run the compile subprocess tree in UTF-8 mode

bdraco:

On a native-Windows build server, the `esp_idf_size` memory table in the compile log renders as mojibake:

```
Γöé Memory Type/Section Γöé Used [bytes] Γöé Used [%] Γöé ...
```

Byte-level analysis of a downloaded job log shows the stored text is valid UTF-8 of `Γöé`, and re-encoding it as CP437 recovers `│` (`e2 94 82`) — i.e. some layer of the Windows toolchain subprocess tree decoded UTF-8 box-drawing output with the OEM codepage and re-emitted it. The dashboard's own pipeline is faithful (it decodes UTF-8 with `errors="replace"`), as are the layers that could be audited from source: esphome's espidf runner, idf.py's stream reads, and `run_size_tool.cmake` (which doesn't capture output at all). The mangling layer is one that consults the platform-default/console encoding.

Fix at the environment level rather than chasing the individual layer: add `PYTHONUTF8=1` and `PYTHONIOENCODING=utf-8` to `ESPHOME_SUBPROCESS_ENV`, the shared constant both the local job runner (`compose_subprocess_env`) and the remote receiver runner already layer onto every `esphome` subprocess. This puts every Python in the build tree (esphome, the espidf runner, idf.py, the IDF penv tools including esp_idf_size) into UTF-8 mode for stdio and for locale-default codecs, so no layer can fall back to the OEM codepage in either direction. On POSIX the defaults are already UTF-8, so this only pins the existing behavior and keeps the local/remote parity contract.

- fixes the `Γöé` mojibake in the esp_idf_size table of Windows compile logs

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/firmware/constants.py
+++ b/esphome_device_builder/controllers/firmware/constants.py
@@ -117,11 +117,19 @@ _NO_ESPHOME_MODULE_MARKER = "No module named 'esphome'"
 # * ``PYTHONUNBUFFERED`` keeps Python subprocesses flushing progress
 #   lines (especially ``\r``-terminated ones) instead of buffering
 #   them until a ``\n`` arrives.
+# * ``PYTHONUTF8`` / ``PYTHONIOENCODING`` force every Python in the
+#   build tree (esphome, idf.py, the IDF penv tools) onto UTF-8 for
+#   stdio and locale-default codecs. Without them a layer of the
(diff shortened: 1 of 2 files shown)


esphbot (approved):
> [!TIP]

PR #2285, opened 2026-07-23
Purge the per-dashboard ESP-IDF tree instead of migrating it

bdraco:

Follow-up to #2281 after field testing on a real Windows build server. 1.6.10b2 *moves* an existing per-dashboard `C:\esphb\<id8>\idf` into the shared `C:\esphb\idf`, and that breaks builds: idf.py refuses with

```
'C:\esphb\idf\penvs\5.5.5\Scripts\python.EXE' is currently active in the environment while the
project was configured with 'C:\esphb\GOEjEGFv\idf\penvs\5.5.5\Scripts\python.EXE'.
Run 'idf.py fullclean' to start again.
```

The IDF tree bakes absolute penv paths into every configured project's build dir, so a moved tree fails idf.py's active-vs-configured python consistency check. Migrating the toolchain was the wrong strategy.

This PR replaces migration with purge-and-reinstall:

- The shared `C:\esphb\idf` always starts empty (`_relocate_into` with no sources: mkdir + completion marker) and esphome installs the toolchain fresh into it.
- An earlier release's per-dashboard `<root>\idf` is **deleted** (esphome's `rmtree`, clears read-only attrs); a failed purge is logged and leaves dead disk, never blocks the shared dir.
- esphome's machine-global IDF cache is no longer swept in either — moving it has the same baked-path problem and steals the cache out from under parallel native CLI use. `_default_idf_cache` remains only to name the fallback location in the not-relocated warning.
- The "intact per-dashboard install serves this session" fallback rung is gone (that tree is now the purge target and is path-broken for existing projects anyway); on shared-dir failure the existing leave-unset-and-warn path stands.
- No cleanup for a `C:\esphb\idf` already populated by b2's move: b2 was live for well under an hour before being superseded, and "Reset build env" already recovers any affected project.

Known consequence: a build dir configured before the prefix change still records the old penv path, so its first compile fails with idf.py's fullclean message. The existing "Reset build env" action recovers it, so no dedicated handling is added.

- fixes the idf.py active-vs-configured python failure introduced by the #2281 migration path (1.6.10b2)

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

PR #2359, opened 2026-07-28
Accept percent suffix on percentage_int fields in the catalog

### LINKED ISSUE #2358: [Bug] Percent sign in the `carrier_duty_percent` value from `remote_transmitter` triggers a warning
### What happened?

The `carrier_duty_percent` value from `remote_transmitter` seems to be a special case. According to the documentation, it is an `int` value, but a percentage value (e.g., 50%) is accepted, as the examples also show. However, the structured editor displays a warning saying "Must be a number" when a percent sign is appended.

Perhaps the implementation in ESPHome is a bit awkward? For example, `light.brightness` is also specified as a percentage, but without a percent sign.

[remote_transmitter docs](https://esphome.io/components/remote_transmitter/)
```
carrier_duty_percent (Required, int): How much of the time the remote is on. For example, infrared protocols modulate the signal using a carrier signal. Set this to 50% if you’re using IR LEDs and 100% for RF applications like 433 MHz transmitters.
```

[light docs](https://esphome.io/components/light/)
```
brightness (Optional, percentage, [templatable](https://esphome.io/automations/templates)): The primary brightness of the light; applies to all channels (both color and white) of the light.
```

[YAML](https://github.com/esphome/rf-proxies/blob/main/m5stack-atom-echo/m5stack-atom-echo.yaml)

Image

### Dashboard version

1.7.0

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The editor rejected `50%` in `remote_transmitter.carrier_duty_percent` with "Must be a number"; the schema bundle carries no type for `cv.percentage_int` fields, so the catalog typed it as a plain integer even though the field description says to use `50%`.

Register `cv.percentage_int` in `_NON_INTROSPECTABLE_UNITS`, the existing table for hand-rolled suffix validators with no introspectable regex; the live introspection pass then derives every affected field from the upstream validator, no field names are hardcoded. Those fields now ship as `float_with_unit` with a `%` unit, which the frontend already renders and serializes, and `range` is preserved. The regenerated catalog updates `remote_transmitter.carrier_duty_percent` plus the climate `visual.min_humidity`/`visual.max_humidity` fields. `cv.percentage` is deliberately not registered, its `%` suffix rescales the value; `remote_receiver` is not reached because its schema is wrapped in a `schema_extractor` closure the walker cannot peel, a pre-existing limitation tracked in #2360.

- fixes https://github.com/esphome/device-builder/issues/2358

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

PR #2367, opened 2026-07-28
Drop unit refinement when a float_with_unit regex has no unit group

### LINKED ISSUE #2361: Unitless float_with_unit validators leak regex fragments into unit_options
Found during review of #2359.

`_extract_validator_units` in `script/sync_components.py` pulls the unit list out of a `cv.float_with_unit` closure by matching the last parenthesized group of its regex anchored to `$`. When the validator is built with an empty unit string, for example tsl2591's `cv.float_with_unit("device_factor", "", True)`, the trailing unit group is empty and the search falls through to the mantissa's `(\w*?)` group, so the regex fragment itself becomes the unit list.

The shipped catalog reflects it; `sensor.tsl2591.device_factor` and `sensor.tsl2591.glass_attenuation_factor` are `float_with_unit` with `unit_options: ["\\w*?", "n\\w*?", "µ\\w*?", "m\\w*?", "k\\w*?", "M\\w*?", "G\\w*?"]`. Each option is appended verbatim to the numeric value on save, so picking one emits an unusable YAML value.

Fix direction: skip the refinement when the matched alternation is the bare mantissa fragment (`\w*?`), leaving unitless validators as plain `float`; then regenerate.

bdraco:

A `cv.float_with_unit` validator built with an empty unit string has no unit group in its regex, so `_extract_validator_units`'s group search landed on the mantissa's `(\w*?)` fragment and the catalog shipped regex fragments as unit options; `sensor.tsl2591.device_factor` and `glass_attenuation_factor` carried `unit_options: ["\w*?", "n\w*?", ...]`, and picking one appends the fragment verbatim to the value on save.

Filter the extracted alternatives through the existing `_SUFFIX_UNIT_RE` unit shape gate; when nothing unit shaped remains the refinement is dropped and the field stays a plain float. All real unit lists are unchanged, verified against every closure introspected validator; the regenerated catalog only touches the two tsl2591 fields.

- fixes https://github.com/esphome/device-builder/issues/2361

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

Issue #2370, opened 2026-07-28
Automations build runs no type refinement, unit fields ship picker-less


bdraco:
Split from #2364, the last remaining item there.

The automations catalog build (`build_automations` in `script/sync_components.py`) converts action, condition, and trigger fields from the schema bundle but never runs the live type refinement pass the component build gets (`_collect_refined_types` and friends), so unit coerced fields in action schemas ship without their pickers. Example: the `http_request` action's `max_response_buffer_size` is `cv.validate_bytes` upstream but lands as a plain field, while the same validator on component fields now renders the B/kB/MB/GB picker.

Fix direction: run the refined types collection against the live action or trigger schema during the automations build, reusing the component side machinery; the wrapper peeling from #2365 and its follow up applies there unchanged. Scope it to refinement lookup by path, the automations bundle shapes are otherwise already correct.

PR #2371, opened 2026-07-28
Peel wrapper validators at any depth in the schema walk

### LINKED ISSUE #2364: Trigger scanners and typed defaults do not peel schema_extractor wrappers
Follow-up from the #2360 fix; findings from its review.

The `schema_extractor` peel (`_hidden_schema` in `script/sync_components.py`) is applied where the dict view of a `CONFIG_SCHEMA` is consumed: `_walk_schema_keys` (all its collectors) and `_collect_required_groups`. Two consumers still see the raw closure and stay inconsistent for wrapper-topped components:

1. `_scan_schema_for_trigger_singles` / `_scan_schema_for_list_triggers` unwrap the module `CONFIG_SCHEMA` directly, and `remote_receiver`'s is exactly the `validate_triggers` closure; `on_*` keys living inside the hidden schema are invisible to the scan. Today this is masked by the module-globals sweep plus the `TRIGGER_REGISTRY` reach-through, so nothing is known to be missing, but the coverage is compensated rather than complete.
2. `_collect_typed_defaults` passes the raw schema to `_typed_default_of`; a closure-wrapped typed schema would yield no defaults. No component currently hits this.

Note for whoever picks this up: do not peel *nested* values, and memoize `_hidden_schema` first if that is ever attempted; such closures build fresh schema objects per call, which defeats the walker's `(id, path)` dedupe and made a nested peel grow to 9 GB on lvgl during development of the #2360 fix. `_registry_from_schema` must keep receiving the raw closure, it digs `__closure__` cells, so a fetch-level "effective schema" accessor is the wrong shape.

bdraco:

Finish the wrapper peeling that #2365 scoped to the top level. The peel now lives in `_unwrap_schema_to_dict` itself, so every consumer (the walker collectors, the trigger scanners, required groups, typed defaults) sees through wrapper validators at any nesting depth, with the two traps that forced the earlier scope limit defused: `_hidden_schema` is memoized so repeated probes of one closure return one object and the walker's id based dedupe holds (the earlier nested attempt built fresh schemas per probe and grew unbounded on lvgl), and only a `vol.Schema` probe result is accepted, so a non schema extract like `cv.enum`'s bare value mapping is never walked as config keys. A second peel handles plain delegating wrappers with no sentinel branch (`maybe_empty_debug` returns `DEBUG_SCHEMA(value)`); a wrapper referencing exactly one module level schema peels to it.

The regenerated catalog changes exactly two fields: `uart.debug.after.bytes` and `ble_nus.debug.after.bytes` gain the byte unit picker they were validated for upstream. A full regen was verified twice at normal runtime and memory, lvgl included. Also carries a small comment correction to `check_catalog.py` that missed the #2368 merge.

The `http_request` action's `max_response_buffer_size` remains picker less; that is the automations build path, which does not run type refinement at all, tracked in https://github.com/esphome/device-builder/issues/2370.

- fixes https://github.com/esphome/device-builder/issues/2364

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

Issue #2373, opened 2026-07-28
Represent lambda-or-plain union fields with the templatable toggle


bdraco:
Split from the #2370 work.

Upstream fields validated as `cv.Any(cv.returning_lambda, <plain form>)` (the hand rolled cousin of `cv.templatable`) accept both a plain value and a lambda; the datetime set actions' `date` / `time` / `datetime` fields are the known cases. The refinement pass now deliberately refuses to type such a union `lambda` (the plain form is the primary YAML shape, a lambda editor would hide it), so they stay plain typed with no lambda affordance.

The faithful representation already exists for `cv.templatable` fields: the bundle stamps `templatable: true` and the frontend renders a lambda toggle beside the plain input. Enhancement: detect the `Any(returning_lambda, X)` union in the refinement pass and stamp the same `templatable` flag (and X's refined type when classifiable) instead of dropping the lambda half entirely. Needs a small applier change since `_apply_refined_types` only sets `type` and `unit_options` today.

PR #2374, opened 2026-07-28
Run type refinement against the live automation registries

### LINKED ISSUE #2370: Automations build runs no type refinement, unit fields ship picker-less

bdraco:

The automations build converted action and condition fields from the schema bundle alone, so unit coerced and lambda validated fields shipped untyped; the `http_request` actions' `max_response_buffer_size` is `cv.validate_bytes` upstream but rendered as plain text with no byte picker.

Run the component path's refined types collection against the live action and condition registries: `_collect_automation_refined_types` mirrors the existing `_collect_automation_registry_groups` (same registries, same fill timing after the import sweep), a `maybe_simple_value` registration peels to its validator half, and the collected refinements apply through the existing `_apply_refined_types` keyed by qualified id. One guard was needed along the way: a lambda branch inside a multi branch `cv.Any` no longer claims the field, since a `cv.Any(returning_lambda, date_time)` union's primary YAML shape is the plain value and a lambda editor would hide it; representing those unions with the templatable toggle instead is tracked in #2373.

The regenerated catalog touches nine automation bodies, all verified against their upstream validators: the three `http_request` actions gain the B/kB/MB/GB picker, `api.respond.data`, `mqtt.publish_json.payload`, and `select.is.lambda` become `lambda` (bare `cv.lambda_` upstream), `espnow.broadcast.address` becomes `mac_address`, `mhz19.detection_range_set.detection_range` becomes `integer`, and `for.setup_priority` becomes `float`. Component bodies are untouched.

- fixes https://github.com/esphome/device-builder/issues/2370

## Types of changes

## Frontend coordination

## Checklist

A `--limit-component` debug run now skips the automations build and emit with a warning instead of rewriting the committed automations catalog from half filled registries.


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2385, opened 2026-07-28
Refine hex_int fields to integers with hex display

### LINKED ISSUE #2380: hex_int_range fields ship as plain strings with no hex handling
Found while auditing #2378's range removals.

The 26 micronova `memory_location` / `memory_address` / `memory_data*` fields are `cv.hex_int_range(min=0x00, max=0x79/0xFF)` upstream, genuinely integer valued hex fields, but they ship as `string` with no hex handling: the custom composed validator escapes both the bundle's `data_type` path (which only fires for module level assignments like `hex_uint8_t`) and the refined type pass. Before #2378 they carried inert ranges on string inputs; after it they are plain text fields.

Fix direction: classify `cv.hex_int_range` (and bare `cv.hex_int`) in `_collect_refined_types`, emitting `integer` with `display_format: "hex"` to match the `_DATA_TYPE_HEX` handling `i2c_address` gets; the `vol.Range` in the chain is already collected by the ranges pass, so the bounds return automatically through the #2378 numeric gate once the fields are integer typed. `_apply_refined_types` applies over `string` typed entries, so no applier change is needed.

bdraco:

Fields validated with `cv.hex_int_range` shipped as plain strings with no hex handling and no bounds; the custom composed chain (`All(hex_int, Range(...))`) escapes both the bundle's `data_type` path, which only fires for module level assignments like `hex_uint8_t`, and the refined type pass. The 26 micronova memory fields were the visible case, and since #2378 they trip the dropped bounds warning on every sync.

Register `cv.hex_int` in the refined type pass; `RefinedType` gains a `display_format` the applier stamps alongside the type, so a field the bundle emits untyped whose chain contains `hex_int` becomes an `integer` with `display_format: "hex"`, the same rendering `i2c_address` already gets. The stamp is suppressed on options backed entries, whose select values arrive from the bundle in decimal and would desync from a hex formatted value; hex fields the bundle already types integer keep decimal display for now, widening needs the refined type analog of the range bleed guard and is tracked in https://github.com/esphome/device-builder/issues/2387. Bounds return automatically through #2378's numeric gate, and the dropped bounds warnings for these fields stop firing. Derived from the validator identity, no field names hardcoded.

Two fail loud hardenings ride along, mirroring the units table: a registered validator name missing from the installed esphome now raises instead of silently skipping (the guard's first run immediately caught `negative_float`, registered but nonexistent upstream, now dropped), and the refinement pass no longer swallows a broken esphome import into an empty map.

The regenerated catalog changes 7 bodies, each field verified against its upstream validator: the 26 micronova memory fields become hex integers with their `[0, 121]` and `[0, 255]` bounds restored, and ethernet's `phy_registers` trio and openthread's `pan_id`, `ext_pan_id`, `network_key`, and `pskc` become hex integers. sx127x's options backed `preamble_polarity` ships byte identical, the suppression keeps the hex stamp off enum entries entirely. Note the openthread `network_key`/`pskc` are 128 bit values; the frontend's hex parser BigInt routes wide integers, and values above 2^53 round trip as `0x` strings by design.

- fixes https://github.com/esphome/device-builder/issues/2380

## Types of changes

## Frontend coordination

## Checklist


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Hex stamp lands on an options-backed enum and desyncs the frontend select (sx127x `preamble_polarity`)

esphbot (approved):
> [!TIP]

PR #2404, opened 2026-07-28
Collect action and condition ranges from the live registries

### LINKED ISSUE #2375: Collect action and condition ranges from the live registries
Follow-up from the #2374 review passes.

Action and condition schemas carry `cv.int_range` style bounds that the automations build still drops; ranges are the plausible third live registry collector after required groups and refined types. Before adding it, do the small prep refactor the second collector exposed: a shared `_iter_automation_registry_entries()` yielding `(registry_type, registry_id, entry)` so each collector is a comprehension over one iterator, and a schema accepting core for `_collect_refined_types` (and `_collect_field_ranges`) so the automations path passes `_registry_entry_schema(entry)` directly instead of the `types.SimpleNamespace(config_schema=...)` shim.

Frontend already enforces `range` on numeric entry types, so collected bounds become live validation immediately; the applier side is one more lookup keyed by qualified id, same as `_apply_automation_refined_types`.

bdraco:

The automations build collected refined types from the live registries but not `vol.Range` bounds, so bounded action fields shipped without validation.

First the prep refactor this issue scoped: a shared `_iter_automation_registry_entries()` both collectors iterate, and schema accepting cores split out of `_collect_refined_types` and `_collect_field_ranges` (`_refined_types_in_schema`, `_field_ranges_in_schema`) so the automations path passes `_registry_entry_schema(entry)` directly, deleting the `SimpleNamespace` manifest shim two prior reviews flagged. On that base, `_collect_automation_field_ranges` gathers per entry bounds through the existing range walker (All chain intersection, partial and disjoint handling included), and `_apply_automation_field_ranges` overlays them keyed by qualified id, ordered after type refinement and routed through the numeric gated `_apply_field_ranges`, so bounds land only on inputs that render them.

The regenerated catalog gains one range: `canbus.send.can_id` gets `[0, 536870911]`, matching upstream's `cv.int_range(min=0, max=0x1FFFFFFF)`. The yield is deliberately honest; most action fields wrap in `cv.templatable`, whose closure hides the validator chain from the walker, so their bounds stay uncollected until the dual form work in #2373 peels it. The mechanism is in place for both that and future upstream schemas.

- fixes https://github.com/esphome/device-builder/issues/2375

## Types of changes

## Frontend coordination

## Checklist


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Regenerated catalog looks incomplete — `lvgl.spinner.update.arc_length` should also gain a range

esphbot (approved):
> [!TIP]

PR #2416, opened 2026-07-29
Mark lambda-or-plain union fields templatable

### LINKED ISSUE #2373: Represent lambda-or-plain union fields with the templatable toggle

bdraco:

Hand rolled lambda or plain unions (`cv.Any(cv.returning_lambda, X)`) had their lambda half dropped entirely: the classifier suppressed the lambda branch so it could not mistype the field, but nothing represented the lambda affordance, so the datetime set actions offered no toggle for a value upstream happily accepts as `!lambda`.

The suppression upgrades to representation: `classify_branches` now inspects every union branch (a lambda may sit after the branch that types the field), takes the first non lambda classification as the type, and marks the refinement `templatable` when a lambda branch coexists; a union whose plain branch does not classify yields a stamp only refinement, so the bundle's typing stands and only the flag lands. The applier stamps `templatable: true` additively, never clearing a bundle set flag. This produces exactly the shape `cv.templatable` fields already carry, which the frontend's toggle renderer paints and device-builder-frontend#1529 validates.

The regenerated catalog changes five automation bodies, all pure `templatable: true` stamps with typing untouched, each verified upstream: the three `datetime.{date,datetime,time}.set` value fields (`cv.Any(cv.returning_lambda, cv.date_time(...))`, this issue's target), and `http_request.{post,send}.json` (`cv.Any(cv.lambda_, cv.Schema({...}))`), where the flag is honest metadata that stays inert in the UI since the toggle opts structural entries out, and the lambda half remains YAML editable as before.

- fixes https://github.com/esphome/device-builder/issues/2373

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1529 (merged; the lambda validation skip these toggles rely on)

## Checklist


esphbot (approved):
> [!TIP]

PR #2418, opened 2026-07-29
Type multi-value entries from their ensure_list item validators

### LINKED ISSUE #2413: templatable(ensure_list) fields are invisible to the walker
Found during the #2411 review passes; pre-existing class, not a regression.

`cv.templatable(cv.ensure_list(...))` fields are invisible to the walker's closure branch: it tries `_typed_branch_schemas` and `_ensure_list_item_validator` on an unresolvable node but not `_templatable_inner`, so the list item validator is never reached. Two upstream occurrences, both `cv.templatable(cv.ensure_list(cv.hex_uint8_t))`: `speaker/__init__.py:77` and `ble_client/__init__.py:142`; the payoff is hex display on the two byte array fields.

Fix is one line in the walker's closure fallback (peel `_templatable_inner` before trying the list item extractor), plus a regen and a pin on one of the two fields.

bdraco:

Scalar `cv.ensure_list` items were invisible to type refinement: the walker's visit fires on the parent field, and a scalar item validator is never itself walked, so a list field whose bundle node carries no `data_type` shipped untyped. `classify` now peels an `ensure_list` closure to its item validator and types the multi value entry from the item's classification; the walker's list descent additionally peels a `cv.templatable` wrapper first so mapping shaped items behind it stay reachable.

Honest scoping against the issue: the two fields it cites (`speaker.play.data`, `ble_client.ble_write.value`) turn out to have been covered all along, their bundle nodes carry `data_type: hex_uint8_t` and the static path already ships them as hex integers with the toggle; the premise held only for list fields the bundle leaves untyped. The regenerated catalog changes exactly one such field: `espnow.peers` (`cv.ensure_list(cv.mac_address)` upstream) becomes a `mac_address` multi value entry. The live registry path is pinned anyway (`speaker.play.data` refines identically through the peel), so the static coverage is no longer the only line of defense.

- fixes https://github.com/esphome/device-builder/issues/2413

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2434, opened 2026-07-29
Promote nested var-less trigger fields to TRIGGER

### LINKED ISSUE #2392: Nested action lists cannot route to the automation editor
Found while fixing #2379.

The frontend's TRIGGER field renderer emits `edit-action-field` with only the leaf key (`_emitEditActionField(entry.key)` in `config-entry-form.ts`), so an action list nested below a component's top level cannot route to the automation editor; the sync therefore ships nested var-less trigger fields (sprinkler's `multiplier_number.set_action`, `repeat_number.set_action`, `valves.run_duration_number.set_action`) as `unknown`, editable only through the YAML pane.

Fix direction is frontend side (esphome/device-builder-frontend): carry the full dotted path in the event (the renderer already computes `fieldKeyAttr(path)`) and resolve the nested action list in the section host and automation editor. Once routing works, the sync's `top_level` scoping on the TRIGGER promotion can be lifted so these fields get the edit button, and the `unknown` fallback retires.

bdraco:

Third of three parts for #2392, the one that closes it. With the backend addressing nested action fields by dotted path (#2433) and the frontend carrying the full field path in the edit event (esphome/device-builder-frontend#1539), the sync's `top_level` scoping on the `type: trigger` promotion has no reason left: a nested var less trigger field can now route, so it gets the TRIGGER type and the edit button instead of the YAML only `unknown` fallback.

The gate lifts to an unconditional promotion (still scoped to no inner config_vars; a trigger with params keeps its nested mapping), and the now dead `top_level` parameter and its call site plumbing go away. The regenerated catalog changes exactly one file: sprinkler's three nested `set_action` entries (`multiplier_number`, `repeat_number`, and the per valve `run_duration_number`) flip `unknown` to `trigger`, verified as the only semantic diff across the whole catalog. The four tests pinning the old scoping now pin the lifted behaviour.

- fixes https://github.com/esphome/device-builder/issues/2392

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1539 (should merge first, with #2433)

## Checklist


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Promotion lands ahead of the routing it depends on — gate the merge on #2433 + the frontend bump

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Promotion still lands ahead of its routing — both companions remain open

Issue #2396, opened 2026-07-28
[Bug] Structured editor doesn't recognoze the api.services keyword


breti:
### What happened?

Actions were previously called Services. The structured editor doesn't recognoze the services keyword:

[Docs](https://esphome.io/components/api/)

Configuration:

```
api:
  services:
    - service: myservice
      then:
        - logger.log: "Hello World!"
```

<img width="962" height="498" alt="Image" src="https://github.com/user-attachments/assets/3f0fbfad-6f00-4d12-b4ae-556cdfd74e14" />

### Dashboard version

1.7.1b1

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

PR #2439, opened 2026-07-29
Locate flush-style api action lists

bdraco:

A flush style api action list, dashes at the same column as the block key, is valid YAML that esphome compiles:

```yaml
api:
  actions:
  - action: start_va
    then:
      - logger.log: go
```

`locate_actions_list` ended the block at the first line whose indent was at or below the api child indent, which is the first flush dash itself, so the list read as empty. An upsert of an existing name then appended a duplicate four indent item and orphaned the flush items into bare list entries directly under `api:` (invalid YAML), and a delete of an existing item raised NOT_FOUND. Parsing was unaffected, so the manage list showed actions the editor could not edit or delete.

The same indent comparison lived in `helpers/yaml/inline.py:_block_end`, where the #2433 review found two more members of the class: a flush style leaf list under a component action field corrupted on flat upsert and delete, and a flush style container broke the sub entity descent. All the splice paths now share one boundary helper, `child_block_end` in `helpers/yaml/scan.py`, mirroring the frontend's `endsBlockAtIndent` rule: a flush dash line (including the lone dash spelling) continues the block, blank and comment lines never end it (a comment at the key's indent between flush items used to hide everything after it), and the end trims back over trailing blanks and shallow comment banners while a comment indented inside the last item stays with it. `locate_actions_list`, the inline handler upsert and remove, the sub entity locator, and the singleton handler splices all use it; `_item_spans` also accepts lone dash items so they round trip. A one space `api:` child indent, which the two space item emit cannot nest under, is refused with INVALID_ARGS instead of writing unparseable YAML.

Surfaced by the Kōan reviews on #2420 and #2433 as a pre existing gap; reproduced on main at e4e0959a. Top level lists (`script:` / `interval:`) and the frontend already handled flush style.

- related to #2396

## Types of changes

## Frontend coordination

- [ ] Companion frontend PR:

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/automations/api_actions.py
+++ b/esphome_device_builder/controllers/automations/api_actions.py
@@ -16,9 +16,7 @@ from __future__ import annotations
 
 import re
 
-from ...helpers.api import CommandError
 from ...helpers.yaml.scan import child_block_end, is_list_item_line
-from ...models.api import ErrorCode
 from ...models.automations import YamlDiff
 
 #: Accepted spellings, canonical first — esphome's ``cv.rename_key``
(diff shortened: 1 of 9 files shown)


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `_trim_trailing_gap` is indent-agnostic — it orphans comments that belong to the last item

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `_block_end` deletion misses the two call sites `origin/main` added in #2433 — rebase must convert them

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2440, opened 2026-07-29
Unify the key-block locators in helpers/yaml/inline.py

### LINKED ISSUE #2436: Unify the key-block locators in helpers/yaml/inline.py
From the #2433 simplify review.

The module now has three spellings of find a key block inside a span: _locate_key_block (nested descent, with dash line first key handling), _locate_subentity_instance's header search (no dash form), and _instance_declared_id's dash line id match. #2433 extracted the shared block tail (_key_block_frame), but the header search loops remain separate, and the sub entity path lacks the dash line first key handling the nested path gained, so a sub entity block spelled as its item's first key is invisible to upsert_subentity_handler. Expressing _locate_subentity_instance via _locate_key_block (or folding the dash form into key_header_re as an option in helpers/yaml/scan.py) would leave one locator; the behaviour delta for the sub entity path is strictly additive but needs its own tests.

bdraco:

Follow up to the #2433 simplify review, stacked on that branch. `_locate_subentity_instance` now delegates to `_locate_key_block` instead of keeping its own header search loop, and `_instance_declared_id` builds its dash line match from a shared prefix in `scan.py`, so the key block spelling lives in one place. The sub entity path picks up the dash line first key handling the nested path already had: a sub entity block spelled as its item's first key is now visible to the writer, and an inline scalar sub key refuses with a clear error instead of a misleading not found. Both deltas get their own writer and branch tests; the existing suite passes untouched.

- fixes https://github.com/esphome/device-builder/issues/2436

## Types of changes

## Frontend coordination

## Checklist


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Indexed sub-entity upsert still leaks `YamlUpsertNotSupportedError` uncaught

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2441, opened 2026-07-29
Collapse resolve_action_field_target into the shared component target walk

### LINKED ISSUE #2435: Collapse resolve_action_field_target into the shared component target walk
From the #2433 simplify review.

resolve_action_field_target is a third parallel resolver beside resolve_component_target and resolve_component_domain: it re-walks _iter_instance_targets because ComponentTarget does not carry the instance's platform, which the action field write path needs for catalog_id. The deeper cut is to stamp the catalog id (or parent platform) onto ComponentTarget where _iter_instance_targets builds the targets, then express resolve_action_field_target as a thin wrapper the way resolve_component_domain already is; one matching semantics instead of two subtly different walks. A ComponentTarget carried answer would also make it impossible for a sub entity id to resolve in the resolver and then miss in the splice locator.

bdraco:

Follow up to the #2433 simplify review. `_iter_instance_targets` now stamps the instance's catalog id onto `ComponentTarget` as it builds each target, with a sub entity carrying its parent's, so `resolve_action_field_target` collapses to a thin wrapper over `resolve_component_target` the way `resolve_component_domain` already is; one matching semantics instead of two subtly different walks, and a sub entity id can no longer resolve in one resolver while missing in the other. Return shape and the write path call site are unchanged; a new parse test pins the stamped catalog id for the singleton, platform item, and sub entity shapes.

- fixes https://github.com/esphome/device-builder/issues/2435

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2442, opened 2026-07-29
Keep the adoption YAML when only the remote package fails to resolve

### LINKED ISSUE #2424: Stale broadcast package_import_url permanently blocks adoption; refusal copy points at a nonexistent editor
## The failure

Adopting a GL-S10 running its factory firmware fails:

> Can't import — config doesn't validate: gl-s10.yaml does not exist in repository. Fix the errors in the editor and try again.

The device broadcasts `package_import_url = github://esphome/bluetooth-proxies/gl-s10.yaml@main`, but the upstream repo was restructured and the file now lives at `gl-inet/gl-s10.yaml`. The broadcast URL is baked into the shipped firmware, so every device on the older firmware carries a permanently dead pointer, and the user cannot influence it.

## Why the current behavior strands the device

`import_device` writes the adoption YAML, runs the pre-write validation (`_validate_rewritten_yaml_or_raise`, action="import"), and on failure deletes the file. For a dead package URL that means:

1. Adoption can never complete. The only way forward is hand-writing a config from scratch.
2. The error copy tells the user to "Fix the errors in the editor and try again", but nothing was written, so there is no editor to fix anything in.

The design principle in CLAUDE.md validates *our* generators, and the adoption YAML generator is ours, but the failing input here is the device-broadcast URL: external content the generator faithfully embedded. The same principle's exception ("land the config so the user can repair it in the editor") describes exactly what a user needs in this situation: the generated YAML is well-formed, and repairing the `packages:` path (`gl-s10.yaml` to `gl-inet/gl-s10.yaml`) is a one-line editor fix.

## Proposal

- When import validation fails only on remote-package resolution (fetch failure, missing file in repository), keep the file and surface a warning-style result instead of refusing, so the user can repair the package path in the editor. The next compile surfaces the same error with full context if they skip the repair.
- Keep refusing on structural errors in the generated YAML itself (those are generator bugs and follow the existing INTERNAL_ERROR path).
- Make the failure copy action-aware: an import refusal must not reference an editor for a file that was not kept.

The trailing-period doubling in the same message is fixed separately in #2423.

Observed live while testing #2412's adopt flow; the identical failure occurs on main since the import path predates that work.

bdraco:

Adopting a device whose factory firmware broadcasts a dead `package_import_url` (the GL-S10 case, the upstream repo moved `gl-s10.yaml` to `gl-inet/gl-s10.yaml`) could never complete: `import_device` validated the freshly written adoption YAML, the remote package failed to resolve, and the rollback deleted the file while the refusal copy said "Fix the errors in the editor and try again" about a file that no longer existed.

The validator now classifies the failure by semantics instead of message patterns (esphome's git errors carry raw stderr, so substrings are brittle): esphome returns early from validation when the packages pass raises, and the vscode validator roots the resulting error's `range` at the `packages:` block, so when every validation error's `range.start_line` falls inside the top level `packages:` span (computed with the existing scan helpers) the file is kept and `devices/import` returns `{configuration, warning}`; the user repairs the `packages:` path in the editor and the next install surfaces the same error if they skip it. Any error outside the span, or without a range, refuses exactly as before, now with import specific copy ("The import was rolled back; adopt again after fixing the device's package source") that no longer points at a nonexistent editor. No second validator round trip is involved. The premise is pinned against a real `esphome vscode` subprocess in `tests/real_compile/test_adopt_package_validation.py`.

The `?full_config` delegated path is unchanged (its download failures happen before validation). Frontend companion (showing the warning as a toast after adopt) is esphome/device-builder-frontend#1546.

- fixes #2424

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1546

## Checklist


esphbot (changes requested):
> [!CAUTION]
> Critical issues found.
>
> - The packages-stripped probe can never validate clean for an adoption YAML, so the fix never fires
> - Probe stub asserts behaviour the real validator does not have

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

Issue #2397, opened 2026-07-28
[Bug] “microphone” keyword accepts either a single value or a list, but the structured editor accepts only a list


breti:
### What happened?

The “microphone” keyword accepts either a single value or a list. However, the structured editor accepts only a list.

[Docs](https://esphome.io/components/voice_assistant/)

[YAML](https://github.com/esphome/media-players/blob/main/onju-voice/onju-voice.yaml)

<img width="974" height="235" alt="Image" src="https://github.com/user-attachments/assets/121f7d45-6daf-4b32-842a-f6ebf4d5a5da" />

### Dashboard version

1.7.1b1

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

PR #2443, opened 2026-07-29
Carry the maybe_simple_value key through the component catalog

bdraco:

Backend half of #2397 (the structured editor renders `voice_assistant.microphone` as list-only and shows "No items yet" for a YAML that carries `microphone: onju_microphone`).

esphome validates that field with `cv.ensure_list(cv.maybe_simple_value(..., key="microphone"))`: a bare scalar is shorthand for the keyed mapping. The schema bundle encodes this as a `maybe: "<key>"` marker, but the catalog converter only consumed `maybe` for actions/conditions; on component config vars it was silently dropped, so the catalog had no way to say "a bare scalar is also valid" and the frontend treated the scalar as an empty list.

This carries the marker through as a new `ConfigEntry.maybe_key` field (the child key a bare scalar expands into; with `multi_value`, each list item may also be a bare scalar), stamps it in the config-var builder, and regenerates the catalog. The regenerated diff is exclusively `maybe_key` additions, verified by stripping the field and comparing every changed file against HEAD: 154 entries across 51 files, all real `maybe_simple_value` shapes. Beyond the reported `voice_assistant.microphone` (and its `micro_wake_word` / `sound_level` siblings), the marker also lands on the entity-shorthand fields (`bus_voltage: "Name"` on ina2xx, ade7880, bl0906, ltr501, veml7700, sprinkler switches, msa3xx), `packet_transport` sensor/id and encryption-key shorthands, `esp32_ble_server` value shorthands, `micro_wake_word.models`, and the lvgl `update` actions' `id`.

Backend validation is unaffected (`editor.validate_yaml` delegates to real esphome, which already accepts the scalar); the fix is catalog information the frontend needs. The mismatched local venv was aligned to `esphome-constraints.txt` (2026.7.3) to regenerate against the stamped schema version.

- part of #2397 (fixed by the companion frontend PR)

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1547

## Checklist


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `maybe` inherited through `extends` is dropped — `sensor.msa3xx` acceleration fields stay unstamped

esphbot (approved):
> [!TIP]

PR #2471, opened 2026-07-30
Fill wizard recovery defaults when adding the wifi component

### LINKED ISSUE #2468: Adding the wifi component doesn't fill the wizard's secret refs and fallback AP
Creating a device through the wizard emits a recovery-capable wifi block:

```yaml
wifi:
  ssid: !secret wifi_ssid
  password: !secret wifi_password
  ap:
    ssid: Kitchen Lamp Fallback Hotspot
    password: "xxxxxxxxxxxx"

captive_portal:
```

Adding the `wifi` component from the component catalog does none of that. `devices/add_component` runs the generic catalog merge with no wifi awareness, so:

- An untouched add form emits a bare `wifi:` block, which esphome rejects at validation (needs at least ssid/networks/ap).
- Typed credentials land inline in plaintext in the device YAML instead of referencing the shared secrets.
- No fallback `ap:` and no `captive_portal:` are generated, so a device that can't reach the configured network has no recovery hotspot.

Expected: when the shared `wifi_ssid` / `wifi_password` secrets are already defined, adding the wifi component should fill the same defaults the wizard produces (`!secret` references for absent credentials, the fallback AP with a generated password, and `captive_portal:` on the platforms that support it). User-typed values should never be overwritten. When the shared secrets are not defined, today's behaviour stays as is.

### LINKED ISSUE #2469: Secret picker values are emitted as quoted literals by add_component
The editor's secret picker fills a field with a value like `!secret wifi_ssid`. When that field is submitted through `devices/add_component`, the component YAML emitter routes every string through scalar-safe quoting, so the merged YAML comes out as:

```yaml
wifi:
  ssid: "!secret wifi_ssid"
  password: "!secret wifi_password"
```

That is a literal string, not a secret reference — the device would try to join a network literally named `!secret wifi_ssid`.

Chain: `add_component` → `merge_component_yaml` → `generate_component_yaml` → `_emit_field` → `_format_yaml_value` → `_safe_yaml_scalar` (`helpers/yaml/component.py`, `helpers/yaml/scalar.py`), which quotes any `!`-leading scalar. Only `!lambda` has an emit path today (the `{_lambda, _tag}` sentinel).

This affects every component whose form offers the secret picker, not just wifi. Expected: a strict `!secret <name>`-shaped string value should be emitted unquoted so it round-trips as an actual secret reference; anything not matching the strict shape keeps today's quoting.

bdraco:

Creating a device through the wizard emits a recovery capable wifi block, but adding the wifi component from the catalog produced a bare `wifi:` (which fails validation) or inline plaintext credentials, with no fallback hotspot and no `captive_portal:`.

`devices/add_component` now fills the wizard's defaults when it is about to create a new `wifi:` block and the shared Wi-Fi secrets are defined: absent `ssid`/`password` become `!secret wifi_ssid` / `!secret wifi_password`, an absent `ap` gains the fallback hotspot (same SSID derivation and generated password as the wizard), and `captive_portal:` is merged in on the platforms that support it. User supplied values are never overwritten, and without shared secrets the add is unchanged. The presence gate uses the same top level block detector the singleton merge keys on, and the recovery pieces are shared with the wizard (`fallback_ap_ssid`, a hoisted `fallback_ap_psk`, the exported `CAPTIVE_PORTAL_PLATFORMS`, and the `!secret` ref spellings now defined once in `secrets_state`) so the two paths cannot drift.

The component emitter also passes a strict `!secret <name>` string value through unquoted. The editor's secret picker submits exactly that shape, and it previously landed as a quoted literal (`ssid: "!secret wifi_ssid"`), a dead string rather than a secret reference. Anything looser than the strict shape keeps scalar safe quoting.

Verified end to end over the WS API: an empty-fields `devices/add_component` for wifi on an esp32 config with seeded secrets returns the exact wizard shape and the result passes `esphome config`.

- fixes https://github.com/esphome/device-builder/issues/2468
- fixes https://github.com/esphome/device-builder/issues/2469

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

Issue #2480, opened 2026-08-01
Offline badge with name_add_mac_suffix: true — mDNS broadcast name ≠ catalog name


aioue:
## Summary

Device Builder shows **Offline** (or stays **Unknown** then goes **Offline via ping**) for ESPHome nodes configured with `name_add_mac_suffix: true`, even when:

- Home Assistant's ESPHome integration is healthy and entities update normally
- The device is reachable on the LAN (ICMP to its IP succeeds)
- mDNS resolution of the **suffixed** hostname works from inside the add-on container

The root cause is a **name mismatch** between the YAML catalog key (`esphome.name`, e.g. `example-node`) and the mDNS broadcast name (`{name}-{mac_suffix}`, e.g. `example-node-aabbcc`).

## Environment

| Item | Value |
|------|-------|
| Device Builder backend | `esphome-device-builder` 1.7.0 |
| ESPHome CLI / add-on | 2026.7.3 |
| Add-on image | `ghcr.io/esphome/esphome-hassio:2026.7.3` |
| `host_network` | `true` |
| Device | ESP32-S3 class board |
| YAML `esphome.name` | `example-node` |
| YAML | `name_add_mac_suffix: true` |
| mDNS broadcast | `example-node-aabbcc._esphomelib._tcp.local.` |

## Steps to reproduce

1. Configure any ESPHome device with `name_add_mac_suffix: true` and a short base name (under 24 chars).
2. Flash and confirm HA integration works.
3. Open Device Builder in the HA add-on ingress UI.
4. Observe the device card shows **Offline** (reachability drawer: "via ping").

## Expected behavior

Device shows **Online (via mDNS)** once the `_esphomelib._tcp` service is visible, matching HA integration health.

## Actual behavior

- Pre-fix log line (repeated every ping sweep):
  ```
  Device example-node (example-node.yaml): unknown → offline (via ping)
  ```
- Post-fix (with patch applied): `unknown → online (via mdns)` within ~1s of browser start.

## Evidence (from inside the add-on container)

**mDNS resolution:**

```
split ('example-node', 'aabbcc')
suffixed ['192.168.x.x']    # example-node-aabbcc.local
base None                   # example-node.local — does not exist
```

**Zeroconf browse** finds `example-node-aabbcc._esphomelib._tcp.local.` at a routable LAN address.

**Ping** to that address succeeds from the container.

Cross-VLAN mDNS is enabled on the relevant VLANs — not a network/reflector issue.

## Root cause

[...]

Issue #2389, opened 2026-07-28
Renamed adopt seeds an un-withdrawable mdns ONLINE claim until the flash


bdraco:
## Summary

Adopting an importable device under a different name than its factory broadcast seeds an mdns ONLINE claim that nothing can withdraw until the device is flashed. If the user never flashes (or the device disappears before flashing), the adopted entry sits ONLINE indefinitely.

## Mechanism

The adopt seed (`controllers/devices/importable.py`) applies `ONLINE` under the **chosen** YAML name with `claim=True` while the device still broadcasts under its factory `mdns_name`. In the rename-during-adopt case there is no PTR for the chosen name, so no browser `Removed` can fire for it, `should_ping` skips the mdns-owned entry, and the claim is un-withdrawable until the post-flash boot announce under the new name re-enters the normal browser lifecycle.

The no-rename case is fine: the chosen name matches the factory broadcast, a live PTR exists, and the `Removed` lifecycle covers it.

Pre-existing before #2386 (the sweep used to ping the entry and drop the OFFLINE verdict at the priority gate). Bounded by the adopt flow in intent, unbounded in time.

## Direction to evaluate

Seed the renamed-adopt case under the `ping` source instead of `mdns` (the ICMP sweep then owns liveness until the flashed firmware announces), or track the seed and withdraw it via `source_withdrawn` when the factory-name service goes away.

PR #2481, opened 2026-08-01
Map name_add_mac_suffix mDNS broadcasts to catalog names

aioue:
> **Note:** I'm not proposing this as *the* fix — it's a patch that resolved the offline-badge behaviour I hit in my setup. Happy for maintainers to take a different approach; see the linked issue for context and evidence.

Related to #2480.

## Summary

- Add helpers to parse ESPHome's `name_add_mac_suffix` broadcast shape (`{base}-{6hex}`)
- Map suffixed `_esphomelib._tcp` announcements to the configured catalog name via `_find_device_for_mdns_broadcast()`
- Apply reachability against the catalog name; gate ONLINE on PTRs under the broadcast name
- Extend the ping sweep to try `{name}-{mac_suffix}.local` when the base `.local` address does not resolve
- Add regression tests in `test_mdns_name_match.py`

## Test plan

- [x] `pytest tests/test_mdns_name_match.py` (10 passed)
- [x] Hot-patched on a live HA add-on (`host_network: true`); device went `unknown → online (via mdns)` instead of `offline (via ping)`

### DIFF
--- a/esphome_device_builder/controllers/_device_state_monitor/controller.py
+++ b/esphome_device_builder/controllers/_device_state_monitor/controller.py
@@ -29,6 +29,7 @@ from functools import partial
 from typing import Any, Protocol
 
 from ...helpers.async_ import create_eager_task, drain_tasks, log_task_exit
+from ...helpers.hostname import mac_suffix_from_address, split_mac_suffix_broadcast
 from ...helpers.ip import drop_unspecified_addresses, is_unspecified_address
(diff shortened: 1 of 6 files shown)


esphbot (changes requested):
> [!CAUTION]
> Critical issues found.
>
> - Claim gate now rides the *resolved service's* PTR — reintroduces the #2389 adopt latch and fails an existing test
> - Suffix fallback matches any base-named device when the MAC is unknown

PR #2488, opened 2026-08-01
Clear stale addresses on retarget and stop loopback Online latches

### LINKED ISSUE #2486: Removing use_address keeps the stale address and can latch the device Online
Removing `use_address` from a config does not take effect on the dashboard side; the device keeps its old effective address and can stay latched Online off a ping to whatever host now answers at that address.

Repro:

1. Give a config `wifi: use_address: <ip of any live host>` (a gateway reproduces it well) and let the dashboard see it; the ping sweep marks the device Online.
2. Delete the `use_address` line and save.
3. The device stays Online indefinitely, even though the YAML no longer points at that host and the real device may not exist at all.

Two pieces compound here:

1. `Device.address` is StorageJSON derived, and a plain YAML edit never refreshes StorageJSON. `scan_change.py` only schedules `--only-generate` for first-sight ADDED configs missing `loaded_integrations` or `expected_config_hash`; an UPDATED config keeps its stale storage until the next real compile. That also means the address-change ping wake at `scan_change.py:51` never fires, because the StorageJSON `address` never changes.
2. The ping sweep's Online claim is identity blind. Whatever answers ICMP at the configured address counts as the device, so a stale `use_address` pointing at any live host latches Online with no path to demote. This is the same class as the #1776 latches, but for the ping target rather than the mDNS cache.

Possible directions, not mutually exclusive: schedule `--only-generate` when an UPDATED config's resolved network block changed, so `address` tracks the YAML; or read `use_address` from the resolved YAML as the priority source over StorageJSON so an edit takes effect immediately; the identity blindness is harder and probably its own discussion.

Found while testing esphome/device-builder#2485 / esphome/device-builder-frontend#1597, which make `use_address` much easier to set and remove, so this staleness will be hit more often.

### LINKED ISSUE #2492: Troubleshoot probe's _pick_target still claims Online off a loopback target
#2488 made unusable addresses (unspecified or loopback) inadmissible liveness targets: the ping sweep's `_resolve_and_ping` filters its final target list through `drop_unusable_addresses`, so `use_address: 127.0.0.1` applies Offline instead of latching Online off the dashboard host.

The `devices/troubleshoot` probe from #2485 has a sibling target chain (`_pick_target`: DNS answer, then mDNS cache, then RAM `ip_addresses`, then `device.ip`) that `ping.py` asks to keep in lockstep, and it does not filter. A loopback `use_address` therefore pings the dashboard host and applies Online with `apply=True` on exactly the surface a user opens to debug an offline device, reintroducing the claim the sweep now refuses.

Once both PRs are on main, route `_pick_target`'s result through `drop_unusable_addresses` (empty result should report the no-target outcome rather than a fabricated Online). Found by Koan while reviewing #2488; the file is not in #2488's base, so it could not be fixed there.

bdraco:

Removing (or changing) `wifi.use_address` never took effect on the liveness side: the device kept its old effective address and stayed latched Online off whatever host answered there. Three compounding gaps, three fixes:

1. **Stale RAM addresses survived an address change.** The dashboard save path already regenerates StorageJSON (`_persist_yaml_mutation` schedules `--only-generate`), so `Device.address` did update; the permanent latch was downstream. The sweeps under the old `use_address` had stored the resolved IP into `runtime_state.ip_addresses`, and once the new `<name>.local` failed to resolve, `_resolve_and_ping`'s RAM fallback kept pinging that old IP and re-applying Online forever. A new `DeviceStateMonitor.address_retargeted` now runs on the scan-change address branch: it drops the resolved set through the existing `clear_resolved_addresses` path, then wakes the sweep. The clear is skipped while mDNS or MQTT owns the name, since their evidence is identity-carrying and address-independent; a ping-owned claim is exactly the latch class being cleared. Verified live: pre-fix the device stays Online at the stale IP indefinitely after the address refreshes; post-fix it drops Offline on the woken sweep.

2. **No loopback guard anywhere on the liveness paths.** `use_address: 127.0.0.1` rides the DNS cache's literal short-circuit past every apply-side filter and latches Online off the dashboard host itself. Loopback now joins unspecified in `helpers/ip.py` (`is_unusable_address` / `is_usable_ip` / `drop_unusable_addresses`, all call sites renamed), and `_resolve_and_ping` filters its final target list, so a loopback-only target applies Offline instead of pinging. Side benefits: a Pi-hole-style `127.0.0.1` DNS sinkhole now caches as a resolution failure, and a sidecar already poisoned with a loopback IP heals on load. Deliberately unchanged: the DNS cache still returns literal loopback verbatim, so `use_address: 127.0.0.1` SSH-tunnel OTA workflows keep uploading; only the Online claim stops. Link-local stays usable; real devices carry `fe80::` addresses.

3. **Out-of-band edits never refreshed StorageJSON.** A YAML changed on disk outside the API (git pull, external editor) fires scanner UPDATED, which scheduled no regenerate, so the effective address itself stayed stale. Devices now carry an off-wire `network_fingerprint` (a digest of the top-level `wifi`/`ethernet`/`openthread`/`esphome`/`substitutions`/`packages` blocks, computed from the already-read YAML at load), compared against the last value persisted in the device metadata store: a moved digest on UPDATED, or on the cold-start ADDED for edits made while the dashboard was down, schedules `--only-generate`; RELOADED (our own post-save or post-regen reload) only re-seeds the stored value. The regen itself resolves everything; the fingerprint only decides when to pay for it. Gating on it avoids a regen storm when a git pull touches dozens of YAMLs for non-address edits, and full-line comment or blank-line churn doesn't move the digest. `esphome:` covers the derived `<name>.local` default, `substitutions:` the `use_address: ${var}` spelling, and `packages:` a repointed package reference; the set is deliberately over-broad, since a false positive costs one deduped regen. An empty digest (unreadable YAML mid-save, or no address-source block) is treated as no information: it neither compares nor overwrites the stored value. Known limitation, noted in the code: an edit to a referenced package file changes no device YAML, so no scan event fires and nothing triggers the regen; catching that would mean the scanner tracking each device's include graph.

With #2485 now on main, the `devices/troubleshoot` probe's `_pick_target` (the sibling chain `ping.py` asks to keep in lockstep) routes each candidate list through the same unusable-address filter, so the debug dialog can no longer fabricate an Online verdict off a loopback target either.

The ping path's identity blindness (whatever answers ICMP at the right address counts as the device) is out of scope here, per the issue.

- fixes https://github.com/esphome/device-builder/issues/2486
- fixes https://github.com/esphome/device-builder/issues/2492

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2519, opened 2026-08-06
Extract the logged background-task spawn into create_logged_task

bdraco:

Extracts the repeated background task spawn pattern into `helpers.async_.create_logged_task(coro, name)`: an eager named task with a `log_task_exit` done callback, so a crash logs at crash time instead of dying silently. Six sites convert: the cold start refine and MQTT nudge spawns, the three state monitor loops (which also gain task names they never had), and the mDNS interface monitor.

The four previously lazy spawns (the three monitor loops and the interface monitor) become eager; all four suspend at their first statement (bootstrap sleeps or an executor hop), so the only observable delta is the loop's running flag flipping during `start()` instead of on the next loop turn.

`TaskWindow.track` keeps its lazy spawn on purpose, callers register cancel bookkeeping against the returned task before its first segment runs, and an eager start lets an extract outrun that registration (a comment now pins the reason). Its redundant `name`/`label` pair collapses to one `name` used for both the task and the log line.

- follow up to #2514

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/_device_state_monitor/controller.py
+++ b/esphome_device_builder/controllers/_device_state_monitor/controller.py
@@ -25,10 +25,9 @@ import asyncio
 import contextlib
 import logging
 from collections.abc import Awaitable, Callable
-from functools import partial
 from typing import Any, Protocol
 
-from ...helpers.async_ import create_eager_task, drain_tasks, log_task_exit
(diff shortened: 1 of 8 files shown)


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

Issue #2534, opened 2026-08-09
Storage regen never reschedules on out of band edits and the session failed set has no TTL


bdraco:
Two related seams remain around background `--only-generate` scheduling after the #2532 retry fix; both leave a config degraded until the YAML changes or the process restarts.

First, `scan_change` only schedules a regen on ADDED. An out of band edit (git pull, external editor) that fixes a previously stamped YAML clears the failure marker and retry state on UPDATED, but schedules nothing unless the network fingerprint moved; the fresh chance the clearing comment promises only materialises on restart, an API path write, or a fingerprint change. Extending the gate to UPDATED with the same missing fields predicate would close it; the pending and failed guards already dedupe.

Second, the session `failed` set has no TTL. The one hour disk stamp only matters across restarts, since the in RAM set short circuits first within a session; a transient outage that outlives the roughly 3.5 minute retry window therefore sticks for the whole session. Letting `schedule` consult the stamp TTL instead of the untimed set would make RAM and disk agree.

bdraco:
Two more shutdown adjacent seams from the PR #2535 review rounds, same family as the two above. The shutdown stamp writes the full one hour lockout even when only one of the four attempts was spent, so a restart landing inside the 210 second window converts a single transient failure into the full suppression; persisting the strike count alongside the stamp (or scaling the stamp TTL to strikes spent) would block restart loop churn while letting a legitimate restart resume the ladder. And a regen still in flight at shutdown, the long window for a package cloning config, is cancelled by the task drain with no stamp and no strike, so the restart loop churn the stamp targets still happens for that config; covering state.regen.pending at shutdown is the fix, with the trade off that it may stamp a run that was about to succeed.

PR #2535, opened 2026-08-09
Retry storage regeneration with backoff before stamping a failure

### LINKED ISSUE #2532: Cold-start clone race on a fresh packages config stamps a bogus one hour regen failure
On the first scan of a brand new config that uses a remote package, two git clones of the same package race each other and both lose; the storage regen failure is then stamped for an hour, so the device keeps empty `loaded_integrations` and `expected_config_hash` until the YAML is touched or the stamp expires.

Repro, deterministic on a fresh dir: create a config dir containing only a YAML with `packages: {X: github://ApolloAutomation/R_PRO-1/Integrations/ESPHome/R_PRO-1_ETH.yaml}` plus its secrets, start the dashboard, watch with `--log-level debug`.

What happens on the cold start ADDED: the scanner's in process `load_device_yaml` calls `resolve_packages`, which starts cloning into `.esphome/packages/<hash>`; concurrently `storage_regen` spawns `esphome --dashboard compile --only-generate`, whose own package pass sees the half written dir, logs `Removing incomplete clone of https://github.com/ApolloAutomation/R_PRO-1.git@None ... will re-clone` and deletes it while the in process fetch is still running. Result on both sides:

```
DEBUG [storage_regen] Storage regenerate for repro.yaml exited 2: ... Removing incomplete clone ... Cloning https://github.c
DEBUG [device_yaml._loading] Package merge failed for .../repro.yaml; using unmerged config
esphome.git.GitCommandError: fetch-pack: invalid index-pack output
```

Then `stamp_failure` persists `regen_failed_mtime` and `regen_failed_at`, honoured for `_REGEN_FAILURE_TTL_SECONDS` (one hour) plus the same session skip, so nothing retries even though a retry would succeed (the second attempt, with the clone present, works fine).

User visible effect: a freshly added packages based device shows no local hash, and the frontend dependency gate from esphome/device-builder-frontend#1630 cannot see the package provided components, so the false "Add ESP32 Platform" detour comes back for exactly the configs that PR fixed, for up to an hour.

Possible directions: serialise the first resolve per package cache dir (the regen already holds `_regenerate_lock`, but the scanner's in process resolve does not participate); a cross process lock around esphome's git cache; or treat a git clone failure as transient and skip the failure stamp so the next scan retries.

bdraco:

A transient failure of the background `esphome compile --only-generate` used to stamp the one hour suppression immediately, so a fresh packages config hit by the cold start clone race sat with empty `loaded_integrations` and no config hash until the YAML changed. The race itself is fixed upstream, esphome 2026.7.4 locks the clone cache (esphome/esphome#17923) and this repo already floors that release; what remained here is the stamp behaviour.

Every regen failure now gets a bounded escalating retry (30s, 60s, 120s) before the session failure marker and the disk stamp. No stderr classification, a transient environment failure is indistinguishable from a broken YAML without parsing subprocess output, and the budget keeps broken config churn to a few fast runs. A retry success also heals the scanner side for free; the reload it triggers re-runs `load_device_yaml`, whose swallowed package resolve failure produced the degraded device row. The regen bookkeeping moved into a `RegenState` dataclass with the lockstep transitions as methods.

Known residuals are tracked in #2534: out of band edits still schedule no regen on their own, and the session failure marker has no TTL, so an outage outliving the retry window sticks until a YAML change or restart.

- fixes #2532

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -402,13 +402,23 @@ Two rules the refine established that every later caller inherits:
   reload-driven path (labels, build size, post-compile refresh) now means "the
   row changed", not "a reload ran". The gate suppresses every RELOADED side
   effect in `on_scan_change`, not just the frame — all are implied by row
-  equality except `regenerate_failed.discard`, whose real repair path is a
-  YAML edit (`UPDATED`) anyway.
+  equality except `regen.reset` (failure marker, armed retry timer, strike
+  count), whose real repair path is a YAML edit (`UPDATED`) anyway.
 - **The refine emits `RELOADED`, never `UPDATED`.** `DEVICE_YAML_UPDATED`
   (version-history commits) keys on `UPDATED`, and the network-fingerprint
   regen deliberately skips `RELOADED` (it fires on `ADDED`/`UPDATED` to catch
   out-of-band edits), so a restart's refine burst can neither commit to git
   nor spawn `--only-generate` per device.
+- **Regen failures retry before they stamp** (#2532). A failed
+  `--only-generate` gets a bounded escalating retry (3 attempts, 30/60/120s)
+  before the session `regen.failed` marker and the one-hour disk stamp — a
+  transient environment failure (interrupted package clone, DNS) is
+  indistinguishable from a broken YAML without parsing subprocess stderr, so
+  every failure retries. The stamp is written only once the budget is spent;
+  a shutdown mid-window stamps any armed retries first so a restart loop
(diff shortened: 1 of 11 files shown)


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2540, opened 2026-08-09
Guard the catalog against callable-repr config-var keys

### LINKED ISSUE #2531: Component sync emits a callable repr as a config entry key in script.execute.json
`definitions/automations/actions/script.execute.json` contains a config entry whose key is a Python callable repr:

```
{"key":"<function validate_parameter_name at 0x7f1cd8122b60>","type":"string","label":"<Function Validate Parameter Name>", ...}
```

Two problems: the key is not a real config key, it leaks from the schema conversion where a validator callable ends up used as a mapping key; and the memory address churns on every sync run, so each catalog refresh PR carries a spurious one line diff in this file (visible in the 2026.7.4 refresh, PR #2516).

`script/sync_components.py` should detect callable repr keys and either resolve them to the intended key or drop the entry, and ideally fail loudly if a new one appears so the leak class cannot ship silently.

bdraco:

`script.execute.json` shipped a config entry keyed `<function validate_parameter_name at 0x...>`; the schema bundle carries that repr verbatim because esphome's schema dumper stringifies the marker wrapped callable key (`cv.Optional(validate_parameter_name)`, the wildcard accepting user defined script parameter names), and the heap address churned a spurious line into every catalog refresh PR.

The sync now rejects repr shaped keys at the single funnel every config entry key passes through. The acknowledged wildcard (`validate_parameter_name`) is dropped; the editor's dynamic parameter form is driven by the scripts snapshot, not the catalog, and decomposition round trips unknown action keys generically, so nothing consumed the entry. Any other repr key aborts the sync with a message naming the offender, checked before each emit, so a new leak fails the nightly sync instead of shipping. The regenerated catalog diff is exactly the bogus entry disappearing. A shipped artifact test scans every body for full string reprs as keys or values so the class stays out of the tree. An upstream fix to the schema dumper will follow separately; the guard stays regardless since released bundles carry the repr.

- fixes #2531

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

PR #2542, opened 2026-08-09
Gate chip-keyed cv.OnlyWith fields to their platforms in the catalog

bdraco:

The adc sensor form shows an "Nrf Saadc" field on every board, including esp32, with no description and no docs link. Upstream gates that field with a chip keyed key marker, `cv.OnlyWith(CONF_NRF_SAADC, PLATFORM_NRF52)`, and that exact shape falls between our two per-field gating passes: `_collect_platform_constraints` only inspects the value validator for `cv.only_on` closures, while `_only_with_gate` reads the key marker but deliberately skips chip names because "chip gating belongs to supported_platforms", a handoff nothing implemented. The editor is the only guard here; `OnlyWith` without a default is behaviorally a plain `Optional`, so esphome itself never rejects the key on the wrong chip.

This teaches the platform constraint pass to read chip keyed `OnlyWith` markers too: a new `_only_with_platforms` returns the chip names from the marker's component list (canonicalizing `Platform` StrEnum members), and the collector intersects them with any value side gate, warning on a disjoint pair the same way `_platform_set` already does. Mixed lists like zigbee's `["nrf52", "zigbee"]` now land both halves, `supported_platforms: ["nrf52"]` plus the existing `depends_on_component: "zigbee"`; the frontend ANDs them. Catalog regenerated in the same PR; the diff is the expected gating additions on the adc domains and zigbee fields.

The missing description is a docs gap, not a sync bug; the adc docs page never mentions `nrf_saadc`, so there is nothing to backfill. A docs bullet upstream would flow in automatically on a later sync.

- fixes: none filed; reported live from the desktop app while editing an esp32dev config

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

PR #2543, opened 2026-08-10
Recover chip gating from requires_component chains as supported_platforms

bdraco:

Follow up to #2542, closing the value side half of the same gap that Kōan's review flagged (https://github.com/esphome/device-builder/pull/2542#issuecomment-5234498857): chip gating expressed as a `cv.requires_component(<chip>)` validator was still dropped. zigbee's `report` / `endpoint` / `use_device_type` are `requires_component("zigbee")` plus `requires_component("esp32")` upstream yet shipped with only the zigbee gate, so an nrf52 zigbee device offered fields `esphome config` then rejects, and `ieee802154_vendor_oui` leaked its chip as `depends_on_component: "nrf52"` with no platform gate at all.

This lands the invariant that `depends_on_component` never names a chip; chips live in `supported_platforms`. Three pieces, all in the sync script: `_platform_set` gains a leaf for `requires_component` closures whose component is a target platform name (the existing All/Any recursion then intersects chains, so the mixed zigbee chain resolves to `["esp32"]` while the zigbee half stays a component gate); `_requires_component_gate` skips chip names, first non chip in chain order wins, mirroring `_only_with_gate`; and `_extract_default` filters chips out of `default_with` component lists, which drops the redundant chip gates on zigbee's `wipe_on_boot` / `sleepy` (their `supported_platforms: ["nrf52"]` from #2542 carries the restriction).

Catalog regenerated in the same PR; the diff was verified programmatically to contain only the expected change classes. No frontend change, same wire fields, and `isEntryVisible` already ANDs both gates.

- fixes: follow up to the review suggestion on #2542, no separate issue filed

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

PR #2554, opened 2026-08-12
Fix cancel not stopping a running compile on Windows

### LINKED ISSUE #2552: [Bug] 1.9.3. - can't stop/cancel compilation
### What happened?

Started device firmware compilation.
Can't stop it: hitting "Stop"  causes message "Stopped by user", but job still in queue and compilation continues.

Image

Hitting "X" in queue also have no effect.

Image

See it "cancelled" after all compiled, not when it requested to cancel:

Image

### Dashboard version

1.9.3

### How did you install it?

PyPI

### Where is it running?

Local — Windows

### Device YAML config

```yaml
n/a
```

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

On Windows, cancelling a RUNNING compile did not stop it; the build ran to completion and only then flipped to CANCELLED. Two gaps combined to produce that. First, `start_new_session=True` is POSIX only, so a Windows spawn had no tree kill primitive beyond `taskkill /F /T`, and every taskkill failure mode fell back to killing only the direct child, which on a pip install is the `Scripts\esphome.exe` launcher; the python child and the whole platformio and gcc tree below it survived. Second, job finalisation was gated on stdout pipe EOF, and since the chain runs with `close_fds=False`, any surviving descendant holds the inherited write handle, so the job could not finalise until the last compiler exited.

The fix makes a kill-on-close Win32 job object the Windows kill primitive: `tracked_subprocess` assigns each spawn to one right after `create_subprocess_exec` returns. The terminate path runs `taskkill /F /T` first, while the tracked parent is still alive to walk the PID tree from, then `TerminateJobObject` atomically kills every job member; `proc.kill()` is the last resort when both fail. Sweep first ordering matters because the launcher stub can spawn python before the job assignment lands, and such a child is in no job; the sweep is the only kill that reaches it. The wrapper uses pywin32's `win32job` bindings (a new Windows only dependency) rather than hand rolled ctypes structs; esphome-desktop uses the same primitive through the official windows crate. The kill-on-close flag also reaps the tree if the dashboard dies with the handle open. Separately, the output loop is replaced by `_pump_output_until_exit`, which races the reader against the tracked process's reap (asyncio's `Process.wait` resolves only after every pipe disconnects, but `returncode` is stamped promptly); once the process is gone and the job is cancel flagged, the pipe gets a one second drain window and is abandoned, so the job finalises promptly even if a straggler survives holding the pipe.

New coverage: cross platform unit tests for the job object wrapper and the three stage kill chain, a windows-only integration test proving the job object kills a grandchild, pipe drain tests for the pump, and an e2e `firmware/cancel` test that spawns a real process tree and runs on the Windows CI leg with no skipif; it fails on the previous code both ways (tree survives, finalisation stalls).

- fixes https://github.com/esphome/device-builder/issues/2552

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -348,8 +348,9 @@ against legacy behaviour before assuming the simpler version suffices.
   `FirmwareState.compile_lane` / `upload_lane` / `thread_upload_lane`
   (`controllers/firmware/_state.py`); each lane spawns
   `Lane.max_concurrency` workers off one FIFO queue, jobs occupying a
-  lane sit in `Lane.active` (job-id keyed), and running subprocesses live
-  in the job-keyed `FirmwareState.processes` registry so cancel
+  lane sit in `Lane.active` (job-id keyed), and running spawns live in
+  the job-keyed `FirmwareState.spawns` registry (`SpawnHandle` pairs
+  the subprocess with its Win32 job object) so cancel
   (`terminate_job_process`) signals exactly one job. `lane_for(job)`
   routes a network flash of an OpenThread device to the thread lane
   (Thread devices share one mesh/border router — concurrent OTAs starve
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -222,8 +222,11 @@ firmware/install {configuration} → QUEUED → RUNNING → output... → COMPLE
   so cancel signals exactly one job. Cancel kills the whole spawn tree:
   the process group on POSIX (`start_new_session=True` at the spawn), a
   kill-on-close Win32 job object on Windows (the parallel
-  `FirmwareState.job_objects` registry; `TerminateJobObject` first,
-  `taskkill /F /T` then a parent-only kill as fallbacks — #2552). A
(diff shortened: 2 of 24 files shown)


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Successful TerminateJobObject short-circuits the taskkill sweep the PR relies on for pre-assignment escapees

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Successful TerminateJobObject short-circuits the taskkill sweep the PR description relies on

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2558, opened 2026-08-14
Treat per-platform AUTO_LOAD arms as alternatives in the catalog sync

bdraco:

2026.8 made bluetooth_proxy's AUTO_LOAD callable with per platform arms; resolved without a target platform it returns the union of every arm, so the closure walk collected esp32 and rp2 as if both were required, and the platform constraint propagation failed the sync as disjoint on every 2026.8 beta.

Closure platform deps naming more than one platform are alternatives, not requirements; they are now dropped from the shipped dependency list. The supported platforms are resolved by probing the AUTO_LOAD closure once per target chip, admitting each chip whose arm pulls in no foreign platform dep. bluetooth_proxy now emits dependencies `["api"]` and supported_platforms `["bk72xx", "esp32", "ln882x", "rp2"]`, matching the docs.

Verified with a full catalog build against the 2026.8.0b3 schema with esphome 2026.8.0b3 installed; the prune and probe engage only for bluetooth_proxy, every other entry is unchanged. This is only the sync fix; the min esphome pin bump and the release tracked upstream happen separately.

- related: https://github.com/esphome/esphome/issues/18386

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/script/sync_components.py
+++ b/script/sync_components.py
@@ -2622,6 +2622,48 @@ def _assemble_dependencies(
     return dependencies
 
 
+def _prune_alternative_platform_deps(dependencies: list[str], meta: dict[str, Any]) -> list[str]:
+    """
+    Drop closure platform deps naming several platforms.
+
+    A callable ``AUTO_LOAD`` resolved without a target platform unions its
+    per-platform arms, so their platform deps are alternatives, not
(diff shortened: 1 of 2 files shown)


esphbot (approved):
> [!TIP]

PR #2563, opened 2026-08-15
Surface pending config migrations as a per-device dashboard flag

bdraco:
# What does this implement/fix?

Adds `Device.migration_available`, a per device flag telling the dashboard the config carries a legacy spelling the migration fold would respell. Today you only find out when you open the editor, which dry runs `editor/migrate_config`; now the device list can show a dot up front.

The flag is computed in `load_device_from_storage` off the raw main file text the scanner already reads, so it costs nothing extra on the wire and recomputes exactly when the file changes or on cold start (the rules only change with a package upgrade, which is a restart anyway). A token prefilter in the new `has_pending_migrations` keeps the per device cost around 10us on clean configs; the full fold, about 5ms with the current 45 rule index, only runs when a legacy token is actually present. Same scope as the editor nudge: packages and `!include` contents are not scanned. The field is always serialized, false when clean.

**Related issue or feature (if applicable):**

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [x] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Frontend coordination

- [ ] No frontend change needed
- [x] Companion frontend PR: esphome/device-builder-frontend#1652

## Checklist

- [ ] The code change is tested and works locally.
- [x] Pre-commit hooks pass (`ruff`, `codespell`, yaml/json/python checks).
- [x] Tests have been added or updated under `tests/` where applicable.
- [x] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (regenerate via `script/sync_components.py` if a sync is needed).
- [x] Architecture-level changes are reflected in `docs/ARCHITECTURE.md` and/or `docs/API.md`.

PR #2566, opened 2026-08-15
Derive per-block scan state once in the generated-rename fold

### LINKED ISSUE #2564: Migration fold rebuilds per-block scan state for every generated rule
`_apply_generated_renames` in `esphome_device_builder/controllers/migrations.py` loops every generated rule independently. For each rule whose target block exists, `_apply_component_block_field` / `_apply_platform_item_field` rebuild `_block_scalar_mask(lines)` (a full O(lines) regex pass), re-run `find_block_header` + `block_end_index`, re-copy `list(lines)`, and re-walk `top_list_item_starts` + `_item_child_keys` for every item in the block.

With the current 45 rule index (32 rules target `sensor`), a 256 line config with a `sensor:` block pays that derivation 32 times. Measured on a dev checkout:

| config | `render_migrations` |
| --- | --- |
| 31 lines | 0.69 ms |
| 256 lines | 5.7 ms (5.4 ms in this loop) |
| 1006 lines | 22.9 ms |

A config with no matching block still pays ~1.1 ms at 305 lines for 45 `find_block_header` full scans.

This is the hot path for `editor/migrate_config` and for the `has_pending_migrations` fall-through added in #2563 (the predicate has a token prefilter, so only configs containing a legacy token pay the fold, but those pay it in the scanner executor).

Proposed shape: group the generated rules by target block once (module level, off `load_migration_rules_index()`), then per render compute the scalar mask, header index, block end, item starts and child keys once per block, and apply all of that block's rules against the already materialised key list. `_block_scalar_mask` is a pure function of the lines, so it can be hoisted and threaded through; `_migrate_ethernet_clk` changes the line count and invalidates it. Expected roughly 5.7 ms to 0.3 ms on the measured config.

bdraco:

`_apply_generated_renames` looped every generated migration rule independently, so each rule whose target block exists rebuilt the block scalar mask, re-ran the header and block-end scans, and re-walked the item starts and child keys. With 45 rules in the current index (32 targeting `sensor`), a 256 line config with a sensor block paid that derivation 32 times, about 5.4ms of the 5.7ms `render_migrations` cost, scaling linearly with file size.

This groups the generated rules by target block and platform on each render (a microseconds scale pass over the cached rules index), indexes the top level headers in one pass, then computes the scalar mask lazily and the block end, item starts, and child keys once per present block, applying all of that block's rules against the materialized key list. `_respell_item_key` now updates the key list in place, so collision checks and chained renames keep the exact sequential semantics. Measured on the same configs: 5.10ms to 0.28ms at 262 lines, 21.1ms to 1.10ms at 1011 lines. This is the hot path for `editor/migrate_config` and the `has_pending_migrations` fall-through, which since #2563 also runs per device at scanner load.

- fixes https://github.com/esphome/device-builder/issues/2564

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/migrations.py
+++ b/esphome_device_builder/controllers/migrations.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 import re
 from collections.abc import Callable
 from dataclasses import dataclass
-from functools import cache
+from functools import cache, lru_cache
 
 from ..definitions import MigrationRule, load_migration_rules_index
 from ..helpers.yaml.scan import (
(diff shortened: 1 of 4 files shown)


esphbot (approved):
> [!TIP]

PR #2567, opened 2026-08-15
Move the sibling-CLI resolvers to a leaf helpers module

bdraco:
# What does this implement/fix?

`import esphome_device_builder.helpers.config_bundle` failed standalone with a circular import: the module pulled `_find_esphome_cmd` from `controllers.firmware.helpers`, which initialises the whole `controllers.firmware` package, and `remote_runner` imports `BundleBuildError` back from the still partially initialised `config_bundle`. It only worked when `controllers.firmware` happened to be imported first.

The fix is the layering one; the sibling CLI resolvers (`_find_sibling_cli`, `_find_esphome_cmd`, `_find_esptool_cmd`, `helper_cli_cmd`) move verbatim to a new leaf module `helpers/sibling_cli.py`, every importer now points there, and no `helpers` module imports `controllers.firmware` anymore. The lazy import workaround in `chip_detect._run_esptool` is promoted to a normal top level import since the cycle it dodged is gone.

A new guard test imports every `esphome_device_builder.helpers.*` module into a purged `sys.modules` so each one is the very first package module loaded; it fails on the old code and keeps any helper from becoming import order dependent again. The resolver tests move with the code to `tests/helpers/test_sibling_cli.py`.

**Related issue or feature (if applicable):**

None.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [x] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Frontend coordination

- [x] No frontend change needed
- [ ] Companion frontend PR: esphome/device-builder-frontend#<number>

## Checklist

- [x] The code change is tested and works locally.
- [x] Pre-commit hooks pass (`ruff`, `codespell`, yaml/json/python checks).
- [x] Tests have been added or updated under `tests/` where applicable.
- [x] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (regenerate via `script/sync_components.py` if a sync is needed).
- [ ] Architecture-level changes are reflected in `docs/ARCHITECTURE.md` and/or `docs/API.md`.

PR #2571, opened 2026-08-15
Remove module-scope controllers imports from helpers, add AST guard

### LINKED ISSUE #2565: helpers/config_bundle.py fails standalone import via a circular import with the firmware controller
`esphome_device_builder/helpers/config_bundle.py` cannot be imported standalone:

```
$ python -c "import esphome_device_builder.helpers.config_bundle"
ImportError: cannot import name 'BundleBuildError' from partially initialized module
esphome_device_builder.helpers.config_bundle (most likely due to a circular import)
```

The cycle runs through `controllers/firmware/remote_runner.py`, which imports back into `helpers.config_bundle` while it is still initialising. It works in the running process only because `controllers.firmware` happens to be imported first, so the breakage is hidden by import order luck; any new consumer that imports the helper first (a script, a test file, a helper subprocess) hits the ImportError.

Fix the cycle (move the shared symbol to a leaf module, or defer the controllers import), and consider a guard test that imports every `esphome_device_builder.helpers.*` module standalone so no helper is import order dependent.

Found while auditing the helpers to controllers import direction for #2563.

bdraco:

Follow-up to #2567. The standalone-import guard added there is behavioral only; esphbot's review noted two helpers still import `controllers.*` at module scope: `helpers/dashboard_identity.py` (`metadata_transaction` from `controllers.config`) and `helpers/remote_artifacts_materialise.py` (the tarball member names from `controllers.remote_build.artifacts_tarball`), the latter staying cycle-free only through a lazy import inside `artifacts_tarball.py`.

This moves the shared symbols to leaf helpers modules, following the `sibling_cli.py` pattern:

* The metadata sidecar's atomic RMW core (`metadata_transaction`, `_load_metadata`, `_save_metadata`, the lock machinery) moves from `controllers/config/metadata.py` to `helpers/metadata_sidecar.py`; `controllers/config/metadata.py` keeps the per-device field functions. All production and test importers now point at the new home, and `controllers/config/__init__.py` no longer re-exports the moved trio.
* The artifacts tarball member-name constants (`STORAGE_MEMBER_NAME`, `IDEDATA_MEMBER_NAME`, `PLATFORMIO_INI_MEMBER_NAME`, `BUILD_INFO_MEMBER_NAME`) move to `helpers/artifacts_tarball_members.py`; both the packer and the materialiser import from there.

A new static test (`tests/helpers/test_no_controller_imports.py`) walks `esphome_device_builder/helpers/**/*.py` with `ast` and fails on any module-scope import of `esphome_device_builder.controllers` (absolute or relative; `if TYPE_CHECKING:` bodies and function-local imports stay legal), closing the class structurally alongside the behavioral guard.

Two notes:

* The sweep surfaced a third offender that landed with #2563: `helpers/device_yaml/_loading.py` imported `controllers.migrations.has_pending_migrations` at module scope. Rather than hide it behind a function-local import, the migrations fold relocates to helpers where its consumers live: `controllers/migrations.py` moves to `helpers/migrations.py`, and its two dependencies `api_actions.py` and `writing_layout.py` (already helpers-shaped, importing only helpers and models) move from `controllers/automations/` to `helpers/yaml/`. The device loader keeps its plain module-scope import.
* The lazy import inside `artifacts_tarball.py` stays as a deliberate choice; hoisting it imports cleanly today but adds a firmware/remote_build module cycle whose safety hangs on import order. The comment now says exactly that.

`docs/ARCHITECTURE.md`'s two references to `controllers/config.metadata_transaction` are updated to the new path.

- fixes https://github.com/esphome/device-builder/issues/2565
- follow-up to #2567

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2572, opened 2026-08-15
Preserve corrupt metadata sidecar instead of silently discarding it

bdraco:

`_load_metadata` swallowed `JSONDecodeError` the same way as `FileNotFoundError`, returning `{}` with no log. A truncated or hand-mangled `.device-builder.json` therefore looked like a fresh install; the next `metadata_transaction` write-back atomically replaced the file, permanently destroying user-authored labels, friendly names, comments and the dashboard identity. This affects all deployment shapes, including every HA-addon flavor sharing `/config/esphome`.

The fix, as it stands after review:

- Corrupt content (unparsable JSON, and also valid JSON that isn't an object) is quarantined to `.device-builder.json.corrupt` before the loader returns `{}`, so the loss is visible and the original bytes stay recoverable. The missing-file fast path stays silent.
- The quarantine rename runs only under `metadata_transaction`'s locks (`quarantine=True` on the internal loader); the ~10 lock-free read sites stay pure reads that just log a warning, so a stale decode racing a concurrent rewrite can never move a freshly written sidecar aside.
- The rename goes through `atomic_io.replace_with_retry` (promoted from the private `_replace_with_retry`), so a transient Windows sharing violation gets the module's existing backoff.
- If the quarantine still fails, the transaction raises `OSError` at exit instead of overwriting the corrupt bytes — the same loud failure surface as a failed `atomic_write`, so callers already handling `OSError` degrade instead of acting on a write that never landed.
- An earlier `.corrupt` copy is never clobbered; repeat incidents land at timestamped siblings, pruned oldest-first to a cap of 3. The freshly written copy is exempt from pruning regardless of its stamp, so a pre-NTP boot or backwards clock correction can't delete the copy just created.

Flagged by esphbot while reviewing #2571; the bug predates that PR.

- follow-up to esphbot's review on #2571; no standalone issue filed

## Types of changes

## Frontend coordination

## Checklist


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Destructive quarantine rename runs on the lock-free read path and can destroy both copies

esphbot (approved):
> [!TIP]

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Sibling pruning can delete the copy it just created (non-monotonic clock)

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2585, opened 2026-08-18
Add data-driven channel_colors migration for led-strip platforms

bdraco:

esphome/esphome#18474 folds `rgb_order`, `is_rgbw` and `is_wrgb` into a single `channel_colors` key on the addressable led strip platforms. The upstream fold is a custom validator, not a `cv.rename_key`, so the rename canary never sees it, and no existing rule kind can express a three keys to one key fold with a value transform.

This adds a `platform_channel_colors` rule kind. The value fold is coded once in the migration engine; the platforms it applies to are discovered by the sync, which detects `light.migrate_channel_colors` validators on platform schemas and emits one rule row per platform. The committed artifact carries no rows yet because the pinned esphome (2026.8.0b4) predates the upstream merge, so the fold stays dormant until the catalog syncs against an esphome that ships it; platforms adopting the upstream helper later join without a code change. Two guards keep it honest: a fold validator outside a platform schema fails the sync, and a full sync against an esphome that ships the fold but yields no rows fails loud, so an upstream refactor cannot silently drop the migration from the artifact.

Also hoists esphome's `cv.boolean` spellings into `helpers/yaml/scalar.py` as their single home; the raw YAML truthy checks and the new fold both decode through it.

Merge note: best merged after the catalog sync bumps the pin to an esphome that ships the fold, so the one click nudge never rewrites toward a key the user's installed esphome rejects; the empty artifact keeps the fold inert either way, so ordering only affects when the nudge starts appearing.

- upstream change: esphome/esphome#18474

## Types of changes

## Frontend coordination

## Checklist


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Existing `channel_colors` is silently overwritten by the deprecated keys — upstream refuses to guess here

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Existing `channel_colors` is silently overwritten by the deprecated keys — upstream refuses to guess here

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Fold fires on an incomplete key view — a merge-key or quoted flag silently drops the W

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2596, opened 2026-08-19
Harden the channel_colors fold against anchors, rgbw_order, and partial detection

bdraco:

Follow up to #2585 addressing the post merge review findings on the channel_colors migration.

The fold now skips an anchored item, since a sibling merging it via `<<:` would inherit `channel_colors` beside its own legacy key, the combination upstream rejects; it also leaves any item carrying the beta only `rgbw_order` untouched so it keeps failing validation loudly. Fold detection tracks validator placement the way the rename walk does, so a fold reachable only through a nested sub schema routes to the canary instead of emitting a rule the depth one runtime fold can never apply. The missing rules guard now cross checks emitted rows against every built catalog platform carrying both `rgb_order` and `channel_colors`; that expectation comes from the schema to catalog conversion, independent of the closure detection it guards, so a partial detection failure aborts the first row bearing sync instead of shipping a healthy looking artifact. The live emit test parametrizes over the committed catalog's capable platforms, all three now that main carries the b5 catalog, and the real artifact test skips visibly while the artifact predates the rules.

- follow up to #2585

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2595, opened 2026-08-19
Harden the fingerprint scans and re-run validates raced by writes

bdraco:
# What does this implement/fix?

Follow up to the post merge review rounds on #2591, closing the three remaining findings. A zero indent list under a source block (`external_components:` followed by `- source:` at column 0, valid YAML) was truncating the block capture to the bare key line, so source edits in that shape never moved the fingerprint and the warm validator kept its stale modules; `_match_top_level_key` now treats zero indent sequence items as block continuations, which also fixes the same truncation in the network address fingerprint. `_ADDRESS_SOURCE_KEYS` gains the top level `<<:` merge key, matching what #2591 taught the component source set, so repointing a root merge include at a different base file schedules the StorageJSON regen. And a validate raced by a config dir write now re-runs once (bounded) instead of returning the known pre-write result uncached, so the caller gets a post write verdict.

**Related issue or feature (if applicable):**

- fixes https://github.com/esphome/device-builder/pull/2591#issuecomment-5336045423

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Frontend coordination

- [x] No frontend change needed
- [ ] Companion frontend PR: esphome/device-builder-frontend#<number>

## Checklist

- [x] The code change is tested and works locally.
- [x] Pre-commit hooks pass (`ruff`, `codespell`, yaml/json/python checks).
- [x] Tests have been added or updated under `tests/` where applicable.
- [x] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (regenerate via `script/sync_components.py` if a sync is needed).
- [ ] Architecture-level changes are reflected in `docs/ARCHITECTURE.md` and/or `docs/API.md`.

PR #2617, opened 2026-08-20
Version-stamp a full-file content fingerprint for out-of-band edit detection

### LINKED ISSUE #2612: [Bug] Dashboard Information Inaccurate / Missing
### What happened?

Upgrading to 2026.8.0 resulted in 'unstable' dashboard information. Please see attached:

Image

Prior to the upgrade...
- Build size was accurate around 252.6MB, about the same as my reference device not in the screenshot. However, now it shows as only 2.7MB.
- 3 devices (FL0/FL1/FL2) now show yellow/orange dot for modified configuration. I didnt modify anything. FL3 has the same exact configuration, but doesnt have the dot. All 4 of these files reference an include file that has the bulk of the esphome configuration, whereas the yaml (or whatever the markup really is) in the device builder just has a few substitutions. Neither file was modified at all. FYI - There are at least 3 or 4 open feature requests to be able to 'pull in' these external files and make changes to them in the device builder GUI. Lots of upvotes as well.
- Another device, FL3, shows online (green status dot) but the current version is missing. All devices are on the same network, same mDNS responder reflect policies, etc. and nothing has changed. Connecting to logs via the GUI works fine, and the version is seen in the log output.

Not sure how to get any 'logs from device builder' to help, but just saw these odd immediately obvious inconsistencies. I have been upgrading to the point releases for ESPHome, but not necessarily pushing the updated firmware. So, I was on the latest release prior to upgrade this morning. Yes, I hard refreshed the browser tab as well (although usually this is not necessary between upgrades).

### Dashboard version

Device Builder 1.12.0 ESPHome 2026.8.0

### How did you install it?

ESPHome container

### Where is it running?

Docker container

### Device YAML config

```yaml
Not applicable. This is all strictly Device Builder GUI impact.
```

### Relevant logs / errors

```shell
N/A
```

### Anything else?

Usually, the dashboard elements are stable between upgrades, this was not the case here. All devices are accessible, and nothing has changed either on the network, nor with the ESPHome version on these devices.

bdraco:

The scanner's out-of-band edit detector was a hand-scoped digest of the top-level blocks that can carry `use_address` (`_ADDRESS_SOURCE_KEYS`). That key set is a hand-maintained approximation, and it failed the way hand-maintained lists fail: #2595 had to widen it for top-level `<<:` merge keys, and the widening itself moved every stored digest for include-shaped configs, so the first scan after upgrading to the 2026.8.0 container scheduled a background `--only-generate` for those devices. Under the new esphome that regen full-cleaned each build dir (framework bump / component-set change) and rewrote `build_info.json` with the 2026.8.0 hash, which is what the reporter of #2612 saw as "build size collapsed to 2.7 MB" and "modified dots on configs I never touched".

This PR replaces the scoped digest with a versioned full-file content fingerprint:

- `extract_config_content_fingerprint` digests every significant line of the YAML (blank lines, full-line comments, and trailing whitespace still don't move it, so an editor's trailing-whitespace trim stays free), so the key set is gone and there is nothing left to grow stale. A cosmetic edit still schedules nothing; any content edit now regens, which also keeps `expected_config_hash` honest after out-of-band edits to non-address blocks (previously the pending-changes verdict stayed stale until the next compile).
- The stored value carries an algorithm-version prefix (`1:`). `_sync_content_fingerprint` only schedules a regen when the stored digest was produced by the current algorithm (`content_fingerprint_is_current`); a stale-version value re-seeds silently, so a future digest change can never spawn fleet-wide regens and build-dir wipes again.
- The field and store key are renamed `network_fingerprint` -> `content_fingerprint` (backend-internal, `serialize: "omit"`), and the metadata store's decoder now drops fields no longer in `STORE_FIELDS`, so the retired key clears off disk. The rename also makes this upgrade itself silent: the old key reads as first sight, which seeds without a regen. Note the decoder prune makes the store file effectively one-way: a build reading a file written by a newer version discards fields it doesn't know, and the next debounced save persists the pruned shape. Acceptable because every store field is re-derivable (scan, probe, or compile re-fills it); nothing user-authored lives there.

The sibling component-source fingerprint deliberately keeps its scoped key set: it is compared in RAM only (never persisted, so no upgrade hazard) and a false positive there respawns the editor validator on a keystroke path, where narrow scoping is load-bearing.

- fixes https://github.com/esphome/device-builder/issues/2612

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2646, opened 2026-08-31
Pass core.fileMode=false on version-history git invocations

### LINKED ISSUE #2642: [Bug] Version history records config files as mode `100755`, breaking cross-platform (Windows/Samba) access to the same repo
### What happened?

When `/config` is a git repo I created and use that's edited and, until now, only ever committed from Windows (via GitHub Desktop over the Samba add-on), version history's auto-commits — made on the Linux side — record the YAML files as file mode `100755`. From the Windows client those same files read as `100644`, so every version-history commit shows up in Git / GitHub Desktop on Windows as a permanent, un-clearable "mode change" (`100755 → 100644`) with **no content difference** — easily mistaken for corruption and very confusing to figure out what's going on.

The execute bit isn't real intent: it's an incidental host artifact (Samba maps the DOS *archive* attribute of Windows-edited files onto the Unix owner-execute bit). Version history bakes it into history because its Linux-side commits honor that bit.

**Suggested fix (constraints, not a prescription):** Version history shouldn't introduce cross-platform mode churn — it shouldn't record file-mode differences that exist only because the same working tree presents different modes to different clients (`0744` on Linux vs `0644` over Samba). Ideally without persisting a config change to a repo it doesn't own, and without force-normalizing modes the owner may intend to keep. (At minimum, making its own commits with `core.fileMode=false` — e.g. `git -c core.fileMode=false …` — avoids *introducing* the churn without touching the repo config or existing entries.)

**Workaround:** `git config core.filemode false` in the repo (makes git ignore the bit).

**Environment:** ESPHome add-on `2026.7.4`; `/config` is a git repo on the HA host (HA Green in this case), edited from Windows and the repo managed via GitHub Desktop over the Home Assistant Samba add-on.

---

This whole issue only exists because of the introduction of automatic git commits to any edited `/config/esphome` YAML. This repo was never intended to be committed to from Linux, and never had been until ESPHome Device Builder started doing so — which surfaced a problem that took quite a while to figure out. I'm not trying to relitigate the auto-commit decision (there's been plenty of that already); but if you're going to do it, you should do it in a way that can't cause this type of problem.

### Dashboard version

1.9.2

### How did you install it?

Home Assistant App

### Where is it running?

Home Assistant App

### Device YAML config

```yaml
not device specific
```

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Version history now runs every git command with `-c core.fileMode=false`. Files edited from Windows over the Samba add-on carry a fake exec bit (Samba maps the DOS archive attribute onto owner execute), so automatic commits recorded YAMLs as mode `100755`; the same repo viewed from Windows reads them as `100644`, leaving a permanent mode change in every diff, and a bare attribute flip alone could even produce an empty commit. The flag is passed per invocation like the commit identity, so the config of a repo we do not own is never touched; history diffs also stop showing mode noise.

- fixes https://github.com/esphome/device-builder/issues/2642

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/version_history/git_repo.py
+++ b/esphome_device_builder/controllers/version_history/git_repo.py
@@ -61,12 +61,16 @@ _COMMIT_EMAIL = "device-builder@esphome.io"
 # contend with a commit; the required lock add/commit take is unaffected.
 # autoDetach=false keeps auto gc/maintenance in the foreground so ``_run``
 # reaps it instead of git daemonizing a child we never wait on.
+# fileMode=false keeps Samba's fake exec bit (the DOS archive attribute
+# mapped onto owner-execute) out of recorded modes and diffs.
 _GLOBAL_GIT_ARGS = (
     "--no-optional-locks",
     "-c",
     "gc.autoDetach=false",
     "-c",
     "maintenance.autoDetach=false",
+    "-c",
+    "core.fileMode=false",
 )
 
 # Files Device Builder must never let into git history: its own
--- a/tests/controllers/version_history/test_git_repo.py
+++ b/tests/controllers/version_history/test_git_repo.py
@@ -999,6 +999,30 @@ def test_commit_paths_no_change_returns_none(tmp_path: Path) -> None:
     assert repo.commit_paths([yaml], "Update kitchen.yaml") is None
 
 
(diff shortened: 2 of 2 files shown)


esphbot (approved):
> [!TIP]

PR #2660, opened 2026-09-05
Serialise non-string keys in /json-config

### LINKED ISSUE #2659: /json-config returns 500 (TypeError: Dict key must be str) for any config containing a mapping with integer keys — regression from #1767
### What happened?

Home Assistant's ESPHome config flow calls the Device Builder's `/json-config?configuration=<file>` endpoint to read a device's API encryption key before prompting for it. For some perfectly valid configurations the endpoint returns HTTP 500, Home Assistant logs `Error talking to the dashboard: 500 ... url=http://127.0.0.1:<port>/json-config?configuration=<file>`, and the flow falls back to asking for the key by hand. The device is added fine after the paste, so this is an inconvenience rather than a breakage — but it is deterministic and easy to hit.

**Cause.** The endpoint serializes the validated config with `orjson`, which only accepts string dictionary keys:

```
File ".../esphome_device_builder/api/legacy.py", line 449, in legacy_json_config
File ".../esphome_device_builder/api/legacy.py", line 339, in _json_config_response
    return json_response(config)
File ".../esphome_device_builder/helpers/json.py", line 63, in json_response
    body=dumps(body),
File ".../esphome_device_builder/helpers/json.py", line 38, in dumps
    return orjson.dumps(obj)
TypeError: Dict key must be str
```

ESPHome itself validates integer-keyed maps without complaint (for example an option validated as a dict keyed by `int`), so any component that declares such a map produces a config the endpoint cannot serialize.

This looks like a regression from #1767 (the fix for #1765). Before that change the endpoint serialized with a dedicated `dumps_str_non_str_keys` wrapper that set `orjson.OPT_NON_STR_KEYS` precisely because config keys are not always strings; #1767 removed that wrapper and returns `json_response(config)`, which goes through the plain `dumps()` in `helpers/json.py`. Current `main` still has this shape.

**Reproduction.** Any configuration whose validated form contains a mapping with integer keys. Three examples on one host, all reproducible with `curl` from inside the add-on container:

```
curl -s -o /dev/null -w '%{http_code}\n' 'http://127.0.0.1:<port>/json-config?configuration=<file>.yaml'
```

- an external-component option `integer_command_widths: {6: 1, 7: 1, 8: 1, 9: 1, 26: 1}` (datapoint id → byte width) → 500
- a light's effect map `{0: Indicator, 1: Constant, 2: Breath, 3: Cycle}` and brightness map `{1: 33%, 2: 67%, 3: 100%}` → 500
- a `miot` humidifier mode map `{0: "None", 1: "Constant Humidity"}` → 500

The 21 other configurations on the same host, none of which contain an integer-keyed map, return 200, and the add-on log shows exactly one `Dict key must be str` traceback per failing request and none for the others.

**Suggested fix.** Restore `OPT_NON_STR_KEYS` for this endpoint (as before #1767), or coerce non-string keys to strings before serializing in `_json_config_response()`. Integer keys become their decimal strings, which is what a JSON consumer would expect, and the config flow only reads the `api.encryption.key` path, so key stringification elsewhere in the document is harmless to it. A regression test with one integer-keyed map in the fixture config would keep it from coming back a second time.

### Dashboard version

ESPHome Device Builder add-on 2026.8.2 (`ghcr.io/esphome/esphome-hassio:2026.8.2`)

### How did you install it?

Home Assistant Add-on

### Where is it running?

Home Assistant OS (Supervisor)

### Relevant logs / errors

```shell
TypeError: Dict key must be str
```

(full traceback above)

bdraco:

`GET /json-config` returned 500 for any configuration whose resolved form has a mapping with integer keys (light effect maps, `miot` mode maps, external component width maps). The endpoint serialised with strict orjson, which rejects non-string dict keys; #1767 dropped the `OPT_NON_STR_KEYS` wrapper that used to cover this.

Restores `dumps_str_non_str_keys` in `helpers/json.py` and uses it for the endpoint's success response only, so integer keys become their decimal strings, the same shape the legacy dashboard emitted via stdlib json. Error responses keep the strict serialiser. Adds a helper test and an endpoint test with an integer keyed map.

- fixes https://github.com/esphome/device-builder/issues/2659

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/api/legacy.py
+++ b/esphome_device_builder/api/legacy.py
@@ -40,6 +40,7 @@ from ..helpers.event_bus import Event, StreamControls, stream_events
 from ..helpers.json import (
     JSONDecodeError,
     dumps_str,
+    dumps_str_non_str_keys,
     json_response,
     loads,
 )
@@ -336,7 +337,8 @@ async def _json_config_response(db: DeviceBuilder, configuration: str) -> web.Re
         return json_response({"error": "esphome config unavailable"}, status=503)
     if config is None:
         return json_response({"error": "Configuration is invalid"}, status=422)
-    return json_response(config)
+    # User YAML maps can carry ``int`` keys (effect / mode maps); strict orjson rejects them.
+    return web.json_response(config, dumps=dumps_str_non_str_keys)
 
 
 def _parse_encryption_key_payload(raw: bytes) -> tuple[str, str, str]:
--- a/esphome_device_builder/api/legacy.py
+++ b/esphome_device_builder/api/legacy.py
@@ -39,8 +39,8 @@ from ..helpers.device_yaml import EsphomeConfigUnavailableError, run_esphome_con
 from ..helpers.event_bus import Event, StreamControls, stream_events
 from ..helpers.json import (
     JSONDecodeError,
+    dumps_non_str_keys,
     dumps_str,
-    dumps_str_non_str_keys,
     json_response,
     loads,
 )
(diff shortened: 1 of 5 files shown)


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

Issue #2684, opened 2026-09-11
Adopt-time key splice still refuses a !secret api key that already equals the pending key


bdraco:
### What happened?

Follow-up to #2682 / #2683. The adopt-time splice in `controllers/devices/importable.py` (`_splice_pending_key_or_cleanup`, `_splice_pending_key`) still uses the literal-only `api_key_settled` check and then infers indirection from "the upsert was a no-op":

```python
if spliced == content:
    return None, (
        "The imported config supplies its own API encryption key via !secret or a substitution."
    )
```

So an import whose api key is `!secret x` that resolves to exactly the pending HA key gets the scariest outcome for a config that is already correct: the warning says the provisioned key was not applied and installing may cut Home Assistant off, and the pending entry is retained. The handoff endpoint answers `unchanged` for the same YAML since #2683.

### Expected

Same treatment as the push path: resolve the indirected key in process (`encryption_key_lookup.get_resolved_api_key`), consume the pending entry and skip the warning when it matches, and keep the warning only for a key that resolves elsewhere or cannot be resolved. `_splice_pending_key_or_cleanup` is async and holds the controller and path, so it can call the same resolver. That also lets the "provided via !secret" wording live in one place instead of two.

This runs once per adoption, not per connect, which is why it was left out of #2683.

bdraco:
Closing with #2689: the shape needs a user-compiled config with import_full_config kept on, a !secret api key, a deleted YAML with secrets.yaml kept, and re-adoption from discovery. Not a real case; the mid-splice pending-key race found along the way is fixed separately.

PR #2690, opened 2026-09-11
Keep a key Home Assistant pushes while an adoption splice is in flight

bdraco:

Adoption consumes the pending Home Assistant key with an unconditional pop after the splice returns. The splice awaits a validate and a write, and Home Assistant re-pushes its key on every connect, so a push that lands inside that window is newer than the key that was just spliced and gets discarded with it. The next handoff then has nothing to apply.

`PendingKeysStore.pop_if` drops the entry only while it still holds the key that was handled, and the adoption finalize uses it. The handoff endpoint keeps its unconditional pop: a landed push makes any pending entry for that name stale by definition. A test pushes a different key while the splice is in flight and asserts the newer entry survives.

- found while working on #2684, closed as not a real case

## Types of changes

## Frontend coordination

## Checklist

### DIFF
--- a/esphome_device_builder/controllers/devices/_pending_keys_store.py
+++ b/esphome_device_builder/controllers/devices/_pending_keys_store.py
@@ -91,6 +91,13 @@ class PendingKeysStore:
         self._state[name] = entry
         self._store.async_delay_save(self._snapshot, delay=_SAVE_DELAY)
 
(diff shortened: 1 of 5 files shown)


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2692, opened 2026-09-12
Surface project_name / project_version / network for configured devices

davidcoulson:

> **AI-assisted contribution.** This change was written with AI assistance (Claude Code). I've reviewed the diff myself and verified it locally — full test suite, type-check, lint and format all clean (details in the checklist below). Flagging it up front so you can calibrate review depth.

ESPHome broadcasts `project_name`, `project_version` and `network` in the same mDNS TXT as the identity trio (`version` / `config_hash` / `mac`), but only the adoption path ever read them — `AdoptableDevice` carries all three, while a *configured* device dropped them on the floor. So the device table has nothing to sort or filter a fleet by, even though the data is already on the wire:

```
board          esp32-c3
config_hash    8600af66
friendly_name  Eilee Bedroom RRN00 4e88e8
mac            dc06754e88e8
network        wifi
platform       ESP32
project_name   dcoulson.ble-rrn00-wroom06
project_version 2026.09.06.0+ble2026.09.12.0
version        2026.8.2
```

This surfaces all three on `DeviceRuntimeState` for configured devices.

**Descriptive, not identity.** The new keys are ingested through their own `_DESCRIPTIVE_TXT_APPLIERS` table rather than joining `_IDENTITY_TXT_APPLIERS`. That separation is load-bearing: the identity appliers' presence is what stamps `deployed_identity_live`, and a `project_name`-only TXT that vouched for identity freshness would stop the verify-resolve loop from ever re-resolving. They also gate no update or pending-changes verdict. Both halves are pinned by tests.

Applied on **both** the `_esphomelib._tcp` and `_http._tcp` paths — ESPHome publishes the same descriptive set on whichever service the device has, so no-API devices get them too.

**Persisted** to the per-flavor device store alongside `deployed_version`, for the same reason that one is: an mDNS-only field leaves offline rows permanently blank and silently drops them out of a project filter.

**One implementation note for reviewers.** `DeviceStateMonitor` sits at exactly ruff's `PLR0904` ceiling of 20 public methods, and the policy in `pyproject.toml` is explicit that the rule exists to catch new growth, with existing offenders grandfathered rather than new ones added. So instead of three public `apply_*` siblings of `apply_version`, the appliers go through one private `_apply_descriptive_observation` driven by a `(txt_key, field, callback selector)` table. Every entry is the same differ-gate-then-forward, so the table loses nothing — but say the word if you'd rather have the explicit trio and a threshold bump.

- n/a — no filed issue; the mDNS TXT already carries the data.

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1737

The frontend PR adds the Project / Project Version / Network columns and the Project / Network facets that consume these fields. This PR is inert without it (the values ride `runtime_state` and simply go unrendered); that PR without this one renders three empty columns and self-hides both facets.

## Checklist

Verification: `8587 passed, 26 skipped`; `mypy esphome_device_builder` clean; `ruff check` + `ruff format --check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jesserockz:
Thanks for your contributions.

I'm not sure I see the usefulness of the project name and version for the majority of users that would be using the ESPHome Device Builder. 

The project name and version are fields to be used by makers of devices that are providing / selling them and that users are simply on-boarding to their network and using. 

For this reason we don't think at this time that we want to have this extra code to have them showing in the tables. 

The network type is something we are willing to merge in and would like you to split that out so we can get it in there.

Jesse

davidcoulson:
Maybe I'm abnormal (wouldn't be the first time), but I use project name/version extensively in my own projects. For example, I have 60+ Lumary lights running ESPHome on esp32-c3s instead of using the tuya MCUs. There's maybe 3-4 different models of lights, so each model has a different project name that maps to the actual project.

99% of the time I am making a bulk change to one type of light (or as population that have addressable LEDs, etc), so being able to filter by project name then 'select all' when doing select multiple would be super helpful. There's always the odd one that chokes during an update bc 2.4ghz wifi is what it is, so being able to find that one 'old project version' out of a list of 60+ things is nice. I've been exposing that sensor into HA then using that to drive a build list, but it should all be doable in ESPHome Builder.

I know there are labels and I've manually added Labels to a bunch of projects, but I've not figured out a good way to handle that automatically and the volume of ESPHome devices I have makes maintaining that super painful. If there is a better approach so I can use labels in a more useful way and not depend on project sensors LMK and I'll be happy to shift my focus :)

PR #2706, opened 2026-09-13
Classify the inherit and outage verdicts where the validator entries are read

### LINKED ISSUE #2705: Classify the validator's inherit and outage verdicts where the entries are read
Two places in the adoption key step classify validator output at the consumer instead of where the entries are read.

1. The tentative-mint gate matches esphome's inherit message (`_INHERIT_ERROR_MARK` in `importable.py`) against `PackageWarning.messages`, a raw field that exists only for this one reader. The classifier that already reads the package-confined entries is `_packages_confined_warning` in `mutations_yaml.py`; a structural verdict computed there (for example `PackageWarning.only_api_key_to_inherit: bool`) would put the string coupling beside the code that reads the validator and let `messages` go. esphome exposes no error code, so the substring stays, but it gets one home.

2. Outage tolerance has two homes: `validate_rewritten_yaml_or_raise(tolerate_unavailable=True)` returns `None` on an outage, indistinguishable from clean, so `_revalidate_keyed` (which must keep its prior warning) re-implements the tolerance with its own `except ValidatorUnavailableError` and log line, and `encryption_key.py` writes a third log for the same event. A distinct outage return kind from the shared helper would serve every caller with one log.

Surfaced by the review of #2701; both touch the shared validator's return shape, so out of scope there.

bdraco:

Two verdicts the adoption key step used to derive at the consumer now come from the validator seam that reads the entries.

`PackageWarning` carries `only_missing_api_key`, computed in `_packages_confined_warning` next to the one esphome message it matches; the raw `messages` tuple and the adoption's own match on it are gone. `validate_rewritten_yaml_or_raise` returns a `ValidationVerdict` (the package warning, or the outage it tolerated), so the keyed re-check on adoption and the HA key handoff read the outage off the verdict instead of catching `ValidatorUnavailableError` themselves; the outage is logged once, in the helper.

No behaviour change on the wire.

The real-compile pin on that esphome message was gated on 2026.10.0, but the wording shipped in the 2026.9 betas (esphome/esphome#18489), so it was skipping on the CI pin; it now gates on 2026.9.0b1.

- fixes https://github.com/esphome/device-builder/issues/2705

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2707, opened 2026-09-13
Land every pending key on adoption through the esphome-checked splice

### LINKED ISSUE #2704: Land every pending key on adoption through the validated splice path
On the adoption path in `controllers/devices/importable.py` a pending HA key has two landing mechanisms, and a key pushed mid-adoption has a second, unvalidated one.

1. The generated adoption YAML bakes the pending key at generate time (`generate_adoption_yaml(api_encryption=False, api_encryption_key=...)`), and the key step then re-derives that it is already settled via `api_key_settled`. The `?full_config` path lands the same key through `_splice_pending_key_validated`, which lets esphome check it. Generating unkeyed always and landing every pending key through the validated splice would drop the generator parameter (a special case for one caller) and the generate-time peek of the store; the cost is one keyed re-validate, which the full-config path already pays.

2. A key pushed while the keyed YAML is being checked is swapped in at the write by `_prefer_pushed_key`: the line walker splices it over the already-keyed text and it is written without a validator round-trip, with its own refusal wording next to `_NOT_APPLIED_TAIL`. Re-running the pending branch on the unkeyed content while the store's entry differs from the key about to be consumed would give every push the same validated path, and `_prefer_pushed_key` goes away.

Both change what the key step does on the wire (an extra validate on the pending path), so they belong in their own PR with the docs matrix in ARCHITECTURE.md "HA encryption-key handoff" updated. Surfaced by the review of #2701.

bdraco:

Adoption had two ways to land a pending Home Assistant key: the generated YAML baked it at generate time and the key step then re-derived that it was already settled, while a `?full_config` import spliced it in and let esphome check it. Now every pending key lands the same way, spliced into the unkeyed YAML and esphome-checked before the write, so a key esphome rejects (a package with its own OTA key, say) stays stored with a warning instead of landing unvalidated. The generator loses its `api_encryption_key` parameter, which existed for that one caller, and the generate-time peek of the pending store is gone; the key step's own peek is the only one.

The cost is a second validator round-trip on the pending path (unkeyed, then keyed), which the full-config path already paid; validating the keyed text first is filed as #2710, and keying the create path through the same splice as #2709.

- fixes https://github.com/esphome/device-builder/issues/2704

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

PR #2713, opened 2026-09-13
Leave the post-adopt probe to the scan's ADDED handler

### LINKED ISSUE #2703: Adoption re-probes the device after the scan's ADDED handler already did
`import_device` ends with `_drop_importable_row_and_probe(controller, name)` after `controller._scanner.scan()`. The scan fires `ScanChange.ADDED`, and the handler in `scan_change.py` already runs `probe_reachability(name)` (the esphomelib `probe_device` plus the ICMP wake) and `_on_importable_removed(name)`. `_register_new_device` in `controller.py` states that contract in its docstring ("callers must not double-probe") and `mutations_clone.py` honours it.

So on the success path adoption walks the zeroconf cache three times: the scan's probe, then `get_cached_addresses` + `apply_ip_addresses`, then a second `probe_device` (a fresh `AsyncServiceInfo`, `load_from_cache`, and either a full `_apply_service_info` re-apply on a cache hit or a second `resolve_then` task the in-flight guard then discards).

Two things to settle before trimming it:

1. The `get_cached_addresses` + `apply_ip_addresses` leg is a bare A-record apply not gated on a live PTR, while the probe's apply is. ARCHITECTURE.md ("Discovery (mDNS)") describes the adopt path as applying the broadcast's cached IP and claiming mdns off the cache hit; if the un-gated IP apply is intentional it should stay, but the second `probe_device` is still redundant with the scan's.
2. The one case where nothing else probed is the `except Exception` branch where the post-write scan failed; the helper could run only there.

Surfaced by the review of #2701; behaviour change, so out of scope there.

bdraco:

After the post-write scan, adoption re-applied the zeroconf-cached address and probed the esphomelib service a second time. The scan's `ADDED` handler already probes the device (the mDNS resolve plus the ICMP wake) and drops its importable row, and `_register_new_device` states that callers must not double-probe. Adoption now leaves probing to the scan; it still retires the adopted name's discovered row itself so a failed scan does not leave the adopt banner up until the next poll. The un-gated cached-address apply goes with it; the ping sweep already targets a zeroconf A/AAAA cache hit for a `.local` host, and the probe's cache-hit path applies the addresses behind a live PTR.

History: the eager probe at adopt came in with #74, before the scan's ADDED handler gained its own probe in #1964; the cached-address apply survived #2395 only for a rename-during-adopt whose chosen name had no PTR, and #2491 established that the frontend always adopts under the factory broadcast name, so that case was already dead.

- fixes https://github.com/esphome/device-builder/issues/2703

## Types of changes

## Frontend coordination

## Checklist


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #2751, opened 2026-09-19
Hold the validate slot until a lingering child exits

bdraco:

Two follow ups from the review of #2739. When a killed esphome config child did not exit within the reap bound, its pool slot was released anyway, so three unkillable children could quietly make the cap of three a no-op; the slot is now released from a done callback on the child's wait, so a survivor keeps its slot until it really exits. The subtree kill also looked the group up through getpgid, which fails once the direct child has been reaped even though grandchildren in its session are still alive; a child spawned with a new session leads its group, so the kill now signals that group by pid directly and logs a warning when the kernel refuses and only the direct child can be killed.

- none

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

bdraco:
Not worth closing as the memory risk doesn't justify the complexity

esphbot (approved):
> [!TIP]

PR #2759, opened 2026-09-19
Answer not_found for a missing config on every automations command

bdraco:
# What does this implement/fix?

Follow up to #2757. `automations/delete` answered `not_found` for a missing config with `save` and `internal_error` with a logged traceback without it, and `parse`, `upsert` and `get_available` did the same. Every automations read now goes through one `read_device_config` helper that answers `not_found`; `devices/get_config` and `rewrite_yaml` use it too.

`rewrite_yaml` caught `FileNotFoundError` around the whole job, so a missing file raised inside the caller's rewrite was relabelled as a missing device; only the config read maps to `not_found` now. Its docstring says which `update_config` guards it skips. New tests run `automations/delete` with `save` against a real devices controller and check that a failed rewrite records no history.

**Related issue or feature (if applicable):**

- none

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Frontend coordination

- [x] No frontend change needed
- [ ] Companion frontend PR: esphome/device-builder-frontend#<number>

## Checklist

- [x] The code change is tested and works locally.
- [x] Pre-commit hooks pass (`ruff`, `codespell`, yaml/json/python checks).
- [x] Tests have been added or updated under `tests/` where applicable.
- [x] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (regenerate via `script/sync_components.py` if a sync is needed).
- [ ] Architecture-level changes are reflected in `docs/ARCHITECTURE.md` and/or `docs/API.md`.

PR #2760, opened 2026-09-19
Refuse a read-modify-write save when the file changed underneath it

bdraco:

Follow up to the Copilot thread on #2757. `devices/add_component`, `devices/edit_friendly_name` and the HA encryption key edit read the config, work out the new text, in two cases run a validation subprocess, and only then take the file's write lock to save. A save that landed in between, from the editor or from `automations/delete` with `save`, was overwritten with the stale text.

Holding the lock across a validation that can take a minute would stall every other save of that file, so these commands now save through one `persist_if_unchanged` helper, which hands `rewrite_yaml` the text they started from: under the lock, one executor job reads the file, compares and atomically writes. When the file moved on nothing is written, no history is recorded, and the command answers `precondition_failed` so the caller can retry (`not_found` if the file is gone); the key handoff keeps the key pending as it does for any other refusal. Whole file saves from the editor stay last write wins. The guard covers the dashboard's own writers, which all take the file's lock; a program outside the dashboard that writes in the instant between the compare and the rename is not covered, as it was not before.

- none

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Taken in 910ac25e8 and deb5415d0: the three call sites share one persist_if_unchanged helper that names the configuration once; add_component has a command level refusal test where a save lands during the merge; the two draft tests guard rewrite_yaml and the disk instead of a method the command no longer calls; the key handoff's refusal says the key was kept for a later attempt.

Left for the small follow up that finishes the not_found contract from #2759: _read_device_yaml_or_raise answers invalid_args for a file missing at the read while a file that vanishes later is not_found. It is shared with rename, so it moves in one place there rather than here.

@esphbot review

esphbot (approved):
> [!TIP]

PR #2764, opened 2026-09-19
Collect required groups from typed schema branches

bdraco:

Adding an `spi` bus with only `clk_pin` set passes the form, then `esphome config` rejects it with `Must contain at least one of miso_pin, mosi_pin`. The rule sits on the `single` branch of spi's `cv.typed_schema`, and `_collect_required_groups` never looked inside typed schema branches, so the catalog never carried the group.

The walker now descends typed branches at the same path, the way `_walk_schema_keys` already does; both share one small helper. A group carried by more than one branch is emitted once. Checked against every live `CONFIG_SCHEMA`: `spi` is the only component whose catalog output changes.

Checked in a browser with frontend 0.1.333 on a board with no default SPI pins. The banner now shows in the add form and in the editor form, clears when either pin is set, and drops when `type` is `quad`, since `miso_pin` / `mosi_pin` are gated on `type: single`. The add form still lets the bus be added with only `clk_pin`: it is required only, so it hides the optional `miso_pin` / `mosi_pin` fields, and it does not block on an unsatisfied group. That part applies to every group of this kind, not just `spi`, and is handled in esphome/device-builder-frontend#1753.

`spi.json` is regenerated with the script against esphome 2026.9.0. A full sync with `main`'s script reproduces the committed catalog with no diff, and with this change the only file that differs is `spi.json`: it gains `required_groups`, and `miso_pin` / `mosi_pin` gain the matching description hint.

The walker also keys its visited set on the path now, like `_walk_schema_keys`, so a sub schema shared by two fields yields its nested constraint at both. No live schema hits that today, so it changes no catalog output.

- none

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1753

## Checklist


bdraco:
@esphbot review

bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #2773, opened 2026-09-19
Omit default fields on the per-device automation rows

bdraco:

Follow up to #2768. The three per-device automation rows that still carried every default now use the shared catalog config: `AvailableComponentInstance` and `AvailableScript` on `automations/get_available`, and `ParsedAutomation` on `automations/parse`. An unset `name`, `title`, `parent_id` or `error`, a false flag and an empty `parameters` list are left off the wire; `from_dict` still fills them in.

The `AvailableAutomations` container and the `AutomationTree` round trip shapes are deliberately left alone, since the frontend reads their lists as required. `AvailableScript.parameters` is the one field the frontend reads without a guard, and esphome/device-builder-frontend#1758 backfills it at the API boundary.

With this in, the MCP tools in #2733 can drop their empty-value stripping and keep only the hidden and advanced field filter; that is a follow-up there once both merge.

- none

## Types of changes

## Frontend coordination

- [x] Companion frontend PR: esphome/device-builder-frontend#1758

## Checklist


bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `AvailableScript.parameters` omission outruns the pinned frontend

bdraco:
Rebased onto main in 042195b20a454da3ff4ff98725a4d505b2566da6, which now carries the frontend pin 0.1.335 from #2774 with the esphome/device-builder-frontend#1758 backfill, so the `parameters` omission no longer outruns the bundled frontend. The docs paragraph now states the `AvailableAutomations` lists and `AutomationTree` carve-out.

@esphbot review

Issue #2771, opened 2026-09-19
Trigger scanners do not look inside list items or typed schema branches


bdraco:
`_scan_schema_for_trigger_singles` and `_scan_schema_for_list_triggers` in `script/sync_components.py` stop when `_unwrap_schema_to_dict` returns `None`, so they never look inside a `cv.ensure_list` item or a `cv.typed_schema` branch. #2764 fixed the same gap in the required groups walker.

**Correction after measuring it:** this has no catalog impact today. Against esphome 2026.9.0, descending list items and typed branches makes `_live_trigger_singles` gain `on_next`, `on_prev` and `on_value` for `lcd_menu`, `graphical_display_menu` and `display_menu_base`, and nothing else across 747 top level keys. Those are menu item triggers, and the trigger catalog has no entries for them (only the component level `on_enter`, `on_leave`, `on_redraw`, `on_value`), so a full sync with the descent is byte identical to one without it. An earlier version of this issue said the catalog ships them as `supports_list: false`; that was wrong.

The descent also has a risk that should be settled first: both scanners key their result by the bare trigger name, not by path. Looking deeper makes it more likely that a nested item's `on_value` is recorded under the component's own `on_value` and stamps the wrong `supports_list`. No flag flips today, but keying by path should come before any deeper descent.

Whether menu item triggers should be in the trigger catalog at all is a separate question.

Fixed separately: `_ensure_list_item_validator` unpacked `getattr(node, "validators", None)` without checking its type. An esphome `MockObj` answers any attribute with another `MockObj` and any index with one too, so `_list_item_schema(<MockObj>)` never returned. Nothing feeds it a `MockObj` today, but module global dicts hold them (ethernet's clock mode map, msa3xx's enums), which is what made an experimental descent hang.

PR #2775, opened 2026-09-19
Keep the list item probe from hanging on a codegen MockObj

bdraco:

`_ensure_list_item_validator` in `script/sync_components.py` unpacked `getattr(node, "validators", None)` without checking its type. An esphome codegen `MockObj` answers any attribute with another `MockObj` and any index with one too, so `_list_item_schema(<MockObj>)` never returned. The probe is now type strict, like the other schema probes in the script, and only unpacks a real list or tuple.

Nothing feeds the probe a `MockObj` today, so the catalog is unchanged; a full sync against esphome 2026.9.0 produces the same output with and without this. It surfaced while testing a deeper descent for the trigger scanners, which walk module global dicts that hold `MockObj` values (ethernet's clock mode map, msa3xx's enums) and hung on them. That descent is not part of this PR; the issue below explains why.

- part of https://github.com/esphome/device-builder/issues/2771

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

PR #2778, opened 2026-09-20
Stop a component depending on itself through its auto load closure

bdraco:

Modbus cannot be added from the dashboard. Its catalog entry lists `modbus` among its own `dependencies`, so the add dialog says "Modbus Component requires other components", offers "Add Modbus Component", and loops with Add greyed out. Anything that needs modbus, such as Hörmann Cover, hits the same wall.

`_auto_loaded_dependencies` folds in the `DEPENDENCIES` of everything a component auto loads. It dropped the auto loaded names themselves but not the component it started from. Upstream `modbus` auto loads `modbus_client`, which declares `DEPENDENCIES = ["modbus"]`, so modbus picked itself up; this arrived with the 2026.8.0b3 sync when `modbus_client` was added. A bare component is now excluded from its own closure. A platform entry keeps a dependency on its stem, since that names its hub.

`modbus.json` and `components.index.json` are regenerated with the script against esphome 2026.9.0; the only change in either is `dependencies: ["uart", "modbus"]` becoming `["uart"]`. Modbus is the only component in the catalog that listed itself. Checked in a browser: with a `uart:` present, Modbus now adds cleanly.

- none

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

PR #2780, opened 2026-09-20
Let update_config refuse a device YAML that changed since the caller read it

bdraco:

`devices/update_config` replaces the whole file last write wins, which is right for the editor, whose user is looking at the file. A caller working from an earlier read, an agent through the MCP endpoint in particular, can overwrite edits saved in between without noticing.

This adds an optional `expected` argument: the text the caller read. When given, the write goes through `persist_if_unchanged`, so the read, the comparison and the atomic write run as one job under the file's write lock, and a file that moved on answers `precondition_failed` with nothing written. Without `expected` nothing changes. The secrets file refuses `expected` with `invalid_args`, since its save path is a different lock and the per key `config/set_secret` is the guarded way to change it.

The MCP `update_config` tool in #2733 will forward `expected` once this is in.

- none

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #2781, opened 2026-09-20
Pin the expected path's side effects and hoist the secrets name

bdraco:

Follow up to #2780, from its review that landed after the merge. The two `expected` tests now also pin the scan request and the storage regenerate on the happy path and their absence on the refusal, since that path writes through `rewrite_yaml` rather than `_persist_yaml_mutation` and nothing else in the suite covered its side effects. The secrets branch computes the file name once. The `devices/update_config` docs row says that with `expected` the save is a read-modify-write, so it cannot create a file and a deleted one answers `not_found`.

- none

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

PR #2783, opened 2026-09-20
Speed up the Windows pytest job

bdraco:

The Windows pytest job was the one we wait on for every PR, around 200s for the pytest step against 60 to 130s on the other runners. No single test was slow; a temporary probe job on this PR showed the cost was in test setup. An autouse fixture gave every one of the 8.8k tests a `tmp_path`, and on the Windows runner that lands in `%TEMP%` on `C:`, which is about twice as slow as the `D:` work disk for small file I/O. 2195 tests had a setup over 20ms on Windows, 5 on Linux.

Two changes:

- `_core_config_path_in_tmp` only uses `tmp_path` when the test already asks for it, directly or through a fixture. Everything else gets a unique config path that is never created. A test that writes storage has to request `tmp_path`, the full suite already does.
- The Windows leg of the `test` job points `TEMP` and `TMP` at the runner temp dir on `D:`, which also helps the tests that do use `tmp_path`.

Windows pytest step went from 198 to 209s down to 133s on this PR. The Linux probe leg went from 90s to 79s with the fixture change alone, so local runs get a bit faster too. Defender is already off on the runner image, raising the timer resolution and a ReFS temp disk made no difference, so none of that is here. `windows-real-compile.yml` is left alone, it runs a few long compiles and temp dir cost does not matter there. The probe job and script are removed again.

- fixes <link to issue>

## Types of changes

## Frontend coordination

## Checklist

- [ ] `components.index.json` / `definitions/components/*.json` have **not** been hand-edited (regenerate via `script/sync_components.py` if a sync is needed).


bdraco:
@esphbot review

bdraco:
On the Copilot note that `RUNNER_TEMP` is on `C:`; it is not on this image. The Windows job log for this PR shows `TEMP: D:\a\_temp` after the step runs, and the probe measured the same small file tree at 1.9 ms there against 3.9 ms under the default `C:\Users\RUNNER~1\AppData\Local\Temp`. If a future image moves the runner temp dir the step turns into a no-op, nothing breaks.

Took the two comment nits, both comments are shorter now and no longer name a drive letter.

@esphbot review

esphbot (approved):
> [!TIP]

PR #2784, opened 2026-09-20
Let automations/delete refuse an automation that changed or moved since it was listed

bdraco:

An automation location is positional for intervals, light effects and list shaped triggers, so a caller that lists, edits the file, then deletes with the location it remembers removes a different automation and gets a success. The editor never hits this, since it deletes against its own draft through `yaml`; a caller working from a listing does, the MCP tools in #2733 in particular.

This adds an optional `expected` to `automations/delete`: the `raw_yaml` a parse returned for the automation. Inside the same locked rewrite job the current file is parsed again, and the delete happens only while the automation at that location still reads that way; otherwise `precondition_failed` with nothing written. Without `expected` nothing changes. The MCP `delete_automation` tool will require it once this is in.

- none

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

bdraco:
412d369ce2a0c2ed14676e5e7aecb27431337ad4 fixes the Linux CI failure: the test's rewrite stand-in ran the rewrite on the event loop, and the new guard parses the file, which reads the catalog index; the real `rewrite_yaml` runs it in the executor job, and the stand-in now does the same through `asyncio.to_thread`. Verified with the BlockBuster guard forced on locally.

@esphbot review

bdraco:
f1d5536253efad2a4394737211ee7b08b2399419: the happy-path test also parsed on the loop to compute the shown text; that runs in a thread now. Clean under the forced BlockBuster guard.

@esphbot review

bdraco:
f94cac2930c4b599c263844e170de33c963da02a adds the positional regression Copilot asked for: an interval listed at index 0, another inserted above it on disk, delete by the old index with the listed text is `precondition_failed` and nothing is saved. Both refusal tests already assert no save.

@esphbot review

bdraco:
21de462eae2695cc51522cf05e1750ab76e5d4c4: with `expected`, a file that no longer parses answers `precondition_failed` too, with the parser's error chained as the cause; any other error still passes through. Test and docs row updated.

@esphbot review

bdraco:
780cc2826d26db5e266268b306111103f1ea3f8a covers the pass-through branch for a non-parser error.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
In 91093d3ae093ad93bebf97bea36a6fde44784fb9 the unloadable-file branch has its own message carrying the parser's reason, still `precondition_failed` with the cause chained.

Not taken here: `expected` on `automations/upsert`. No caller outside the editor uses it, the editor upserts against its own draft, and the MCP tools in #2733 have no upsert; edits there go through `update_config`, where `expected` is required. If an MCP edit verb on locations ever appears, the parse-and-compare half of this guard is where it starts.

@esphbot review

esphbot (approved):
> [!TIP]

PR #2791, opened 2026-09-20
Say what a null upload_job_id or cancel status means

bdraco:

Follow up to #2733, from the review that landed after its merge. Since `install` keeps the job id when the dependent upload cannot be read, a null `upload_job_id` with `deferred: false` is reachable, and the tool description and the docs row said it was not; a model reading that could install again and flash a device already being flashed. Both now say to poll the compile job instead. The `cancel_job` row also says a null status means the job left the retained history, as its description already did.

- none

## Types of changes

- [x] Documentation only — `docs`

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Late review taken in #2802: the install guidance now says no upload will follow and to install again once the compile is terminal.

PR #2792, opened 2026-09-20
Show a hunk header only when a line under it is shown

bdraco:

Follow up to #2787, from the review that landed after its merge. The per side budget covered removed and added lines but not the `@@` hunk headers, so a scattered rewrite still produced an excerpt with no upper bound, ending in a run of headers with nothing under them. A hunk header is now buffered and written only when a budgeted line under it is shown, and a multi hunk test pins it.

- none

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All three in 0fdce8eabdb93801bc2fe7b54b82d0a315d97c98: the file headers are taken as the first two records before the loop, so the position guards, the enumerate and the truncated comment are gone; the hunk header is appended as is, since unified_diff always terminates it; and a test pins the asymmetric case, one side's budget exhausted while a later hunk still shows a line from the other side under its own header.

@esphbot review

esphbot (approved):
> [!TIP]

PR #2802, opened 2026-09-20
Treat an install without its upload job as a server fault

bdraco:

Follow up to #2791 from its late review, which turned into a different fix. The `install` tool documented a null `upload_job_id` with `deferred: false`, but that state cannot occur: the install chain registers the upload job in the firmware state before `firmware/install` returns, the lookup scans that state with no status filter in the same tick, and a superseded chain is cancelled, not evicted. #2791 and the first version of this PR wrote guidance for a phantom. The branch is now an `internal_error` (a bug if it ever fires, logged as a server fault by the translator), the tool description names the two real outcomes (an upload job id, or null when deferred), and the API row says the same.

- none

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Taken in a414c584af28a7dcde088a81f6c7f40e0e5326eb: the reinstall is conditioned on a successful compile in both the tool description and the API row, with a failed compile pointed at fixing the config instead. Copilot's overview note (a duplicate install while an upload is active) does not apply on this branch: a null upload_job_id with deferred false means no upload job exists, and a second install supersedes the config's queued jobs rather than stacking.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Reworked in 22626eb33535270ca526c2c4a3c88a4cdd1a8b78 after checking whether the case can happen: it cannot (the upload job is in the firmware state before firmware/install returns, the lookup is status agnostic and runs in the same tick, and supersede cancels rather than evicts). So instead of writing guidance for it, the branch is now internal_error and the description names only the two real outcomes. Retitled accordingly.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Taken in 7514d82cc414d1466a0b2da16a863dd34beb7409: the test keeps the firmware controller and registers the head alone, so the branch is reached through the shape the error names.

@esphbot review

esphbot (approved):
> [!TIP]

PR #2808, opened 2026-09-20
Leave an untypable typed hub unfiltered and require the restrictive set

bdraco:

Follow up to #2782 from its late review. A typed hub with a variant the bundle cannot type used to fall through to the union of its typed variants, so it could make a class look satisfiable for its domain while never being judged itself; `_variant_id_classes` now returns the typed key with an empty variant map for that case, and the declarer pass leaves such a hub out of the judged pool with a warning, the same fail open rule the pass applies when no declarer can name the candidates. `build_automations` also takes `restrictive_references` as a required keyword, so a caller that forgets it fails at the call site instead of shipping automation references the component catalog never cross checked; the tests pass an empty set.

- none

## Types of changes

## Frontend coordination

## Checklist


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Late review taken in #2809: the warning names the variant, the fail-open comment says it widens to the hub's siblings on purpose, and restrictive_references is documented beside the other required keyword.

